PR #4, opened 2017-09-28
Implement different scroll excmds

Koushien:
In addition to scrollByLine and scrollByPage, an excmd to scroll by half pages is included to mimic default <C-u> and <C-d> behavior. In a future revision, it may be preferred to have some variable V for scroll values that pulls current window height when called: a user can set their scroll value to V for their own maps/cmds.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -6,9 +6,10 @@ namespace content {
         window.history.go(message.number)
     }
 
-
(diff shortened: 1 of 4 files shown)


cmcaine:
Good call on changing the function declarations in excmds.ts to `export const`.

PR #5, opened 2017-09-30
Implement tabnext, tabprev

Koushien:
Just an equality bug. Consider '===' by default?
Default to 1 (literally next), implies number.
Rename prevtab to match naming convention.
Currently typing "gT" comes out [ "g", "🄰Shift" ].

### DIFF
--- a/src/excmds_background.ts
+++ b/src/excmds_background.ts
@@ -40,20 +40,21 @@ namespace ExCmds {
     export function scrollupline(n = 1) { scrolldownline(n*-1) }
 
     // Tab functions
-    
-    export async function tabnext(increment: number) {
+
+    export async function tabnext(increment = 1) {
         try {
             let current_window = await browser.windows.getCurrent()
             let tabs = await browser.tabs.query({windowId:current_window.id})
             let activeTab = tabs.filter((tab: any) => {return tab.active})[0]
             let desiredIndex = (activeTab.index + increment).mod(tabs.length)
-            let desiredTab = tabs.filter((tab: any) => {return tab.index == desiredIndex})[0]
+            let desiredTab = tabs.filter((tab: any) => {return tab.index === desiredIndex})[0]
             tabgo(desiredTab.id)
         }
         catch(error) {
             console.log(error)
         }
     }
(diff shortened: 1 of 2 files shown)


src/excmds_background.ts
```diff
@@ -40,20 +40,21 @@ namespace ExCmds {
...
             let desiredIndex = (activeTab.index + increment).mod(tabs.length)
-            let desiredTab = tabs.filter((tab: any) => {return tab.index == desiredIndex})[0]
+            let desiredTab = tabs.filter((tab: any) => {return tab.index === desiredIndex})[0]
             tabgo(desiredTab.id)
         }
         catch(error) {
             console.log(error)
         }
     }
+    export function tabprev() { tabnext(-1) }
```
cmcaine: `n*-1` instead

src/excmds_background.ts
```diff
@@ -40,20 +40,21 @@ namespace ExCmds {
...
-    export async function tabnext(increment: number) {
+
+    export async function tabnext(increment = 1) {
         try {
             let current_window = await browser.windows.getCurrent()
             let tabs = await browser.tabs.query({windowId:current_window.id})
             let activeTab = tabs.filter((tab: any) => {return tab.active})[0]
             let desiredIndex = (activeTab.index + increment).mod(tabs.length)
-            let desiredTab = tabs.filter((tab: any) => {return tab.index == desiredIndex})[0]
+            let desiredTab = tabs.filter((tab: any) => {return tab.index === desiredIndex})[0]
```
cmcaine: I don't understand how this would have broken it. 

FYI, increment is probably a string if it's not 1 becuase the exmode parser doesn't know about the type information yet.

cmcaine:
Merged.

src/excmds_background.ts
Koushien: Hilariously, this means entering `tabprev -n` for lets you tabnext as expected.

src/excmds_background.ts
Koushien: Yeah I jumped the gun on that, PTSD from Javascript strings.

PR #8, opened 2017-10-01
Prepend http:// to given address if no schema

Koushien:


bovine3dom:
Ah, you've hit on a problem that I hadn't seen: browser.tabs.create is much more limited than I would have ever imagined.

I assumed it would just put whatever was in the URL field into the location bar and hit enter.

https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/create

We can't even let people open "about:config" et al using it. It's unclear how to let people search using it, either.

cmcaine:
Relevant bug for about: pages: https://bugzilla.mozilla.org/show_bug.cgi?id=1371793

cmcaine:
Relevant bug for file:// URLs

Koushien:
While we wait for resolution from Mozilla, I think I should make it a more general check for `/^(\w+):/` since not all schemes require the two trailing slashes and I'd rather not make life harder for intentful alternatives and too punishing for minor typos. Entering, for example, tabopen https:firefox.com resolves fine for me. Let me know if I should go ahead and do so.

cmcaine:
Yes, that sounds sensible, Koushien

PR #14, opened 2017-10-06
Expose scrollto, add winopen, pin and tabmove

Koushien:
The tabmove function could be made more concise, I'm sure, but initializing m to -1 did not have the desired effect of no arguments being passed -> moving tab to last in index.

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -22,7 +22,7 @@ function process() {
 
 function changecommand(newcommand?: string){
     if (newcommand !== undefined) {
-        clInput.value = newcommand
+        clInput.value = newcommand + " "
     }
     // Focus is lost for some reason.
     clInput.focus()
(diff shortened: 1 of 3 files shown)


cmcaine:
All of these scroll functions are kind of messy. We should just have scrollby, scrollbylines, scrollbypages (a version that accepts real numbers), and scrollTo (version that accepts a % or a tuple of pixel values).

The up/down variants can die: let's rename the down one to be the normal one and remove the up one. We can just pass +ve or -ve values.

Counts won't work properly from normal mode, but it's something I'll get to.

cmcaine:
Similarly scrolltotop/bottom should die: they're just parameterised versions of the scrollto variant I suggested above...

bovine3dom:
> real numbers

Rational numbers are probably a better bet.

cmcaine:
ScrollTo(π) is required for my workflow.

Koushien:
>scrollTo (version that accepts a % or a tuple of pixel values)

Would that be % of total document height? No idea how to get that.

cmcaine:
window.document.scrollHeight in content script.

cmcaine:
Fixed in eb06c64be3f6825a663aa5c49cc38b8e0a782ebb

PR #17, opened 2017-10-16
Implement winclose and reorganize excmds_background.ts

Koushien:


src/excmds_background.ts
```diff
@@ -179,6 +136,40 @@ export function open(url: string) {
...
+    }
+    else if (args[0]) address = args[0]
+    if (address) createData['url'] = (hasScheme(address)? "" : "http://") + address
+    browser.windows.create(createData)
+}
+export async function winclose() {
+    browser.windows.remove((await browser.windows.getCurrent()).id)
+}
+export async function tabopen(address?: string){
+    if (address) address = (hasScheme(address)? "" : "http://") + address
```
dumblob: There are few more frequently used schemes supported by browsers. Namely `https://` and `file://`. `https://` is usually solved by auto redirects or plugins. But `file://` is the pain in the ass. I would therefore propose detecting, whether the `address` variable contains any valid HTTP address (i.e. does not begin with `/` nor `./` nor `../`). If it's a valid HTTP address, then proceed like now, but if it's not, then I would add the `file://` scheme instead.

cmcaine:
We can't open file uri's anyway due to a bug in Firefox. I've volunteered to help and have proposed a sketch of a solution. I am hoping the Firefox Devs get back to me on it soon...

hasScheme does recognise other schemes, though.

dumblob:
>hasScheme does recognise other schemes, though

Does `hasScheme` also recognize URLs without a scheme? In other words, will `hasScheme('/home/test/abc.pdf')` and `hasScheme('./somefile.txt')` and `hasScheme('../yetanotherfile.csv')`, all of them return `true`?

Koushien:
``` javascript
function hasScheme(uri: string) {
    return uri.match(/^(\w+):/) // \w <==> [A-Za-z0-9_]
}
```
There would be no matches in the example you gave.
In addition to what Colin said about not being able to open file uri's from a WebExt context, I explain why I wanted to keep the check simple in https://github.com/cmcaine/tridactyl/pull/8#issuecomment-333414277. For what it's worth I can't get Nightly's native location bar to resolve your uri's either. I must explicitly type `file:///home/test/abc.pdf` and even `file://./somefile.txt` and `file://./yetanotherfile.csv` collapse to `file:///$FILE` for the respective files. Between getting the above bug fixed or having the Firefox team expose their awesome bar completion API, I think an enhancement like the one you suggest can wait for another PR.

cmcaine:
@dumblob Your examples don't have schemes, so hasScheme will correctly return falsey (null).

When we have some way of opening files, we'll put in some logic, probably just detecting a leading '/', '.' or '~'. Firefox doesn't have native support for relative or homedir paths, but we can still sort of support them.

We may do file URIs with a nativeMessaging client.

dumblob:
>Between getting the above bug fixed or having the Firefox team expose their awesome bar completion API, I think an enhancement like the one you suggest can wait for another PR.

Yep, sounds wise.

>Firefox doesn't have native support for relative or homedir paths, but we can still sort of support them.

Did this change in the newest Firefox builds? Last time I used Firefox properly I can remember I used web pages with resources (pictures, icons, *.html files etc.) generated with relative paths like `href="../../icons/yes.jpg"` and it worked well (when served with Apache or nginx; I can't remember, whether also locally, but I would guess yes, because that's the way I was testing the pages).

cmcaine:
Firefox does support relative paths like that, I meant relative paths from your current dir or from your current page.

dumblob:
Hm, that's really weird or I misunderstood. I've checked it right now with the web page I was creating in Sept/Oct 2015 (which I was testing both with and without a server with Firefox as the web browser) and it still works with Firefox 56.01 . The vast majority, if not all, of the links are relative in the form ` ` or `<link rel="icon" type="image/x-png" href="../img/favicon196.png">` or `<script type="text/javascript" src="../jquery.min.js">` and they really work.

cmcaine:
You're misunderstanding me. Those kind of relative links work fine. What doesn't is just typing './foo.jpg' into the url bar.

Vimperator supports the latter functionality and opens the file foo.jpg in the browser's "current directory", which can be changed. Firefox does not support the latter functionality, but has always and will always support the former because it's an important part of the html spec.

cmcaine:
Merged

PR #18, opened 2017-10-16
Implement getnexttabs

Koushien:
realod and tabclose can now operate on the
n-1 tabs to the right of the active tab


cmcaine:
Merged

PR #22, opened 2017-10-21
Undo logic

P4Cu:
So here's my first change :-) All comments are welcome

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -195,9 +195,9 @@ export async function tabnext(increment = 1) {
     let desiredIndex = ((await activeTab()).index + increment).mod(tabs.length)
 
     // Find and switch to the tab with that index
-    let desiredTab = tabs.filter((tab: any) => {
+    let desiredTab = tabs.find((tab: any) => {
         return tab.index === desiredIndex
-    })[0]
+    })
     tabSetActive(desiredTab.id)
 }
 
@@ -236,10 +236,24 @@ export async function tabclose(ids?: number[] | number) {
     }
 }
 
+/** restore most recently closed tab in this window unless the most recently closed item was a window */
 //#background
 export async function undo(){
-    let id = (await browser.sessions.getRecentlyClosed({maxResults: 1}))[0].id
-    browser.sessions.restore(id)
+    const current_win_id : number = (await browser.windows.getCurrent()).id
+    const sessions = await browser.sessions.getRecentlyClosed()
+
+    // The first session object that's a window or a tab from this window. Or undefined if sessions is empty.
+    let closed = sessions.find((s) => {
+        return (s.window || s.tab && (s.tab.windowId == current_win_id))
+    })
+    if (closed) {
+        if (closed.tab) {
+            browser.sessions.restore(closed.tab.sessionId)
+        }
+        else if (closed.window) {
+            browser.sessions.restore(closed.window.sessionId)
+        }
+    }
 }
 
 //#background
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -463,3 +463,4 @@ async function listTabs() {
 // }}}
 
 // }}}
+// vim: tabstop=4 shiftwidth=4 expandtab
--- a/src/tridactyl.d.ts
+++ b/src/tridactyl.d.ts
@@ -34,7 +34,7 @@ declare namespace browser.find {
 }
 
 declare namespace browser.sessions{
-    function getRecentlyClosed(filter: any): any
+    function getRecentlyClosed(filter?: any): any
     function restore(sessionId: number): any
 }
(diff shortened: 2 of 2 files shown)


P4Cu:
Aww I've of course included a file which was my private note...
One more edit

src/excmds.ts
```diff
@@ -236,10 +236,24 @@ export async function tabclose(ids?: number[] | number) {
     }
 }
 
+// restore most recently closed tab in this window unless the most recently closed item was a window
 //#background
 export async function undo(){
-    let id = (await browser.sessions.getRecentlyClosed({maxResults: 1}))[0].id
-    browser.sessions.restore(id)
+    const current_win_id : number = (await browser.windows.getCurrent()).id
+    const win = (await browser.sessions.getRecentlyClosed({}))
```
cmcaine: The filter is optional, so you don't need the empty object.

Typescript complains, but that's only because our definition of the browser.sessions interface is wrong, you can fix it in tridactyl.d.ts (and I'll eventually merge it into the web-ext-types npm module).

cmcaine: More importantly, the object you get back isn't a window, it's an array of session objects, so you should use a different name for this variable

cmcaine: My rewrite:

```typescript
/** restore most recently closed tab in this window unless the most recently closed item was a window */
//#background
export async function undo(){
    const current_win_id : number = (await browser.windows.getCurrent()).id
    const sessions = await browser.sessions.getRecentlyClosed()

    // The first session object that's a window or a tab from this window. Or undefined if sessions is empty.
    let closed = sessions.find((s) => {
        return (s.window || s.tab && (s.tab.windowId == current_win_id))
    })
    if (closed) {
        if (closed.tab) {
            browser.sessions.restore(closed.tab.sessionId)
        }
        else if (closed.window) {
            browser.sessions.restore(closed.window.sessionId)
        }
    }
}
```

Note the change in style of the function comment (two leading stars tell typedoc it's a doccomment), the lack of parentheses around the second await (don't need them unless you want to address a member of the awaited object), and the changes to variable names

src/excmds.ts
```diff
@@ -463,3 +477,4 @@ async function listTabs() {
 // }}}
 
 // }}}
+// vim: tabstop=4 shiftwidth=4 expandtab
```
cmcaine: Maybe add `foldmethod=marker`? I don't know how much other people like folds.

src/excmds.ts
```diff
@@ -195,9 +195,9 @@ export async function tabnext(increment = 1) {
     let desiredIndex = ((await activeTab()).index + increment).mod(tabs.length)
 
     // Find and switch to the tab with that index
-    let desiredTab = tabs.filter((tab: any) => {
+    let desiredTab = tabs.find((tab: any) => {
```
cmcaine: Good catch!

src/excmds.ts
bovine3dom: I never use folds FWIW.

P4Cu: I'm using folds a lot in my stuff ;-)
I'll add the stuff you mentioned in review later.

Koushien: Are modelines usually in the footer? I never use them myself. I know [Mozilla coding style][0] is to put it in the beginning, but I'm aware we are not Mozilla. I think they are more visible at the top so anyone not too keen on having their environment changed can see them right away (and if we want to start enforcing style, a linter would be more effective, I feel).

I also rarely use folds coding. Wikis, however...

[0]: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Mode_Line

cmcaine:
Thanks!

src/excmds.ts
bovine3dom: I've only ever seen them at EOF (I actually assumed they had to be there).

The plan is to use a format tool such as prettier, but we need to fix the macro script to deal with multi line function declarations first.

cmcaine: FWIW, I use vim-sleuth to automatically choose tabstop, shiftwidth, etc.

P4Cu: It's a styling thing but having them on the beginning is good because they're never moved.
On the other hand having them at the end is less disrupting for non-vim users. And IMHO we don't need emacs modeline as clearly nobody here uses emacs (not in a vim-like project) \m/

Koushien: Right, you can do `set nomodeline` so anyone really bothered should be using that. After that, I think we should add this modeline to each file for consistency. I don't feel one extra line at the top is significantly more distracting than one at the bottom, but I'm not too invested in this either way.

P4Cu: I'll do that later.

PR #33, opened 2017-10-27
move clipboard to background

maoanran:
@cmcaine

### DIFF
--- a/src/commandline_content.ts
+++ b/src/commandline_content.ts
@@ -48,10 +48,13 @@ export function resize() {
 }
 
 export function hide(){
-    cmdline_iframe.setAttribute("style", "display: none")
+    cmdline_iframe.setAttribute("style", `position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: 0px; border: 0; padding: 0; margin: 0;`);
 }
 
 export function focus(){
-    show()
     cmdline_iframe.focus()
+}
+
+export function blur() {
+    cmdline_iframe.blur()
 }
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -51,8 +51,38 @@ export function changecompletions(newcompletions: string) {
     completions.innerHTML = newcompletions
 }
 
-function handler(message) {
-    SELF[message.command](...message.args)
+function applyWithTmpTextArea(fn, context = null) {
+    let textarea = document.createElement("textarea")
+    // Scratchpad must be `display`ed, but can be tiny and invisible.
+    // Being tiny and invisible means it won't make the parent page move.
+    textarea.style.cssText = 'visible: invisible; width: 0; height: 0; position: fixed'
+    textarea.contentEditable = "true"
+    document.documentElement.appendChild(textarea)
+    fn && fn.call(context, textarea)
+    document.documentElement.removeChild(textarea)
+}
+
+export function setClipboard(content: string) {
+    applyWithTmpTextArea(scratchpad => {
+        scratchpad.value = content
+        scratchpad.select()
+        document.execCommand("Copy")
+        // // todo: Maybe we can consider to using some logger and show it with status bar in the future
+        console.log('set clipboard:', scratchpad.value)
+    })
+}
+
+export function getClipboard(sender, sendResponse) {
+    applyWithTmpTextArea(scratchpad => {
+        scratchpad.focus()
+        document.execCommand("Paste")
+        console.log('get clipboard', scratchpad.textContent)
+        sendResponse(scratchpad.textContent)
+    })
+}
+
+function handler(message, sender, sendResponse) {
+    return SELF[message.command](...message.args, sender, sendResponse)
 }
 
 try {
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -24,7 +24,6 @@ npm run update-buildsystem
 ### Development loop
 
 ```
-npm run watch &
```
maoanran: we-ext can auto reload from changes.

src/commandline_content.ts
```diff
@@ -48,10 +48,14 @@ export function resize() {
 }
 
 export function hide(){
-    cmdline_iframe.setAttribute("style", "display: none")
+    const height = '0px'
+    cmdline_iframe.setAttribute("style", `position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: ${height}; border: 0; padding: 0; margin: 0;`);
```
maoanran: Set height to 0px to hide cmdline, not sure if this cause bug.

src/commandline_frame.ts
```diff
@@ -47,6 +47,21 @@ export function changecompletions(newcompletions: string) {
...
+    document.execCommand("Copy")
+    // todo: Maybe we can consider to using some logger and show it with status bar in the future
+    console.log('set clipboard:', content)
+}
+
+export function gotourl() {
+    clInput.focus()
+    document.execCommand("Paste")
+    const url = clInput.value
+    console.log('goto url:', url);
```
maoanran: Not done yet, I think we need to notify the content document to do this jump.

src/excmds.ts
```diff
@@ -372,29 +373,21 @@ export function fillcmdline(...strarr: string[]) {
...
+            message("commandline_frame", "gotourl", [])
+            break
+        default:
+            // todo: maybe we should have some common error and error handler
+            throw new Error(`[clipboard] unknown excmd: ${excmd}`)
     }
-    document.documentElement.removeChild(scratchpad)
-    // let pastecontent = scratchpad.textContent
-    // console.log(pastecontent)
+    hidecmdline()
```
maoanran: For blur from cmdline, cause I change the hide way of cmdline.

readme.md
cmcaine: It can, but excmds_macros.py won't be re-run and the typescript won't be recompiled.

So this line watches the development files and rebuilds, and the next line watches the built files and reloads the extension as required.

src/commandline_content.ts
cmcaine: We should probably use `cmdline_iframe.style.cssText = "blah"` rather than setAttr: it's a bit less verbose.

Why have you made height a const?

src/commandline_frame.ts
cmcaine: I think it makes more sense to have a dedicated textarea for copying and pasting and an interface here that just offers copying and pasting.

The paste command can just return a string that the caller can do something with.

Then excmds.ts:clipboard becomes something like:

```ts
export async function clipboard(excmd = "open"){
    switch (excmd) {
        case 'yank':
            const atab = await activeTab()
            await message("commandline_frame", "setclipboard", [atab.url])
            break
        case 'open':
	    open(await message("commandline_frame", "paste"))
            break
        default:
            // todo: maybe we should have some common error and error handler
            throw new Error(`[clipboard] unknown excmd: ${excmd}`)
    }
    hidecmdline()	
}
```

Maybe the functions should be called `getclipboard` rather than paste. Your discretion.

I'm not sure if excmds.ts uses the better messaging system yet. If not, I'll fix it after this review.

src/excmds.ts
```diff
@@ -372,29 +373,21 @@ export function fillcmdline(...strarr: string[]) {
...
+    switch (excmd) {
+        case 'yank':
+            const atab = await activeTab()
+            message("commandline_frame", "setclipboard", [atab.url])
+            break
+        case 'open':
+            message("commandline_frame", "gotourl", [])
+            break
+        default:
+            // todo: maybe we should have some common error and error handler
```
cmcaine: Yes, definitely. I think it's in messaging.ts already, but excmds.ts uses the old messaging system. I'll fix it.

src/excmds.ts
```diff
@@ -372,29 +373,21 @@ export function fillcmdline(...strarr: string[]) {
...
+            throw new Error(`[clipboard] unknown excmd: ${excmd}`)
     }
-    document.documentElement.removeChild(scratchpad)
-    // let pastecontent = scratchpad.textContent
-    // console.log(pastecontent)
+    hidecmdline()
 }
 
 //#content
```
cmcaine: Style issues:

Use lowerCamelCase except for the names of excmds: e.g. `setClipboard`, `goToURL`

src/commandline_content.ts
maoanran: Just easy to review, if this is ok, I think we'd better move the style into css, and use class instead.

src/commandline_frame.ts
maoanran: I want to do this at first, just cause handler in `commandline_frame` does not deal with sendResponse, I can do this.

readme.md
maoanran: Ah, yeah, you are right, will revert this.

PR #34, opened 2017-10-28
cmdline iframe css refactor

maoanran:
@cmcaine

### DIFF
--- a/src/commandline_content.ts
+++ b/src/commandline_content.ts
@@ -19,6 +19,7 @@ function init(){
             console.log("INIT")
             cmdline_iframe = window.document.createElement("iframe")
             cmdline_iframe.setAttribute("src", browser.extension.getURL("static/commandline.html"))
+            cmdline_iframe.setAttribute("id", "cmdline_iframe")
             hide()
             window.document.body.appendChild(cmdline_iframe)
         } catch (e) {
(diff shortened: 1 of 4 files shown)


src/commandline_content.ts
```diff
@@ -34,21 +35,17 @@ document.addEventListener("DOMContentLoaded", init)
 // This second call will fail in the most common case, but makes web-ext reloads effective.
 init()
 
-// The CSS in this file should probably go in static/
 export function show(){
-    // I don't understand how CSS works - but this ensures that the commandline is always at the bottom of the page.
-    cmdline_iframe.setAttribute("style", "position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: 24px; border: 0; padding: 0; margin: 0;");
```
maoanran: I just remove this and don't find any issue, so how can this issue recur? :)

src/commandline_content.ts
```diff
@@ -34,21 +35,17 @@ document.addEventListener("DOMContentLoaded", init)
...
     const height = cmdline_iframe.contentWindow.document.body.offsetHeight + 'px'
-    cmdline_iframe.setAttribute("style", `position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: ${height}; border: 0; padding: 0; margin: 0;`);
+    cmdline_iframe.setAttribute("style", `height: ${height};`);
 }
 
 export function resize() {
-    const height = cmdline_iframe.contentWindow.document.body.offsetHeight + 'px'
-    cmdline_iframe.setAttribute("style", `position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: ${height}; border: 0; padding: 0; margin: 0;`);
+    show()
```
maoanran: Why do we need this resize? looks like same with show.

src/commandline_content.ts
bovine3dom: I can't reproduce the issue either on current master.

It is still reproducible at 242fcce, though, where commenting out that line will stop the command line from showing. Something somewhere else must have fixed the underlying issue.

src/commandline_content.ts
bovine3dom: It is the same - we don't need it any more.

src/commandline_content.ts
```diff
@@ -34,21 +35,17 @@ document.addEventListener("DOMContentLoaded", init)
 // This second call will fail in the most common case, but makes web-ext reloads effective.
 init()
 
-// The CSS in this file should probably go in static/
 export function show(){
-    // I don't understand how CSS works - but this ensures that the commandline is always at the bottom of the page.
-    cmdline_iframe.setAttribute("style", "position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: 24px; border: 0; padding: 0; margin: 0;");
     const height = cmdline_iframe.contentWindow.document.body.offsetHeight + 'px'
-    cmdline_iframe.setAttribute("style", `position: fixed; bottom: 0; left: 0; z-index: 10000; width: 100%; height: ${height}; border: 0; padding: 0; margin: 0;`);
+    cmdline_iframe.setAttribute("style", `height: ${height};`);
```
bovine3dom: Just a little point on style - we aren't using semi-colons at the end of lines. I'm not sure why there was one in the original code.

bovine3dom:
I believe you have introduced a bug where pressing ESC to close the command line does not cause it to lose focus.

Steps to reproduce:

1. Press `:` to open the command line.
2. Press `Escape` to close it.
3. Try scrolling around with j/k, opening command line with `:`, etc. (It shouldn't work).
4. Click on the page to focus it.
5. Press `:` to open the command line and observe all the j/k etc that you've typed.

I haven't yet been able to pinpoint what is causing it.

maoanran:
@bovine3dom HI, thanks for review, do you mean this PR introduce the issue, it seems works after rebase. :)

bovine3dom:
Yes, it does work after rebase. When I did `hub pull (your PR)` it had the issue, so I assumed it was your fault. If you rebase it, I'll merge the PR.

PR #45, opened 2017-11-07
scrollpage: use scrollByPages

hdhoang:
as an alternative, we can use `scrollBy(0, window.innerHeight * n)`, which will open up the possibility of scrolling by partial pages. That would be useful for `C-f` and `C-u` bindings.


cmcaine:
Hmm. We did used to do exactly that, with the inner pages. I must have lost it when I moved over to the new macro system.

Fixed and credited to you :)

PR #46, opened 2017-11-07
POSIXify the shell script

hamhut1066:
Hi,
I'm using a laptop which doesn't have bash installed in the _/bin/bash_ directory, and as such I cannot run your project (without some minor modifications).

This was also the case in a dependency of yours (which I have also submitted an issue to: http://github.com/kilianc/shared-git-hooks/issues/13)

Great work on this, I'm looking forward to playing with it.

### DIFF
--- a/src/commandline_injector.sh
+++ b/src/commandline_injector.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 shopt -s globstar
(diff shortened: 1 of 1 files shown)


cmcaine:
/usr/bin/env isn't a guaranteed path either, but sure, maybe it's more likely to exist than /bin/bash

Thanks for the PR and for your interest.

I'm having some trouble merging the hinting code due to some very confusing bugs, but I hope to have it merged by the end of the day.

PR #58, opened 2017-11-15
Move buffers to general completion system.

Koushien:
User shouldn't notice any difference besides:
* removing "buffer " from cli clears buffers
* retyping "buffer " repopulates buffers
* text after "buffer " filters with smartcase (no tab completion)

We no longer directly set completion's innerHTML.

Thanks for the help @cmcaine.


src/excmds.ts
```diff
@@ -427,10 +427,11 @@ export async function clipboard(excmd = "open", content = ""){
...
 // TODO: Move autocompletions out of excmds.
-
-/** @hidden */
-//#background_helper
-const DEFAULT_FAVICON = browser.extension.getURL("static/defaultFavicon.svg")
+/** Ported from Vimperator. */
+//#background
+export async function tabs() { fillcmdline("buffer") }
+//#background
+export async function buffers() { tabs() }
```
cmcaine: macros.py can't handle function bodies on the same line as the function signature.

Issue #41, opened 2017-11-04
Support keybinds with modifiers


bovine3dom:
e.g, ^U/^D.

This would involve reverting 37119e510a82755c4e57774476bfa2870555c1c4, which drops all keypresses containing modifiers.

dumblob:
Let me note, that there any key can behave like a modifier - it's dependent on OS and the keyboard layout (yep, I'm one of those, who has keys remapped already in a DLL under Windows and therefore I can use e.g. the physical key `Tab` as `Right Alt`).

Why I'm writing this? Because reverting the above mentioned commit would result in support just for, and I quote ` || keyevent.ctrlKey || keyevent.altKey`.

mMontu:
A possible workaround, at least for Windows users, is to map through [AutoHotKey](https://www.autohotkey.com/):

    ;; --------------------------------------------------------
    ;; -- Mapppings for Tridactyl --
    ;; --------------------------------------------------------
    #ifWinActive .*Firefox
       Capslock::Esc
       Ctrl & n::Send gt
       Ctrl & p::Send gT
    #ifWinActive

wnereiz:
@mMontu There's [autokey](https://github.com/autokey/autokey) on Linux

bovine3dom:
@cmcaine and I were planning to actually fix this this weekend, so you might not need to use your work around for long :)

mMontu:
@wnereiz nice, I've never heard about it - have you tried it? Is it close to AHK?

@bovine3dom great news!!! And thanks for this amazing project, it is really good for something in "in a pretty early stage of development".

xerebz:
Any one actually have a workaround for this on Linux? Can you provide an autokey example?

wnereiz:
@mMontu I didn't try it myself. I'm still waiting the support of modifiers.

There's another "dirty" way to solve this problem - to change hardcoded key binding in the firefox source code. The downside is you have to rebuild every time when there is a new update. With the help of [open build service](https://github.com/openSUSE/obs-build), eg. [openSUSE build service](https://build.opensuse.org/), it is possible to trigger this process automatically. And you could even have your own repository.

But it is still the "dirty" way, although it could solve the problem theoretically. I will not do that until there is no choices. Now, the modifier support is on the way. So, I am waiting.

timokau:
Has there been any update on this? I don't want to nag, just to make sure this issue wasn't forgotten (and maybe the problem already resolved in master?)

[...]

PR #75, opened 2017-11-18
Add incrementurl command

johnbeard:
Increments the last number in a URL by a given step (positive or
negative)

Also includes a url_util helper module and some tests for it

Logic taken from vimperator browser.js:17

Any thoughts on keybindings for this? Vimperator uses C-a and C-x for increment and decrement, respectively, but modifiers can't be used in keybindings: #41 (yet?)

Also, should there be a way to open the inc/dec'd address in a new tab?

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -27,6 +27,8 @@ import {ModeName} from './state'
 import * as keydown from "./keydown_background"
 //#background_helper
 import {activeTab, activeTabId} from './lib/webext'
+//#content_helper
+import {incrementUrl} from "./url_util"
 
 /** @hidden */
 //#background_helper
@@ -181,6 +183,19 @@ export function clicknext(dir = "next"){
     // whereas lots of blogs have "VIEW MORE" etc. plastered all over their pages.
     // Stops us from having to hardcode in RPS and reddit, for example.
     window.location.href = nextlinks.slice(-1)[0].href
+}
+
+/** Increment the current tab URL
+ *
+ * @param count   the increment step, can be positive or negative
+*/
+//#content
+export function incrementurl(count = 1){
+    let newUrl = incrementUrl(window.location.href, count)
+
+    if (newUrl !== null) {
+        window.location.href = newUrl
+    }
 }
 
 //#background
--- a/src/url_util.test.ts
+++ b/src/url_util.test.ts
@@ -1,0 +1,30 @@
+/** Some tests for URL utilities */
+
+import {incrementUrl} from './url_util'
+
+function test_increment() {
+
+    let cases = [
+        // simple increment
+        [1, "http://example.com/item/1",   "http://example.com/item/2"],
+        // test non-1 increment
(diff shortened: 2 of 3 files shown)


bovine3dom:
How about ]] and [[:  if the "next" button check fails, call incrementurl? We can't currently use modifiers in normal mode, otherwise ^a and ^x would be ideal.

johnbeard:
Sorry, I sneakily edited that bit in but didn't say so!

I'm not sure increment URL is really the same as click next, especially when you can have a count. Not all URL increments are necessarily a reaction to failing to find a next button.

How about a binding something like like `[]` and `][`, using the existing prefix `[` and `]` as a generic prefix for actions like this? 

Then maybe you could have `[O` (or whatever) be "put the number in the command line for editing, then interpolate into the URL on enter" and so on for other variants?

OTOH, Vimperator also used `g` as a prefix for things like going up a level with `gu` and going to parent on `gU`, so maybe `g[` is appropriate instead? Vimperator seems to mix within `g`, as `gU` is a URL mutator, but `gi` finds a page element, while `[[` and `[f` both find page elements.

cmcaine:
Thanks for the high quality PR!

We'll have modifier keys soonish, so don't sweat too much about choosing a default bind... Though, personally, I find select all with C-a pretty useful, so I'm not sure that should be a default bind for this.

If you're interested in contributing again, we could do with `gu` and friends from vimperator. To make testing slightly less verbose, I wrote a tiny test_utils library in the keyseq branch that might be helpful.

johnbeard:
Indeed, I never liked how Vimperator killed C-a for select all. 

Increment/decrement doesn't seem so common that it even deserves a "top-level" single letter binding, when nearly all of `[` and `g` prefixes (and others) are available.

I'm looking at 'gU`, with `gu` up next.

I'll had a look at the test utils, but when will that likely get merged? I wouldn't want to get in a tangle waiting for code in other branches, or stealing it and causing conflicts! Seems it might save a line or two, but TBH, I wasn't feeling overly hard-done by, it's already fairly terse for unit testing.

cmcaine:
Agreed on the binds.

You can just steal that one file from the keyseq branch. There won't be conflicts unless either file is edited out of sync and they'll be easy to merge.

Command name should maybe be `urlincrement` so it's pseudo namespaced with the future url manipulation commands?

If you don't object, I'll pull this, change the name and add some default binds of `]c` and `[c`, c for count.

johnbeard:
Yes, please pull away and make any changes you like!

Agree on the name change and bindings look good to me.

I'll look at incorporating the keyseq test stuff in the gu/U branch.

bovine3dom:
Merged in eae9cd0dabc47129c73b81ac433c795fdf513f35. Sorry it took so long!

gsbabil:
You may bind your own keys like the following:

```
:bind ,a urlincrement +1
:bind ,x urlincrement -1
```

PR #78, opened 2017-11-19
gobble mode and quickmarks

thejan2009:
Contents of this PR:
 -  gobble mode, which records a requested number of keys and then calls an ex cmd with the keys appended. This would mainly be used for 'marks' features. It can also be used for jumping to some open buffers with 2 keypresses: `bind J gobble 1 buffer`.
 - quickmarks; bookmarks that are bound to keys with `M` in normal mode. `gn[key]` opens a new tab and `go[key]` opens the quickmark in current tab.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -21,6 +21,7 @@ import * as controller from './controller'
 import * as convert from './convert'
 import * as dom from './dom'
 import * as hinting_background from './hinting_background'
+import * as gobble_mode from './parsers/gobblemode'
 import * as itertools from './itertools'
 import * as keyseq from './keyseq'
 import * as msgsafe from './msgsafe'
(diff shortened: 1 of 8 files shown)


cmcaine:
Thanks for this PR!

Your commit broke tridactyl on normal pages because you imported `controller.ts` from `gobble.ts`. You shouldn't import `controller.ts` from anywhere but the background frame.

Fortunately, nothing you're doing needs to be in the content script, so I've moved it all to the background and deleted gobble.ts. Quickmarks now work!

If you've no objections I'll merge this tomorrow.

thejan2009:
I see. After removing console logs, I think this PR will be ready. Except if there are any objections regarding the naming (and missing documentation).

h2ero:
how to use quickmark, I type Mg and type gng, no effect

cmcaine:
There was a race condition in quickmarks and I didn't test it properly. Thanks for asking, I've found and fixed it from 1.4.2 onwards.

h2ero:
nice , it works !

PR #92, opened 2017-11-21
Refactor normal mode parser and SEARCH_URLS, start options UI

hppavilion1:
I've moved all of the normal mode keybindings nmaps into the state module (rather than the normal mode module saved in the browser somehow), changed it from an object to a map, and added a simple configuration UI that allows the user to change their search engine (and thus, what comes up when 's' is pressed); I have also moved the search engine data into its own file (`data.ts`). It is not yet possible to add new search engines from the options UI.

(I was asked to create a pull request that won't automatically merge with master, but I'm not certain how to do that, so... hopefully it can be edited?)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,24 @@
+import state from './state'
+import {SEARCH_NAMES} from './data'
+
+console.log(SEARCH_NAMES)
+
+function setup_ds() {
+    let ds = document.getElementById('default_search')
+    for(let k of SEARCH_NAMES.keys()) {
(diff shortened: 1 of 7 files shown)


hppavilion1:
Wait, I'm uncertain: does storing the state in `browser.storage` (used to save keybindings prior to this PR) allow user settings to automagically move to Firefox on other machines? Because I assumed it was just an ad hoc setup used before `state.ts` was added, but now I'm not so sure.

PR #93, opened 2017-11-21
 Refactor normal mode parser and SEARCH_URLS, start options UI

hppavilion1:
I've moved all of the normal mode keybindings nmaps into the state module (rather than the normal mode module saved in the browser somehow), changed it from an object to a map, and added a simple configuration UI that allows the user to change their search engine (and thus, what comes up when 's' is pressed); I have also moved the search engine data into its own file (data.ts). It is not yet possible to add new search engines from the options UI.

(think I got it right this time)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,24 @@
+import state from './state'
+import {SEARCH_NAMES} from './data'
+
+console.log(SEARCH_NAMES)
+
+function setup_ds() {
+    let ds = document.getElementById('default_search')
+    for(let k of SEARCH_NAMES.keys()) {
+        let opt = document.createElement('option')
+        opt.appendChild(document.createTextNode(SEARCH_NAMES.get(k)))
+        opt.id = k
+        if(k === state.search) {
+            opt.selected = true;
+        }
+        ds.appendChild(opt)
+    }
+
+    ds.addEventListener("change", function(e) {
+        let sel = <HTMLSelectElement>(e.target)
+        state.search = sel.options[sel.selectedIndex].id
+    })
+}
+
+setup_ds()
--- a/src/data.ts
+++ b/src/data.ts
@@ -1,0 +1,27 @@
+export const SEARCH_URLS = new Map<string, string>([
+    ["google","https://www.google.com/search?q="],
+    ["googleuk","https://www.google.co.uk/search?q="],
+    ["bing","https://www.bing.com/search?q="],
+    ["duckduckgo","https://duckduckgo.com/?q="],
+    ["yahoo","https://search.yahoo.com/search?p="],
+    ["twitter","https://twitter.com/search?q="],
+    ["wikipedia","https://en.wikipedia.org/wiki/Special:Search/"],
+    ["youtube","https://www.youtube.com/results?search_query="],
+    ["amazon","https://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords="],
+    ["amazonuk","https://www.amazon.co.uk/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords="],
+    ["startpage","https://www.startpage.com/do/search?query="],
+])
+
+export const SEARCH_NAMES = new Map<string, string>([
+    ["google", "Google"],
+    ["googleuk", "Google (UK)"],
+    ["bing", "Bing"],
+    ["duckduckgo", "DuckDuckGo"],
+    ["yahoo", "Yahoo!"],
+    ["twitter", "Twitter"],
+    ["wikipedia", "Wikipedia"],
+    ["youtube", "YouTube"],
+    ["amazon", "Amazon"],
+    ["amazonuk","Amazon (UK)"],
+    ["startpage","StartPage"],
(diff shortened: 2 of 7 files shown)


cmcaine:
Thanks for this PR!

> (I was asked to create a pull request that won't automatically merge with master, but I'm not certain how to do that, so... hopefully it can be edited?)

You did fine :) I just meant don't worry about ensuring it can be merged automatically. Olie or I'll pull it down to our local machines for testing and we can fix any merge issues then.

cmcaine:
I've rebased it to 1.4.2 because there are some bugs in 1.4.0 and I've converted all the files to unix line endings.

These issues need to be fixed before this PR can be merged. It's up to you if you want to do them or if you want to leave it to us:

This one's easy:

- doing a text replace on the `s` map is pretty hacky and should be avoided (could write a syntax for variable substitution in exstrs, or just leave `s` as it is (people can rebind it easily enough), or add a `-s` option to open etc. that forces a search with the default provider)
- normal_namps is redundant naming: nmap means 'normal map'

These are less easy:

- nmaps no longer stored in firefox sync storage (state.ts uses local storage)
- user's old nmaps not loaded by new system

Perhaps we should accept the search URL bit of the PR now and do the other bits later?

bovine3dom:
Thanks for the PR!

I'm worried that storing the nmaps in state.ts like that will make all of the default binds be stored as soon as a user stores any binds, such that we will never be able to add any more default binds. I had a moderately complicated system for merging saved binds with the default binds to get around that.

Is that something that has been considered in this PR?

hppavilion1:
That's an excellent point that I had not considered, I will have to look into that; moving it to `state.ts` seems like the natural thing to do (especially for configuration purposes), but I had not considered that possibility.

On an unrelated not, I've been having trouble with `config.ts`; since it's only loaded by `settings.html`, it doesn't interact with the rest of the app quite right; specifically, when I change the settings, it _does_ properly update the rest of the app, but it doesn't load them right. Specifically, if I open settings, it _always_ lists Google as the default search engine (because from the perspective of the script, google always _is_ the default search engine), but changing the default works properly. Do you have any clue how to fix this??

johnbeard:
What's the end goal for the config options? I was imagining something a bit more like `.vimrc`/`.vimperatorrc`, with a big text field something like:

    set option
    set nofoo
    set default_search="google"
    set listopt+="quux"

    nmap k j " never go back!

`set`, `nmap` and friends would then be excmds and can also be run from the command line for session-wise changing of options. Then reading the file is just line-by-line execution of excmds. This would be easy to phase in, as you could start with just the simpler `set` commands and add more complex commands one by one.

Things that depend on options or variables set by commands would register with the config module for updates on their variables, so they can react to changes.

The config page would have the textarea and an import export (native messenger can fix that in the longer term), and the text value could be stored in FF sync or local storage. I think this would require a lot less config UI and related interface functions, but it would need a more featureful excmd set and careful documentation of options, or discoverability will suffer.

azuwis:
[Surfingkeys for Firefox](https://addons.mozilla.org/en-US/firefox/addon/surfingkeys_ff/) is out, and the settings page of it is pretty good. It can even load settings from local file.

The [settings](https://github.com/brookhong/Surfingkeys#edit-your-own-settings) is normal javascript, which I think is easier to implement and maintain.

surfinkeys

bovine3dom:
@hppavilion1 Thanks again for going to the effort of making this PR.

I have made a backend to user configuration in the range da68d5f..19d03e6 which is now live on master.

If you (or anyone else) would like to make a settings page that accesses that interface, we would gladly merge it. (We particularly want a "reset" button for the binds, just in case anyone decides to unbind `:`...).

johnbeard:
I have a basic settings page at [my config branch](/johnbeard/tridactyl/tree/config).

It includes a textarea and a button, but I'm not quite sure how to get into config.ts from the settings_page.ts. Anyone know why it doesn't work?

bovine3dom:
@johnbeard I tried to look at your branch but it's giving me a 404. Did you fix the issue?

johnbeard:
Sorry, I moved it to `rcconfig` to avoid aliasing the existing config branch. It's _really_ not much, just a settings page with a textarea and a place that would call a parser. The problem is every time you click "apply", Tridactyl resets and the RC is lost.

https://github.com/johnbeard/tridactyl/tree/rcconfig

hppavilion1:
I do see why a plaintext file (like `.vim.rc` on most systems) is probably preferable, but I still wonder if there's a way to include a GUI for...

...

I guess for users who are savvy enough to use vim and yet not enough to use a basic config file? Point rescinded.

bovine3dom:
@hppavilion1 as I said above,

> If you (or anyone else) would like to adapt this settings page to access that interface, we would gladly merge it. (We particularly want a "reset" button for the binds, just in case anyone decides to unbind :...).

We're happy to have a settings page like this, and it's totally achievable, but nobody has had the time/inclination to make one yet.

bovine3dom:
See also issue #280.

bovine3dom:
Closing due to inactivity in favour of #314.

PR #104, opened 2017-11-21
Add reader excmd (toggle reader mode)

aplaice:
It was already bound to `gr`.

Unless and until the keyboard api is ready or Firefox allows webextensions to inject scripts into the very security-sensitive about:reader context, this will only switch on reader mode. One can get out of reader mode, by pressing <kbd>Alt-←</kbd>.

Once one of the above conditions is met, the toggling should work without any further changes.

I'm not sure if having a command/keybinding for reader mode is something desired at the moment considering that no keybindings will work on the reader page itself. OTOH at least firefox's <kbd>Alt-←</kbd> to exit and <kbd>Space</kbd>/<kbd>Shift-Space</kbd> to scroll up/down, do work, so it's not totally useless. 

A possibly over-kill / duplication-of-effort alternative would be to extract the code from the original readability bookmarklet or even Firefox's reader mode and use that.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -60,7 +60,7 @@ import {ModeName} from './state'
 //#background_helper
 import * as keydown from "./keydown_background"
 //#background_helper
-import {activeTab, activeTabId} from './lib/webext'
+import {activeTab, activeTabId, firefoxVersionAtLeast} from './lib/webext'
 //#content_helper
 import {incrementUrl, getUrlRoot, getUrlParent} from "./url_util"
 
@@ -309,6 +309,18 @@ export function urlparent (){
 export function zoom(level=0){
     level = level > 3 ? level / 100 : level
     browser.tabs.setZoom(level)
+}
+
+//#background
+export async function reader() {
+    if (await l(firefoxVersionAtLeast(58))) {
+	let aTab = await activeTab()
+	if (aTab.isArticle) {
+	    browser.tabs.toggleReaderMode()
+	} // else {
+	//  // once a statusbar exists an error can be displayed there
+	// }
+    }
 }
 
 // }}}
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -311,6 +311,20 @@ export function zoom(level=0){
...
 }
 
+//#background
+export async function reader() {
+    let aTab = await activeTab()
+    if (aTab.isArticle) {
+	// need tabs as any, since typescript believes that:
+	// Property 'toggleReaderMode' does not exist on type 'typeof tabs'.
+	// (Though actually, the code transpiles correctly...)
+	(browser.tabs as any).toggleReaderMode()
```
bovine3dom: The preferred way of doing this is to add some lines to tridactyl.d.ts and add the function there.

Every now and then we'll merge it upstream into web-ext-types.

bovine3dom:
Thanks for the PR. It looks mostly fine.

Would you like to make the minor change (see inline comment) regarding the type of tabs and the definition in tridactyl.d.ts, or shall I?

aplaice:
I've never used typescript before (it seems slightly better than js, though that's not much of an achievement), so I just used the ugly hack that Stackoverflow suggested, but your approach is definitely The Right Thing To Do™.

I'll try to fix this.

bovine3dom:
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/toggleReaderMode <- suggests that this is only available in FF58+, so you should wrap the code like it is in hinting.ts on lines 256 & 265.

Thanks!

aplaice:
> https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/toggleReaderMode <- suggests that this is only available in FF58+, so you should wrap the code like it is in hinting.ts on lines 256 & 265.

Will do it and thanks for the guidance!

bovine3dom:
You're exactly right about setZoom. I was too lazy to do it properly last time. Thanks!

bovine3dom:
All works nicely, thanks!

PR #115, opened 2017-11-22
Add image hinting (;i, ;I)

johnbeard:
This supports hinting of  tags, and will open the src attribute in
the current or BG tab.


bovine3dom:
Thanks!

PR #116, opened 2017-11-23
add `pre`, `code`, `span` to HINTTAGS_text_selectors

emanresusername:
i'm pretty bullish on `pre` and `code`

could see the argument that `span` is overkill in some cases, 

i tend to err on the side of more hints than less i guess

and for precedent's sake, vimfx seems to include `span`s in it's `yv` binding

on that note, there was also a [`cycle overlapping hints`](https://github.com/akhodakivskiy/VimFx/blob/376486f4dc61d75add1c26263d1f4baf39d3abd4/extension/locale/en-US/vimfx.properties#L122-L123) feature there, so even if it did get a little cluttered, you could always get at the hint you wanted. not sure what the overlapping hint handling story is like here?

### DIFF
--- a/src/hinting.ts
+++ b/src/hinting.ts
@@ -246,7 +246,10 @@ iframe,
 textarea,
 button,
 p,
-div
+div,
+pre,
+code,
+span
 `
 
 import {activeTab, browserBg, l, firefoxVersionAtLeast} from './lib/webext'


bovine3dom:
Thanks.

I agree with you about needing to change the order of hints. I'll add it to #28.

lydell:
In case you're curious, VimFx showed hints for every element (regardless of type) that had text non-empty, non-whitespace-only text nodes as direct children.

emanresusername:
Interesting

I'm a "the more the better" guy, I'd be all for that 

I'm already thinking to my self titled now `h\d` maybe `b` or `i` might also want to be copied 🤔 

Not totally sure what the implementation of that would look like? 
Or if that's a performance concern? 
Or if there were other considerations that went into to the "whitelist of tags" approach? 
Maybe a stricter binding for the common case elements like this
And then a hail-mary/everything-and-the-kitchen-sink one for special occasions?

bovine3dom:
Hints for all non-empty textContent added in 6cd93b2.

Thanks @lydell.

PR #125, opened 2017-11-24
Add focusinput command (gi)

johnbeard:
Can focus the nth input from the start or end of a page.

Also has some special modes:

  -l: last focussed input (bound to 'gi')
  -p: first password input
  -b: the largest input by area

----
There are several helpers now in `dom.ts`. I'm open to putting them elsewhere if that's not a suitable place. I think the `isSubstantial` filter makes sense (it works when `isVisible` doesn't), but could do with a sanity check!

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -95,6 +95,27 @@ export function* elementsByXPath(xpath, parent?)
     }
 }
 
+/** Type for functions that can filter element arrays */
+interface ElementFilter { (element: Element): boolean }
+
+/** Is the element of "substantial" size and shown on the page. The element
+ * doesn't need to be in the viewport. This is useful when you want to
+ * scroll to something, but still want to exclude tiny and useless items
+ */
+export function isSubstantial (element: Element) {
+    const clientRect = element.getClientRects()[0]
+    const computedStyle = getComputedStyle(element)
+   // remove elements that are barely within the viewport, tiny, or invisible
+    switch (true) {
+        case !clientRect:
+        case clientRect.width < 3:
+        case clientRect.height < 3:
+        case computedStyle.visibility !== 'visible':
+        case computedStyle.display === 'none':
+            return false
+    }
+    return true
+}
 
 // Saka-key caches getComputedStyle. Maybe it's a good idea!
 /* let cgetComputedStyle = cacheDecorator(getComputedStyle) */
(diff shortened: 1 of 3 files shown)


src/dom.ts
```diff
@@ -158,3 +179,52 @@ export function isVisible (element: Element) {
...
+
+        return <HTMLElement>inputs[index]
+    }
+
+    return null
+}
+
+/** Get the largest element out of an array of Elements
+ */
+export function getLargestElement(elems: Element[]) {
```
bovine3dom: I think a neater way to do this would be more like completions.ts:321 - use sort and then pick the first element.

I'm happy to be persuaded otherwise : )

johnbeard:
I have changed it to a `sort` based function, but I left the "sorter" in `dom.ts` so that other functions can use it. "Nth largest element" sounds like it could be a handy thing to have on hand for other commands?

src/dom.ts
```diff
@@ -158,3 +179,44 @@ export function isVisible (element: Element) {
...
+}
+
+/** Sorter functon by offsetWidth/Height, used for sorting elements by their
+ *  area on the page
+ */
+export function elementAreaSorter(a: HTMLElement, b: HTMLElement): number {
+    const aArea = a.offsetWidth * a.offsetHeight
+    const bArea = b.offsetWidth * b.offsetHeight
+
+    return (aArea > bArea) ? 1 : -1
```
cmcaine: compareFunction is supposed to return 0 if the elements are equal.

Suggest renaming to compareElementArea (because it doesn't actually sort) and getting the equal condition right with: `return aArea - bArea`

src/excmds.ts
```diff
@@ -370,6 +373,103 @@ export async function reader() {
...
+input[type='month'],
+input[type='time'],
+input[type='week'],
+input[type='number'],
+input[type='range'],
+input[type='email'],
+input[type='url'],
+input[type='tel'],
+input[type='color'],
+textarea:not([disabled]):not([readonly])
```
cmcaine: Consider using the `:-moz-any()` pseudo selector (the vendor prefixed version of `:match()`) to make this shorter and to disallow disabled and readonly inputs.

```
input:not([disabled]):not([readonly]):-moz-any(:not([type]), [type='text'], [type='search'], ...)
```

the `isTextEditable` function in dom.ts also identifies `object` nodes and `[role="application"]` as editable elements, though I've never noticed an object node in the wild and there's nothing about their spec that requires them to be editable. 


document.querySelectorAll('h4:-moz-any(.browser--safari, .browser--firefox)')

johnbeard:
Fixed the compare functions, sorry about that, that was just silly!

I also made the selector change, and added `object` and `role=application`, though I didn't know where to find something to test it on. At least it's consistent with the isEditable function.

bovine3dom:
That looks neater, thanks!

I tested this and `:focusinput {-b,-p}` work well. The `-l` option didn't do what I expected it to, as I have mentioned in line. Is that by design?

johnbeard:
The -l option focuses the last input that was focused with `focusinput`, and falls back to the first one if there wasn't a "last used" one found.

What is missing is a focus listener for _any_ focus event, which is what Vimperator has (events.js:687). This means that you don't have to use the Vim binding to reach the input to get it into the "last used input" variable.

I'm open to adding that (would be useful in general for lots of things, probably), but I'm not quite sure how to go about it in Tridactyl.

bovine3dom:
Sorry @johnbeard, I didn't realise that I had to click "comment" twice to get the inline comments to show up...

The change is so small that I'll just make it myself and get it merged : )

bovine3dom:
The changes I made are here - https://github.com/cmcaine/tridactyl/commit/8f49130c6388ed1811fbdd26bf08905643b8ad86 - I do hope you don't think me too rude for doing it for you; because it was just a line it seemed to make more sense to just put it in the code than to tell you to type it out for us...

Thanks again for all your PRs so far, they've added some really useful functionality!

johnbeard:
No, that's absolutely fine. I'm not that familiar with all the WebExt APIs or Typescript/modern JS in general, so I appreciate seeing the right way to do things - I can take notes!

bovine3dom:
Oh god, don't copy from me! This repository contains the sum total of all the TypeScript and Javascript I've ever written in my life...

PR #130, opened 2017-11-24
hint: display characters in lowercase

Blatinox:
The characters are displayed in uppercase but the user must enter a
lowercase character. So let's display them in lowercase.

### DIFF
--- a/src/static/hint.css
+++ b/src/static/hint.css
@@ -3,7 +3,6 @@ span.TridactylHint {
     font-family: sans-serif;
     font-size: 12px;
     font-weight: bold;
-    text-transform: uppercase;
     color: white;
     background-color: red;
     border-color: ButtonShadow;


bovine3dom:
Having upper-case letters was a stylistic choice - we think they are more legible (it's harder to tell the difference between "hn" and "il" than "HN" and "IL" at a glance, for example),

Do you strongly disagree?

We could potentially allow the shift key to be pressed and have it still select the hint, if you think that it is very confusing.

Blatinox:
I understand that uppercase is more readable with this font. The uppercase/lowercase confused me the first time I tried tridactyl so yes, I think that allowing to be pressed would be a good idea.

I am wondering if it's better to force the characters to be in uppercase in the CSS or if it would be better to set hint characters as uppercase in the code. This way, if one day the hint chars are settable (I don't know if settings are in the roadmap), the user could choose easily between lowercase and uppercase. What do you think ?

johnbeard:
For the alphabetic hints, a user-selectable alphabet would probably make sense, as some might prefer home-row only, or all keys. Also might be good for non-latin keyboard users?

cmcaine:
When configuration exists, we'll allow customising tridactyl's CSS (for presentation and case of the hints) and choosing what keys are permitted.

Adding a better config is on the roadmap, but nobody is really working on it at the moment. Please see PR #93 for some work towards it and the problems with that approach.

bovine3dom:
hintchars are now configurable with `set hintchars`.

I suspect we should turn off uppercasein the CSS if anyone puts a capital letter in their hintchars, or there will be indistinguishable hints.

Blatinox:
Very cool for the set command !
I'm working on a patch to turn off uppercase in CSS, I'll send a pull request as soon as possible.

jcrowgey:
For the record, I got confused by the upper case letters for a while.  I had to go to the help docs, when that didn't give me anything useful, I finally thought to try lower-case.  It worked.

I don't know if we could repoen this issue or if I could start another.  In my opinion, displaying upper case when you expect people to type lower case is really quite bad.  I understand about the legibility, but unless I'm writing SQL, my default thinking is that computer input is *case-sensitive*, and I don't think I'm alone in that.  It seems like this issuer was closed because someday there may be a configuration option to change which hint chars are displayed and/or which font.  In my opinion, that doesn't really address the current issue, which is that tridactyl shows you one character but (silently) expects you to type something else.

Perhaps if you want to display upper case it would make sense to allow the user to type upper case or lower case?

bovine3dom:
> Perhaps if you want to display upper case it would make sense to allow the user to type upper case or lower case?

I'd much rather that people got confused and then figured it out than having to type uppercase characters their whole lives.

I mention it on the tutorial, I think, and I'm happy to mention it everywhere we mention hinting.

bovine3dom:
> No one has to type uppercase characters their whole lives if the input is case insensitive ;)

My point is thus: how would they find out that they didn't have to?

jcrowgey:
In my opinion the harm done by showing the user one thing but expecting them to type another is worse than the harm done by a font face where I looks like l.

One option which I believe hasn't been discussed, avoid homoglyphs when doing alpha prompts.   That is, use lower-case letters but simply skip over the "l" when choosing the letters.  Then, when a user sees an 'a', they can happily type it without any confusion.  If that still looks bad, maybe the issue of drawing lower case glyphs in a way that they're easy to see can be taken up directly.

bovine3dom:
> done by a font face where I looks like l.

It's not just this; capital letters are all roughly the same size, so they look much more even in their labels. As far as I'm aware most other vim add-ons use this same approach.

Making the UI worse for the majority of users to make the learning curve a bit less step for a minority is something I really want to avoid...

You can always style the hints yourself with Stylus, by the way.

Flimm:
Hi. I also was confused by the fact that the hints are displayed in uppercase letters, but typing uppercase letters did not work. I agree that they should be displayed in lowercase. I know some countries deliberately use lowercase letters on signs because they are more legible (different letters taking different sizes increases legibility)

I'm glad to find out that there is an option to change this behaviour (which I think should be default): `set hintuppercase false` . Thank you!

PR #131, opened 2017-11-24
URI-encode search query string

ollef:
This enables searching for e.g. 'C#'.

Also remove a now no longer valid comment.


bovine3dom:
Works well, thanks!

PR #133, opened 2017-11-25
Add hint focus mode (;;)

johnbeard:
This focuses the hinted element with excmd ":hint -;".

Includes Vimperator binding to ";;"
----
Not sure about the use of "-;" for the excmd parameter, it's hardly a mnemonic except by virtue of the historical Vimperator binding. "-f" is probably frame focus and "-F" is multiple BG open, so not sure otherwise. Perhaps there should be equivalent long-form parameters too, so an alternative excmd is something like `hint --focus`?

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -777,7 +777,7 @@ export async function quickmark(key: string) {
 //#background_helper
 import * as hinting from './hinting_background'
 
-/** Hint a page. 
+/** Hint a page.
 *
 * Pass -b as first argument to open hinted page in background.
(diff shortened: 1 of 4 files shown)


bovine3dom:
Thanks! Works nicely.

The `--focus` suggestion is sensible, but we should probably improve flags handling in the ex parser before we get too concerned about that sort of thing.

FWIW, the current plan with "multiple open" is to have another wrapper mode/function that repeatedly calls an excmd until Esc is pressed, so it probably won't be on `-F`.

Issue #82, opened 2017-11-19
support Ctrl-[ as alternative to Esc


m0tive:
In Vim you can do <kbd>Ctrl</kbd>-<kbd>[</kbd> as an alternative to <kbd>Esc</kbd>. Tridactyl doesn't seem to support this.

lhyaax:
I need the mappings too.

bovine3dom:
@cmcaine: Would you mind briefly looking at 4b3ec57? I can't quite get my head around all of the places I need to suppress `Ctrl-[` which is bound to back by default.

hppavilion1:
Last time I checked, the codebase doesn't currently support modifiers (other than shift) in bindings.

bovine3dom:
@hppavilion1 That's only in `normal` mode. Every other mode is much more hard coded, so we can write whatever we want in.

The code I referred to above works, but also triggers the Firefox back action.

bovine3dom:
This is partially possible now:

`bind --mode=insert <C-[> unfocus`
`bind --mode=input <C-[> unfocus`

Bindings for the command line are still not user configurable, I'm afraid.

PR #135, opened 2017-11-25
Add C-j and C-m keybindings

ardrabczyk:
In pentadactyl and in various line-editing libraries such as readline
used in Bash or linenoise it's possible to accept input with Control-m
and Control-j just like with the Return key.  I often use this
keybindings because they happen to be easier to use as you don't have to
move your hand to the Return key.

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -97,6 +97,18 @@ clInput.addEventListener("keydown", function (keyevent) {
             }
             break
 
+       case "j":
+            if (keyevent.ctrlKey){
+	       process()
+	    }
+            break
+
+       case "m":
+            if (keyevent.ctrlKey){
+	       process()
+	    }
+            break
+
         case "Tab":
             // Stop tab from losing focus
             keyevent.preventDefault()


bovine3dom:
Thanks for the PR!

`^j` focuses the omnibar in Firefox by default, so I quickly fixed that - hope you don't mind.

ardrabczyk:
That's OK, I didn't even know what omnibar is.

Actually I was planning on adding 2 more Control keybindings:

1. C-p and C-n as ArrowUp and ArrowDown equivalents

2. C-[ as Escape equivalent

I consider adding them in 2 separate commits as they are not really
related and would be easier to revert.  Is it OK?

bovine3dom:
The omnibar is Firefox's fancy name for the address bar : ).

1. sounds great, and yes please, as a separate commit.

2. I tried to implement myself as others asked for it in #82, but I couldn't stop Firefox from navigating backwards: https://github.com/cmcaine/tridactyl/commit/4b3ec57ec442e5dbc180caf08246d00f31736d39 - if you can figure out how to stop it, that would be fantastic!

Thanks.

bovine3dom:
Actually, I think I might have figured out 2 on my own (the magic of just leaving things alone for a while), so leave that one with me : )

bovine3dom:
On third thoughts... I made a bit of progress with ^[, https://github.com/cmcaine/tridactyl/compare/ctrlbracketasescape, but I still can't figure out how to stop it from going back in hinting mode or in insert mode, so any help would be appreciated!

ardrabczyk:
C-p was easy but C-n cannot be redefined similarly to C-[ as you
described.  It's of course easy to find complaints from other users
about this, you've probably seen all of them:

https://github.com/lusakasa/saka-key/issues/53
https://github.com/pragmader/navimacs/issues/7

They say it's a security feature baked into Firefox that prevents C-n
redefinition.  It's a shame, really.

bovine3dom:
Ah, yes, I'd forgotten about that.

cmcaine's keyboard API *might* allow us to circumvent it, eventually.

PR #143, opened 2017-11-27
Add tabfirst and tablast commands

RobbieMcMichael:
This adds two new commands along with keybindings, both of which have existed in Vimperator for about a decade.

* `g^` - `:tabfirst` - Switches to the first tab.
* `g$` - `:tablast` - Switches to the last tab.

In case we're not just trying to emulate Vimperator, both `tabfirst` and `tablast` exist in Vim so there's a good argument for having them, but the `g^` and `g$` keybindings might be a Vimperator invention (they don't clash with any existing keybindings though).

I've also done some minor refactoring to how `tabnext` and `tabprev` work to keep the code nice and simple.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -482,29 +482,49 @@ document.addEventListener("focusin",e=>{if (DOM.isTextEditable(e.target as HTMLE
 
 // {{{ TABS
 
-/** Switch to the next tab by index (position on tab bar), wrapping round.
-
-    optional increment is number of tabs forwards to move.
- */
-//#background
-export async function tabnext(increment = 1) {
+/** Switch to the tab by index (position on tab bar), wrapping round. */
+/** @hidden */
+//#background_helper
+async function tabindex(index: number) {
     // Get an array of tabs in the current window
     let current_window = await browser.windows.getCurrent()
     let tabs = await browser.tabs.query({windowId: current_window.id})
 
-    // Derive the index we want
-    let desiredIndex = ((await activeTab()).index + increment).mod(tabs.length)
-
     // Find and switch to the tab with that index
     let desiredTab = tabs.find((tab: any) => {
-        return tab.index === desiredIndex
+        return tab.index === index.mod(tabs.length)
     })
     tabSetActive(desiredTab.id)
 }
 
+/** Switch to the next tab, wrapping round.
+
+    If increment is specified, move that many tabs forwards.
+ */
 //#background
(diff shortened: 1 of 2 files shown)


bovine3dom:
This is a really nicely written PR, and it works well, so thanks!

I was just about to merge it when I realised that most of this functionality is already offered by buffer.

Wouldn't it make more sense to modify buffer to accept negative indexes?

The `g$` and `g^` binds are exactly what we would want.

RobbieMcMichael:
Cheers, I forgot about `buffer` since I don't normally use that by index.

I refactored it so that `buffer` calls `tabindex`. The reason I opted for this way round is because there's a lot of potential for confusing indices which start from zero and those that start from one. I've kept `tabindex` as a hidden helper function and I think internally we should use indices starting from zero as we're prone to doing things like modulus or someone may just start grabbing the `id` from a tab.

I think it makes sense to have `buffer` as the user facing function (since it's accepting a number or a string) and then converting from the user facing index (starting from one) to the internal index (starting from zero) as early as possible.

Obviously this is quite opinionated compared to the previous commit. We could try creating our own types to try to guard against people getting the indices confused, but I don't usually write TypeScript so I'm not sure how much use I'll be there.

I should also mention that the wrapping for `buffer` doesn't work very well for users since if the index is outside the bounds but matches the completion text (index, title or URL) of another tab, then you end up selecting that tab rather than being able to provide the index. `:buffer 0` is a really good example since instead of selecting the last buffer (as you would expect from wrapping) it just matches the 0 in 10 so you end up at the 10th tab. I don't have any good solutions for this right now, but it seemed worth pointing out.

bovine3dom:
> I should also mention that the wrapping for buffer doesn't work very well for users since if the index is outside the bounds but matches the completion text (index, title or URL) of another tab, then you end up selecting that tab rather than being able to provide the index. :buffer 0 is a really good example since instead of selecting the last buffer (as you would expect from wrapping) it just matches the 0 in 10 so you end up at the 10th tab. I don't have any good solutions for this right now, but it seemed worth pointing out.

This is a bug with the completions. I believe @cmcaine has code locally that fixes it.

The PR could be a bit controversial because of the indices, so I'll wait to see what the others say before merging it. Personally, it seems sensible (well, as close to sensible as you can get in a language that has 0-based indexing).

cmcaine:
Merged. Thank you!

cmcaine:
I also fixed completions.

PR #148, opened 2017-11-28
Add save link/img hint submode (;s, ;S, ;a, ;A)

johnbeard:
This adds the ability to save link targets or images. The save location
can be default, or the save as dialog can be invoked.

Somewhat sensible default filenames are provided.

Data URLs are also supported (though they need quite a bit of massaging
to get past the WebExt security limitations). Specificially, they need
to be round-tripped though a Blob, and must be saved from the background
context.

----

I _think_ I've done the right thing here WRT content/background/messaging, as well as the roundabout saving of data URLs.


johnbeard:
The were problems with URLs that ended in "/" causing a error in `downloads.download()` Added some tests to exercise that code (and moved the code to `url_util.ts`)

johnbeard:
Rebased over conflicting recent changes and used DOM helpers in hinting.ts.

johnbeard:
Rebased over the config framework changes

bovine3dom:
@cmcaine It's been a while, so can we just merge this? People will file issues if something is broken, and I don't think we can have many security holes since we're ultimately using Firefox's downloader.

cmcaine:
There won't be security holes. The risk is just that some files are
corrupted in obscure ways (possibly happens for files including the unicode
reserved area bytes, haven't tested yet).

I'll merge tomorrow evening after checking the last thing out.

johnbeard:
I can't say for sure that I have understood this correctly, but from the [HTML standard](https://html.spec.whatwg.org/#dom-windowbase64-atob):

>result = self.atob( data )
>
>    Takes the input data, in the form of a Unicode string containing base64-encoded binary data, decodes it, and returns a string consisting of characters in the range U+0000 to U+00FF, each representing a binary byte with values 0x00 to 0xFF respectively, corresponding to that binary data.

I interpret this as saying that you can decode _any_ binary data into a string with `atob()`, and each "Unicode char" in the string is guaranteed to be the LSB only. This could easily be invalid _Unicode data_ (say `0x80, x81`), so it's very important not to attempt to decode it _as Unicode_. On the other hand, as long as you don't do that, the binary data is "safe" in this string.

In this PR, the string is then byte-wise copied to a `Uint8Array` (which is safe as only the LSB is set by `atob()`) exactly so that it will not be corrupted if something (i.e. the following `Blob` ctor, which does interpret as UTF-8) tries to decode as a Unicode string.

johnbeard:
Rebased over `hintkill` commit

cmcaine:
Merging now. I'm convinced that the data corruption stuff won't happen. Very sorry it has taken me so long to get around to this. I'll do better next time.

Enhancement suggestion:

Many downloads are behind redirects. We should consider using fetch() or XHR to resolve the real URL. Could also catch content-disposition header and check for a 'filename' part.

Something like this would do:

```ts
    const response = await fetch(urlToSave.href, {method: 'HEAD'})
    let fileName = getDownloadFilenameForUrl(response.url)
    if (response.headers.has('Content-Disposition')) // do something else.
```

I haven't implemented it because 1) I need to do some crimbo shopping and 2) We might want to worry about http servers that don't support HEAD or something/other ways this might go wrong.

On the upside, this is the first I've seen of fetch() and it has a much nicer interface than XHR.

Issue #149, opened 2017-11-28
Support for vim's `tabonly` command


remofritzsche:
It would be amazing if `:tabonly` / `:tabo` could be used within tridactyl to close all tabs except the one currently open. This should probably ignore sticky tabs though.

Thanks a lot for considering this and keep up the amazing work!

PR #153, opened 2017-11-28
Added `tabonly` command

osleg:
Executing `tabonly` will close all tabs but active and pinned
Solves #149

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -594,6 +594,18 @@ async function idFromIndex(index?: number): Promise<number> {
     }
 }
 
+/** Close all tabs but current */
+//#background
+export async function tabonly() {
+    let tabs = browser.tabs.query({pinned: false, active: false})
+    let tabsIds = []
+    for (let id of await tabs) {
+        tabsIds.push(id.id)
+    }
+    browser.tabs.remove(tabsIds)
+}
+
+
 /** Duplicate a tab.
 
     @param index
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -594,6 +594,18 @@ async function idFromIndex(index?: number): Promise<number> {
     }
 }
 
+/** Close all tabs but current */
+//#background
+export async function tabonly() {
+    let tabs = browser.tabs.query({pinned: false, active: false})
```
bovine3dom: I think you also want `currentWindow: true` in this object.
https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/tabs/query

Without it, it closes all unpinned tabs across all windows except those which are the 'selected' tab in each window, which isn't what I would expect as a user.

bovine3dom:
Thanks for the PR! Please see the in line comment - the command didn't behave quite as I would expect it to.

remofritzsche:
That's a pretty quick turnaround time for my issue - you guys rock!

src/excmds.ts
osleg: Oh idd, I didn't even think about multiple FF windows as I olways use just one, will update asap

cmcaine:
Rebased, squashed and slightly modified.

cmcaine:
Merged. Thanks!

PR #157, opened 2017-11-28
Add yank anchor URL hint submode (;#)

johnbeard:
This yanks the page URL with the hinted elements id or title as a hash
fragment, which can be used to link to the page at that element's
location.

The second commit is a refactor using the DOM CSS query helpers. It's not required for functionality, but I think it tidies the functions in `hinting.ts` quite well?


bovine3dom:
This is great, I've really been missing this functionality. Thanks!

PR #159, opened 2017-11-29
Hint CSS: make easier to see stacked hints

johnbeard:
This has two major aspects:

 - Add a border so the extent of a hint is clear
 - Make hint BG translucent, so stacked hints can be discerned

 Also used more Tango-ish colours to soften the intense yellow.

----
Related to #69. I'm not claiming this is the best CSS ever, and CSS will be configurable in the end, but I think this makes it easier to see where hints are when they get crowded?

Note: outline is used over border so as not to change the element sizes when styled.

Here is the effect for `;p` (very stacked) and `f` (sparse):

2017-11-29-030023_1005x601_scrot
2017-11-29-030028_1005x601_scrot

### DIFF
--- a/src/static/hint.css
+++ b/src/static/hint.css
@@ -5,7 +5,7 @@ span.TridactylHint {
     font-weight: bold;
     text-transform: uppercase;
     color: white;
-    background-color: red;
+    background-color: #CC0000;
     border-color: ButtonShadow;
     border-width: 0px;
     min-width: .75em;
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks. I think this is an improvement.

I kept the red the same as the old Vimperator red (call it Stockholm syndrome), and reduced the opacity of yellow a bit more so that hints that are stacked three or so layers are still distinguishable.

If there are riots in the streets, I'll revert it back to the old Vimperator style completely...

johnbeard:
Thanks! I'm not strongly attached to any particular colours, and they'll hopefully be user-controlled one day anyway.

PR #161, opened 2017-11-29
Bound relative tabmove by identical 'pinned' value.

Koushien:
Firefox natively sorts tabs by index such that
any tab that is not an app tab
(i.e., tab.pinned == false) is placed after all
app tabs.

Currently, directly moving an [app] tab to an index where
a[n] [app] tab cannot be placed fails silently, but
Vimperator would handle such +/- tabmoves by
moving them as far in that direction as possible.
This replicates that functionality.


src/excmds.ts
```diff
@@ -686,6 +686,21 @@ export async function tabmove(index = "0") {
     let newindex: number
     if (index.startsWith("+") || index.startsWith("-")) {
         newindex = Math.max(0, Number(index) + aTab.index)
```
Koushien: I didn't actually benchmark any of this and I'd be pleased if anyone has input on how to do so efficiently, but I embed the work of constructing Number(index) here before checking if the newindex doesn't do anything (newindex == aTab.index which is equivalent to Number(index) == 0) just to avoid calling the Number constructor twice if we were to check Number(index) == 0 separately or setting newindex twice. I'd say someone who actually sends +0 or -0 to tabmove could live with this extra bit of computation before the short circuit.

src/excmds.ts
```diff
@@ -686,6 +686,21 @@ export async function tabmove(index = "0") {
...
+        if (newindex === aTab.index) return
+        let candidates = await browser.tabs.query({windowId: aTab.windowId})
+        newindex = Math.min(newindex, candidates.length)
+        candidates.sort((a,b) => {
```
Koushien: Sort of heavy-handed, but we get everything in one async query plus a mergesort, if that's still what Mozilla uses. The subsequent slice is straight-forward and we skip the need to filter separately by just iterating over the eligible items and stopping on the first match.

Hope these comments aren't excessive; I actually had this sitting for quite a while and needed to justify to myself that this was a good enough way to do it.

cmcaine:
Thanks for this PR, Koushien. I didn't have time to review it today, but I'll get to it tomorrow for sure :)

src/excmds.ts
cmcaine: Don't need to clamp this value. Array.slice does its own clamping.

src/excmds.ts
cmcaine: These comments are good. A condensed version should appear in the code.

Here's my version:

```ts
    if (index.startsWith("+") || index.startsWith("-")) {
        newindex = aTab.index + toNumber(index)
        // A non-pinned tab may not have an index less than any pinned tab.  So
        // pinned tabs can't move into the pinned area and pinned tabs can't
        // move out of it.

        // Sort tabs in ascending index order.
        const tabs = await browser.tabs.query({windowId: aTab.windowId})
        tabs.sort((a, b) => {
            return a.index - b.index
        })

        // Slice tabs so that our first choice is index 0, next best is index 1, etc.
        let candidates
        if (newIndex < aTab.index) {
            candidates = tabs.slice(newindex, aTab.index)
        } else {
            // This will also handle +0, -0.
            candidates = tabs.slice(aTab.index, newindex + 1).reverse()
        }

        // Select the first candidate that shares our pinned-ness
        for (const tab of candidates) {
            if (aTab.pinned === tab.pinned) {
                newindex = tab.index
                break
            }
        }
    } else {
```

src/excmds.ts
```diff
@@ -686,6 +686,21 @@ export async function tabmove(index = "0") {
...
+        candidates = newindex < aTab.index
+            ? candidates.slice(newindex, aTab.index)
+            : candidates.slice(aTab.index + 1, newindex + 1).reverse()
+        for (const candidate of candidates) {
+            if (aTab.pinned === candidate.pinned) {
+                newindex = candidate.index
+                break
+            }
+        }
     } else newindex = Number(index) - 1
```
cmcaine: In Vimperator, tabs move as far as they can in their zone. So `:tabmove 10` on a pinned tab will probably move it to the end of the pinned section and `:tabmove 0` on a non-pinned tab will move it to the start of the non-pinned section.

Could put your above code in a new function and call that?

Could just query by pinned tabs in the current window, find the max index and then adjust the start position accordingly?

Koushien: Oh, good catch. I probably should've tried that.

bovine3dom:
Functionality added by #892.

PR #166, opened 2017-11-30
Add Text To Speech API (and hint mode)

johnbeard:
This adds an excmd interface to the Web Speech API (TTS), which allows
users to read text out and set voice and parameters via config options.

Excmds:

    - ttsread: reads the given text or CSS selector content out
    - ttsvoices: lists available voice names (can be used in the
      'ttsvoice' option)
    - ttscontrol: stops the current reading (should also pause/resume,
      but that doesn't seem to work right now)

Config options:
    - ttsvoice: the name of the voice to use
    - ttsrate: (0.1-10)
    - ttsvolume: (0-1)
    - ttspitch: (0-2)

Also the ;r hint submode is added which reads the textcontent of the
element with the configured voice

----
I quite like being able to bind to, eg, `ttsread -c .storylink` on HN.

This uses the new config API. I'm happy to wait a bit in case that API needs to bed in, and piling options and call sites on top will make it hard to work on. 

I also don't know where to document options (as opposed to excmds). Vimperator and Pentadactyl seem to use (the same?) XML based system (c.f. options.xml in each project).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -70,6 +70,7 @@ const DEFAULTS = o({
         ";I": "hint -I",
         ";y": "hint -y",
         ";p": "hint -p",
+        ";r": "hint -r",
         ";;": "hint -;",
         ";#": "hint -#",
         "I": "mode ignore",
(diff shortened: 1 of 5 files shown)


bovine3dom:
Thanks!

As I said on Matrix: I've been documenting the options in the docstrings of the related excmds.

PR #192, opened 2017-12-02
[RFC] `Tab` to cycle throught inputs, fix #191

### LINKED ISSUE #191: [Feature request] `Tab` to cycle through available inputs after `gi` to focus input
This is a feature from Vimium, which I find is really convenient to input something in web pages(for searching, login, etc.).

It highlights all inputs, make them easy to find. See in action:

output

azuwis:
Not sure about the implementation, comments are welcome.

Keybinding has problem, `Tab` still focus on next any element, and then focus on next input element.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -471,6 +471,8 @@ let LAST_USED_INPUT: HTMLElement = null
  *
  * @param nth   focus the nth input on the page, or "special" inputs:
  *                  "-l": last focussed input
+ *                  "-n": input after last focussed one
+ *                  "-N": input before last focussed one
  *                  "-p": first password field
  *                  "-b": biggest input field
  */
@@ -493,6 +495,22 @@ export function focusinput(nth: number|string) {
         // will need to serialise the last used input to a string that
         // we can look up in future (tabindex, selector?), perhaps along with
         // a way to remember the page it was on?
+    }
+    else if (nth === "-n" || nth === "-N") {
+        // attempt to find next/previous input
+        let inputs = DOM.getElemsBySelector(INPUTTAGS_selectors,
+            [DOM.isSubstantial]) as HTMLElement[]
+        if (inputs.length) {
+            let index = inputs.indexOf(LAST_USED_INPUT)
+            if (nth === "-n") {
+                index++
+            } else {
+                if (LAST_USED_INPUT === null) index = -1
+                else index--
+            }
+            index = index.mod(inputs.length)
+            inputToFocus = <HTMLElement>inputs[index]
+        }
     }
     else if (nth === "-p") {
         // attempt to find a password input
@@ -522,7 +540,10 @@ export function focusinput(nth: number|string) {
                                          index, [DOM.isSubstantial])
     }
 
-    if (inputToFocus) inputToFocus.focus()
+    if (inputToFocus) {
+        inputToFocus.focus()
+        if (nth === "-l") input_mode()
+    }
 }
 
 // Store the last focused element
@@ -1265,6 +1286,21 @@ import * as gobbleMode from './parsers/gobblemode'
 //#background
 export async function gobble(nChars: number, endCmd: string) {
     gobbleMode.init(nChars, endCmd)
+}
+
+// }}}
+//
+
+// {{{ INPUT mode
+
+//#content_helper
+import * as inputMode from './parsers/inputmode'
+
+/** Initialize input mode.
+*/
+//#content
+function input_mode() {
+    inputMode.init()
 }
 
 // }}}
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -67,6 +67,7 @@
 
 import * as Messaging from "./messaging"
 import {l} from './lib/webext'
+import state from "./state"
 
 //#content_omit_line
 import "./number.clamp"
@@ -80,8 +81,6 @@ import {messageActiveTab} from './messaging'
 
 //#background_helper
 import "./number.mod"
-//#background_helper
-import state from "./state"
 //#background_helper
 import {ModeName} from './state'
 //#background_helper
@@ -489,12 +488,16 @@ export function focusinput(nth: number|string) {
     if (nth === "-l") {
         // try to recover the last used input stored as a
         // DOM node, which should be exactly the one used before (or null)
-        inputToFocus = LAST_USED_INPUT
+        if (LAST_USED_INPUT) {
+            inputToFocus = LAST_USED_INPUT
+        } else {
+            // Pick the first input in the DOM.
+            inputToFocus = DOM.getElemsBySelector(INPUTTAGS_selectors,
+                [DOM.isSubstantial])[0] as HTMLElement
 
-        // failed to find that? - look up in sessionStorage?
-        // will need to serialise the last used input to a string that
-        // we can look up in future (tabindex, selector?), perhaps along with
-        // a way to remember the page it was on?
+            // We could try to save the last used element on page exit, but
+            // that seems like a lot of faff for little gain.
+        }
     }
     else if (nth === "-n" || nth === "-N") {
         // attempt to find next/previous input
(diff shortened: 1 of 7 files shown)


src/excmds.ts
```diff
@@ -462,6 +462,7 @@ let LAST_USED_INPUT: HTMLElement = null
  *
  * @param nth   focus the nth input on the page, or "special" inputs:
  *                  "-l": last focussed input
+ *                  "-n": next to last focused input
```
Koushien: > next to last
A bit ambiguous between "second to last" and "(the) input (occurring/found) after last focused input".

src/parsers/insertmode.ts
```diff
@@ -6,6 +6,8 @@ export function parser(keys){
     if (! hasModifiers(keys[0])) {
         if (keys[0].key === 'Escape') {
             return {keys: [], ex_str: 'unfocus'}
+        } else if (keys[0].key === 'Tab') {
+            return {keys: [], ex_str: 'focusinput -n'}
```
Koushien: This functionality is invoked whenever in 'Insert Mode' (virtually whenever an input field is selected) and not just after 'focusinput -l'. Was this your intention? Because I maintain a hint mode selecting only input fields is more suited for that and this is a change in core behavior that I feel deserves a setting.

bovine3dom:
Please see my inline comment for fixing tab.

I agree with Koushien that this behaviour might confuse users. We could hide it behind a configuration option or have it only happen when `gi` is pressed. The former is easier and probably makes more sense.

src/excmds.ts
azuwis: How about "input after last focused one"?

Or would you please suggest a clear explanation of `focusinput -n`? Not a native English speaker here.

azuwis:
Updated with:

1. Change the description of `focusinput -n`
2. Add a userconfig `tabfocusnextinput`, please suggest a better config name

@bovine3dom Still can't get preventDefault/stopPropagation to work, get `preventDefault not is a function`.

One thing I noticed is commandline_frame.ts runs as content script, while insertmode.ts runs as background script.

azuwis:
Another update, add input mode.

Now, `Tab` to cycle through inputs will be enabled only after `gi` pressed.

src/excmds.ts
```diff
@@ -1242,6 +1256,21 @@ export async function gobble(nChars: number, endCmd: string) {
...
+// {{{ INPUT mode
+
+//#content_helper
+import * as inputMode from './parsers/inputmode'
+
+/** Initialize input mode.
+*/
+//#content
+export async function input_mode() {
+    inputMode.init()
```
cmcaine: Put this logic in `function mode()`, please.

src/parsers/inputmode.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function parser(keys: MsgSafeKeyboardEvent[]) {
+    const key = keys[0].key
+
+    if (key === 'Escape') {
+        state.mode = 'normal'
+        return { keys: [], ex_str: 'unfocus' }
+    } else if (key === 'Tab') {
+        return { keys: [], ex_str: 'focusinput -n' }
```
cmcaine: No shift-tab?

src/excmds.ts
```diff
@@ -443,6 +443,7 @@ let LAST_USED_INPUT: HTMLElement = null
  *
  * @param nth   focus the nth input on the page, or "special" inputs:
  *                  "-l": last focussed input
+ *                  "-n": input after last focussed one
```
cmcaine: No reverse option?

cmcaine:
The parsers run in the background script and receive copies of the keyevents, not the real keyevents, which is why you can't call preventDefault etc on them. We did this partly as an experiment becuase Mozilla was saying that the new keyboard API wouldn't allow us to block keyboard events asynchronously.

We will probably move more of this logic to the content script in the future, but I see from the diff that you worked out how we're doing it at the moment.

azuwis:
Updated with `Shift+Tab`.

@cmcaine Do you mean replace `input_mode()` call with `mode("input")`? But `mode()` is a background function, and `focusinput()` is a content function. I'm not sure how to do that.

cmcaine:
> Do you mean replace input_mode() call with mode("input")?

Yes. 

> But mode() is a background function, and focusinput() is a content function. I'm not sure how to do that.

You don't call focusinput from input_mode(). In fact, input_mode just calls inputMode.init(), which just changes state.mode, so `:mode input` does the same as `:input_mode` at the moment.

Do you want to be able to call input_mode from focusinput? Why?

How do you imagine this whole thing working? `bind gi composite mode input | focusinput` ?

src/excmds.ts
```diff
@@ -467,6 +469,26 @@ export function focusinput(nth: number|string) {
...
+        }
+    }
+    else if (nth === "-N") {
+        // attempt to find previous input
+        let inputs = DOM.getElemsBySelector(INPUTTAGS_selectors,
+            [DOM.isSubstantial]) as HTMLElement[]
+        if (inputs.length) {
+            let index = inputs.indexOf(LAST_USED_INPUT) - 1
+            if (index <= -1) index = inputs.length + index
+            inputToFocus = <HTMLElement>inputs[index]
```
cmcaine: You shouldn't just copy and paste code like this: it obscures that two code paths are basically the same and increases the chance that bug fixes or other changes to one branch won't be applied to the other. Better to find the points of commonality and abstract or reuse.

In this case you can just use the mod function to replace this line: `if (index <= -1) index = inputs.length + index` and then the only difference is the sign in `let index = inputs.indexOf(LAST_USED_INPUT) - 1`.

The minimal change would be something like this:

```ts
    else if (nth === "-n" || nth === "-N") {
        // attempt to find next input
        let inputs = DOM.getElemsBySelector(INPUTTAGS_selectors,
            [DOM.isSubstantial]) as HTMLElement[]
        if (inputs.length) {
            let index = inputs.indexOf(LAST_USED_INPUT)
            if (nth === "-n") index++
            else index--
            index = index.mod(inputs.length)
            inputToFocus = <HTMLElement>inputs[index]
        }
    }
```

src/excmds.ts
```diff
@@ -467,6 +469,26 @@ export function focusinput(nth: number|string) {
         // we can look up in future (tabindex, selector?), perhaps along with
         // a way to remember the page it was on?
     }
+    else if (nth === "-n") {
+        // attempt to find next input
+        let inputs = DOM.getElemsBySelector(INPUTTAGS_selectors,
+            [DOM.isSubstantial]) as HTMLElement[]
+        if (inputs.length) {
+            let index = inputs.indexOf(LAST_USED_INPUT) + 1
```
cmcaine: What happens if LAST_USED_INPUT is null?

cmcaine:
Thanks again for your PR. Please let me know if you find this review process tiresome - I want to give you the opportunity to fix the issues so that you feel greater ownership of the work; so you can challenge me over my criticisms; and so you have the opportunity to learn from mistakes, but I know it can be annoying if you just want an issue fixed.

So... feedback appreciated on my approach here :)

src/excmds.ts
azuwis: If LAST_USED_INPUT is null, both `focusinput -n` and `focusinput -N` will focus first input. I think that's fine.

azuwis:
Updated, combine `focusinput -n` and `focusinput -N` to single block.

Still not sure about what to do about `input_mode()`, if replace the call with `mode("input")`, will get:

    ERROR in [at-loader] ./src/excmds_content.ts:450:27 
        TS2304: Cannot find name 'mode'.

Not familiar with WebExtension development and Tridactyl code base at all, and in facts, this is my first time writing typescript.

cmcaine:
Thanks for your PR. I tidied it up a bit and removed the input_mode function, in hindsight I should also have removed the init function.

To call the init function from the content script, one would have to use the messaging API either explicitly or implicitly with the excmds macros and neither is very neat. The init function doesn't do anything and state transition logic will change soon (TM), so I just skipped straight to setting the state.

This PR introduces the `vimium-gi` config option which defaults to true. If true, focusinput calls that cause an input to be focused will also change to the new input mode.

azuwis:
Thanks for the explanation and clean up.

PR #193, opened 2017-12-02
Add chat badges to help page

johnbeard:
Taken from https://github.com/blog/273-github-ribbons

This is under an MIT license, so it's compatible with Tridactyl's Apache?

----

I think this is useful for users, as it lands them in a place they can get to the issues, chat badges, etc.


Koushien:
I find the Fork Me! ribbons kitschy. The new tab page already explicitly tells users (who may not necessarily be familiar with Github) to visit the repo if they have issues/requests. If anything, I'd say condensing the new tab page and adding the chat badges directly would be more helpful.

johnbeard:
I'm not that attached to the ribbon, it was just an idle thought.

It seems the new tab page comes from `static/newtab.md`, and the `:help` preface is from `excmds.ts`. AFAICT, the main difference is the justification for the newtab page.

Koushien:
I was reading your reply and tried `:help` on the machine I happened to be on and sure enough I saw the section from `excmds.ts`. I tried again on a new profile on this machine and no dice. I pulled master and rebuilt and nothing. Then I realized what was going on.

```
Error: The output directory "/.../tridactyl/generated/static/docs" exists but does not seem to be a documentation generated by TypeDoc.
Make sure this is the right target directory, delete the folder and rerun TypeDoc.
Error: Documentation could not be generated due to the errors above.
```
Particularly the last three lines. I knew one of the scripts was doing *something* to the generated docs and thought those messages were to be expected for the longest time (I might've missed a memo), but I `rm -r generated/`, rebuilt with those errors gone and now see what everyone else sees on `:help`. A lesson for me.

Back to the topic, I guess the redundancy on the help page makes sense for users who deactivate the Tridactyl's new tab; add them to both places?

cmcaine:
There's no obvious nice way to share content between newtab.md and the module doccomment in excmds.ts, hence the repetition.

I don't really like the fork me on github banner, but it does demonstrate an error with our hint positioning code. I also don't think we should link to external resources from tridactyl, where that can be avoided: it makes our users (slightly) easier to track.

Changes to the help page are welcome, though, and it's good to highlight the chat channel. Should also list IRC, though: we get far more IRC users than on gitter.

johnbeard:
Good point with the external resource. I actually tried an inline SVG one first, but I went for the "traditional" one in the end. I've removed it from the PR, 'tis a silly thing anyway!

I've added an IRC badge to the :help preface with a link to the webclient, and I've added the badges to the repo, to avoid more external loads.

PR #195, opened 2017-12-02
Sanitize

caputchinefrobles:
Adding very basic function for sanitizing private browsing data. 

Still not using option "since" for timespan range. 

PR to check opinion if it is in the right place in the code as well as hearing for suggestions on the process to communicate with the user regarding progress of the sanitize work (messages in the command line??? Maybe not a good idea).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -829,28 +829,58 @@ async function getnexttabs(tabid: number, n?: number) {
 
 /** Clear private browser data. 
 *
-* Item could one or multiple items from the list following (hopefully): 
+* Item could be one or multiple items from the list following:
 *
 * - cache
 * - cookies
+* - downloads
+* - passwords
 * - formdata
 * - history
-* - marks
 * 
-* 
+* E.g: To clear cookes and browsing cache
+* :sanitize cookies cache
+*
+* E.g: To sanitize all browsing cache (from the listed above)
+* :sanitize
 */
 //#background
-export async function sanitize(item?: string){
-    //switch(item) {
-    //    case 'history':
-    //        //await not resolving deletingAll
-    //        let deletingAll = browser.history.deleteAll()
-    //        console.log("Deleting history...")
-    //        deletingAll.then(function(){console.log("History deleted.")})
-    //    default:
-    //        console.log("Sanitize everything on default?")
-    //}	
-    console.log("Tdçlgkajdçglak jdçglakjdgç alkdjgça")
+export async function sanitize(...strarr: string[]){
+
+    /* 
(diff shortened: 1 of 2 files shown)


caputchinefrobles:
It seems like there's already work being done after this feature.... Is that possible to cancel the PR?

bovine3dom:
Thanks so much for going to the trouble of making this PR!

However, I think you may not have seen PR #179 which implements much of this functionality.

Is there any difference between the implementations?

bovine3dom:
Ah, sorry, I see you have just noticed it.

Please do check to see if there is anything you have done differently or better than the previous PR as I do not want your efforts to go to waste. If there is truly nothing salvageable, we can close the PR.

Thanks again for your interest in contributing, we would be very happy to receive help in other areas too (particularly for example in making sure `winopen -private` is not stored).

caputchinefrobles:
No problem... happy to help. 

I'm just not sure about the github flow. Is there a way of pushing a branch directly into cmcaine/tridactyl without having to fork first?

I assume this way would be easier to see if there's a feature already being worked on because of the visible branches. Or those are only the ones that were accepted from PRs?

Thanks in advance.

Koushien:
It would be dangerous, and very messy, to allow anyone to push into this repo directly. I've been told there are git helpers that simplify the process, but I'm rather certain they just fork quietly in the background([0]). Duplication of effort is unfortunately a reality of development and all the processes (e.g., tickets, Kanban boards) project managers implement can only do so much to mitigate it by themselves; it really relies on pro-active communication in the end (do hang out in the Matrix channel if you like). Even just reading off branches of all the forks wouldn't help you know if someone has changes they haven't pushed yet. The ones on this repo, for instance, are either the maintainers' WIP or for staging before the merge into master.

[0]: https://github.com/github/hub/issues/1553#issuecomment-327034673

bovine3dom:
You also could comment on or file an issue to let others know you are working on something.

cmcaine:
FWIW, winopen -private is not stored.

cmcaine:
Re workflow:

We encourage contributors to say hello in our chat channel or open an issue to talk about what they'd like to do. We can let you know if anyone else is working on the same thing or advise you of any other issues.

Then just make a feature branch off master, do your thing and then submit a PR. Don't worry too much about merge conflicts: we'll just sort them out for you, but if you do want to work them out just run `git fetch && git rebase origin/master` (or `git merge` if you prefer) then work through any conflicts and re-test that your code works.

I've removed a few branches that aren't being worked on, so maybe that's less confusing :)

PR #199, opened 2017-12-03
Improve the appearance on AMO

Skillmon:


bovine3dom:
Thanks so much for this @Skillmon.

I'm a bit concerned that amo.md isn't particularly welcoming to newcomers - most people (even most Vim users) won't know what ex mode is, for example.

I think we can very quickly persuade many Vimperator users by just saying "it's like Vimperator", and that potential first converts (maybe even non-Vim users) should be who the AMO flavour text is aimed at.

Also, the hinting screenshot is now a little out of date: the first few hints on a page are now always single characters.


Please don't feel any pressure to change any of this - we're very grateful for your contribution : )

Skillmon:
Reworked that description. Now I feel like a bad American ad-texter.

bovine3dom:
That's much better, thanks.

We ought to advertise `:help` and, if we're going to tell people about ignore mode on `I`, we really ought to tell them that they can leave it with `Shift-Esc`.

bovine3dom:
Thanks! I'll put it on the AMO today.

Koushien:
From the Vimperator exit poll in #203 there's a strong mention for improved accessibility
>Vim style extensions could motivate improved accessibility within Firefox and other programs. This should be supported and embraced as it could help many who struggle with program UI.

and a couple (maybe more that are indirectly worded) votes for reduced strain
>I will absolutely refuse to upgrade my browser unless I can be assured that there is a way to operate it without the rodent. Using the rodent is slow and time-wasting, and gives my wrists an aneurysm. Please consider future projects to support rodent-free browsing; you will definitely have my support!

>I started using Vimperator to help with my repetitive strain injury caused by using the mouse. I tried several other plugins but Vimperator was the best by far. It's unfortunate to see it go.

and talking about those can go a way in getting people to see the value of grokking tridactyl. In fact I feel the ideal copy would condense the idea from https://stackoverflow.com/a/1220118/5255922 that you can efficiently define your browsing habits in terms of simple Tridactyl "statements" (commands). I'm aware that is no small feat especially given AMO limitations (do we know what they exactly are, for reference?) and the attention a given AMO user may have; I'm not a pro copywriter either by any stretch. The examples are really good, but I feel it's selling the project short having seemingly random (to the layman) keys at the forefront and listing items such as `- bind any supported command or commands to the key (sequence) of your liking` at the end as if they aren't what really set Tridactyl apart from a convenient hotkey wrapper. The project may still be a ways away from custom commands and the like that are also mentioned with frequency in the survey (and the md should mention the intention to specifically implement such features), however Tridactyl is fast becoming featured and stable enough for everyday browsing to start pushing the benefits of mastering Tridactyl.

This may probably be more appropriate in a separate "Improving amo.md" issue and I don't mean to devalue @Skillmon's contribution (_Thanks_, really). I just don't think I collected my thoughts enough to express actionable items for an issue proper much less confidently open a PR myself. This is more like pontification to me.

Skillmon:
@Koushien: I thought that the point was to seduce new user to use tridactyl. In the end the cool stuff might be possible with the high customizability but a new user would most likely discover and try that later and start with the default stuff. So I think to display what is possible by *default* is the selling point to new users. Feel free to change stuff. I'm not a salesperson :)

bovine3dom:
Skillmon's stuff is a lot better than what is currently there, I think.

I still haven't got around to properly proof reading it yet, which is why I haven't updated the AMO.

I agree it can be improved, and also agree that I don't know how to improve it.

Feel free to make a new issue or hijack #162. Maybe someone will have ideas : )

Koushien:
@Skillmon: Oh I definitely think your additions are, unambiguously, an improvement. I just started writing about considering adding a mention to accessibility over time and that led me to write about what I believe the "final draft" of the copy should cover. I couldn't think of a better place to add these comments so I just submitted them here. I'll take @bovine3dom on his offer and make a request for him to append "/ AMO description" or "/ AMO copy" to #162.

rscircus:
@Skillmon, that history-screenshot...

Skillmon:
@rscircus Isn't it great? Totally xkcd inspired (don't ask me which xkcd it was, but it pointed me towards that wiki entry).

PR #200, opened 2017-12-03
Hints: do not force uppercase display of hintchars

Blatinox:
Those 2 commits allows to use both uppercase and lowercase hintchars.

Please see pr #130 for the origin of this pr.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -105,7 +105,7 @@ const DEFAULTS = o({
     "newtab": undefined,
     "storageloc": "sync",
     "homepages": [],
-    "hintchars": "hjklasdfgyuiopqwertnmzxcvb",
+    "hintchars": "HJKLASDFGYUIOPQWERTNMZXCVB",
 
     "ttsvoice": "default",  // chosen from the listvoices list, or "default"
     "ttsvolume": 1,         // 0 to 1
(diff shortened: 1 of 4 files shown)


bovine3dom:
Thanks for this PR!

I would prefer the CSS to remain uppercase unless an uppercase letter is in `hintchars`. Is that something you would be able and happy to do?

Blatinox:
What's the interest of keeping the uppercase property in CSS ? My main idea is to add the possibility to display hintchars in lowercase, not only to allow both lowercase and uppercase in `hintchars`.

bovine3dom:
As I said in #130, I think the hints are easier to read in all caps. It's how most similar extensions do it.

I'm happy for you to add a setting that turns off the uppercase CSS in all cases.

Blatinox:
I understand, that's why by default my PR keeps the uppercase. It only moves it away from the CSS so it's easier to change characters.

What do you think about the first commit of this PR that allows to press Shift while entering the hint key ?

bovine3dom:
@Blatinox sorry for the delay in getting back to you. I hadn't realised you had replied.

I'm happy for this to go in with the proviso that hintchars should be displayed in uppercase when there are no uppercase hintchars, but as soon as there is an uppercase hintchar they are displayed in mixed case.

bovine3dom:
Closing as #481 / some theme API will allow people to style the hints however they please.

PR #206, opened 2017-12-04
Allow levelled logging for debug

johnbeard:
This is useful for suppressing spew, but still allowing easy debugging
when needed, without rebuilding.

To use, use excmd:

     setlogginglevel <prefix> <level>

Where the level is one of:

    - never:  never log
    - error
    - warning
    - info
    - debug:  most verbose

Output is directed to console methods "error", "warn", "log", "debug".

-----

I've added a few uses, in `{message,commandline_frame,controller}.ts`, and with the defaults at WARNING, this quietens down the console. It used to lag for me on backspace with the console open, and now it doesn't.

Advice welcome on what to do about the `{config,logging}.ts` imports, as currently importing `logging` in `config.ts` is circular, so I didn't use the enums, just the numbers. Split to another file?

### DIFF
--- a/src/commandline_content.ts
+++ b/src/commandline_content.ts
@@ -1,5 +1,7 @@
 /** Inject an input element into unsuspecting webpages and provide an API for interaction with tridactyl */
 
+import * as Logging from './logging'
+
 /* TODO:
     CSS
     Friendliest-to-webpage way of injecting commandline bar?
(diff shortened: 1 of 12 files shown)


cmcaine:
I might refactor this as an OOP thing to reduce code re-use...

PR #207, opened 2017-12-05
excmds: add "quit" and "q" commands

peff:
Closing the final tab via "tabclose" doesn't actually close the window. Instead, we just get an empty placeholder tab.
Let's instead add a "quit" command which can be used to quit the current tab, and if it's the last tab in the window,
close the _window_.

This is similar to how vim (and pentadactyl) work, which lets you ":q" your way through any open buffers.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -766,6 +766,23 @@ export async function qall(){
     windows.map((window) => browser.windows.remove(window.id))
 }
 
+/** Close the current tab, and close the current window if no tabs remain. */
+//#background
+export async function quit() {
+    let tabs = await browser.tabs.query({currentWindow: true})
+    if (tabs.length > 1) {
+        tabclose()
+    } else {
+        winclose()
+    }
+}
+
+/** Convenience shortcut for [[quit]] */
+//#background
+export async function q() {
+    quit()
+}
+
 // }}}
 
 // {{{ MISC


bovine3dom:
@peff what version of Tridactyl and which O/S are you using?

For me, tabclose will close the window if it is the last tab in the window.

peff:
@bovine3dom It's the tip of "master" (cmcaine/tridactyl@4ed2031d693097701abb13d9a78eca9851625b27), but the behavior was the same with earlier versions. It's Firefox 57.0.1 on Debian. The behavior is the same whether I `:tabclose`, use `d`, or use Firefox's own `^W` hotkey: closing the last tab gets me a new-tab window (which has the tridactyl content, but that's expected).

It has been that way long before I started using tridactyl. I wonder if it's a funny about:config I have set.

peff:
> I wonder if it's a funny about:config I have set.

Hmph. Indeed, my setting of `browser.tabs.closeWindowWithLastTab` was false. I have no recollection of setting that, so probably it was done literally years ago while playing with vimperator/pentadactyl config (or possibly even tweaked indirectly by one of those extensions). So with that fixed, `d` does what I want.

I do still have a slight preference to provide a `:q` for symmetry with vim (and because years of using other vim-likes has trained my fingers on it), but it probably ought to just be a synonym for `tabclose`. What do you think?

bovine3dom:
Yeah, `:q` as a synonym for `:tabclose` sounds sensible to me.

You could also do `:qa` for `:qall` if you're feeling generous : )

Thanks!

peff:
Thanks. Closing this; I'll put the new approach in its own PR.

PR #209, opened 2017-12-06
Blacklist default newtab url from history completions

azuwis:
Was: Remove moz-extension:// items from history completions

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -497,8 +497,12 @@ export class HistoryCompletionSource extends CompletionSourceFuse {
     }
 
     private async scoreOptions(query: string, n: number) {
+        const newtab = browser.runtime.getManifest()['chrome_url_overrides'].newtab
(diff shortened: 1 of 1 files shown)


bovine3dom:
I think blacklisting the new tab page is sensible. Why do you want to hide everything else?

azuwis:
@bovine3dom Whenever I type `:tabopen tridactyl`, moz-extension://7dd100a5-eb70-48d9-b963-0d92d21d3cc9/static/newtab.html is always the first option. For me it's annoying, I want the http(s) link, more specifically https://github.com/cmcaine/tridactyl/, but not a Firefox internal link.

If you think the moz-extension:// links are important, I'll update the pull request with only changes to topSites.

bovine3dom:
Thanks for this. It works well.

Sorry it took us so long to get around to merging it.

PR #211, opened 2017-12-06
excmds: add `q`, `qa`, and `quit` synonyms

peff:
Other vim-likes map `q` and `quit` to `tabclose`, so let's do the same to appease their muscle memory.

While we're at it, we can add `qa` as a shortcut for `qall` (vim also supports this shorthand).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -732,6 +732,18 @@ export async function undo(){
     }
 }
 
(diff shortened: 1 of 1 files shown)


peff:
This is a rework of #207, after @bovine3dom helpfully pointed out my bogus config. :blush:

bovine3dom:
Thanks!

PR #216, opened 2017-12-06
Include default bind for tabonly function

caputchinefrobles:
Add in config.ts a sugestion for default binding to the orphaned
function tabonly (close other tabs)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -47,6 +47,7 @@ const DEFAULTS = o({
         "H": "back",
         "L": "forward",
         "d": "tabclose",
+        "D": "tabonly",
         "u": "undo",
         "r": "reload",
         "R": "reloadhard",


bovine3dom:
I'm slightly worried that D is a bit easy to press by accident, for example by missing F.

Is there a precedent for this? I would otherwise prefer that it was a multiletter bind.

As an aside - are you aware that you can execute the command with `:tabonly<CR>`?

Thanks for your interest in this!

caputchinefrobles:
Not particularly...  It was just as a suggestion for that. I tried to think of multichar bind but felt like wrong not following the idea "command char + movement char", but it was not possible since "d" could not be used as command char ("t" as well), then using "D" seemed OK. 

Yes i'm aware of the possibility of running tabonly from command line. Maybe with the command completion, tabonly would be as fast as ":tab<TAB>".

Issue #223, opened 2017-12-08
Improve modifier support on buffer command


behrmann:
In pentadactyl pressing `b` would show the list of buffer, as it does in tridactyl. Pressing `<num>b` would jump directly to tab `<num>`. In tridactyl it prefills the commandline with `buffer <num>`. This is sufficiently similar, but has its own bugs. On my current window with 19 tabs pressing `6b` fills the commandline with `buffer 6`, but highlights tab 16 and pressing enter jumps to tab 16.

Directly jumping to tab `<num>` when pressing `<num>b` would be a nice addition.

behrmann:
Thanks for the quick reaction. This is mostly fixed on `master` now, the only missing part, that pressing `2b` should not only prepopulate the commandline with a jump to buffer 2, but also do it right away.

bovine3dom:
I think we should just make counts automatically execute commands.

cmcaine:
@Koushien you were doing something to do with this, did you get anywhere?

Koushien:
Huh, I thought you were talking about the options refactor, then I realized I did say I was to take a look at this specifically. Here's a naive implementation: https://github.com/Koushien/tridactyl/commit/3b899b552b7d0936586751601f861d1fa13334fe. It works as you'd expect with the default binds, but `fillcmdline (tab|win)?open #count` makes even less sense. Sorry for blocking.

bovine3dom:
Described behaviour - `6b` filling the command line with `buffer 6` no longer happens.

Did it get broken in keyseq, @cmcaine?

bovine3dom:
Closing in favour of #466.

PR #226, opened 2017-12-09
Let tabs() accept filters, ameliorates #223

Koushien:
> In Vimperator, buffers(...args) displays a
> filtered buffer list with ...args as the query.
> This commit has tabs() resemble _that_ functionality
> making use of fillcmdline_notrail.
> 
> Exploiting the fact that providing the buffer
> completions system with a single number n and no
> trailing space completes only the tab at index
> (n % #tabs), using this in normal mode grants us
> a count-supported tabIndexSetActive() that ties
> "gt" in keypresses while retaining the usual
> "fillcmdline buffer" behavior if no counts are
> supplied.

I say it only ameliorates #223 because behavior still diverges from Vimperator/Pentadactyl, but I think this is acceptable at present.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -64,7 +64,7 @@ const DEFAULTS = o({
         "M": "gobble 1 quickmark",
         "xx": "something",
         // "B": "fillcmdline bufferall",
-        "b": "fillcmdline buffer",
+        "b": "tabs",
         "ZZ": "qall",
         "f": "hint",
         "F": "hint -b",
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -272,7 +272,7 @@ export function open(...urlarr: string[]) {
 
 
 /** Go to your homepage(s)
- 
+
     @param all
         - if "true", opens all homepages in new tabs
         - if "false" or not given, opens the last homepage in the current tab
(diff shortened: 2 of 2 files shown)


bovine3dom:
I think 8c32dcb fixes this more nicely. I had only scanned this PR when I made that commit - I thought this PR was aiming to fix the "`3b` should take me directly to the third tab" issue. Sorry for gazumping you!

Regarding #223, perhaps `fillcmdline` should just execute the ex cmd if its last argument is a count?

Koushien:
I'm of the mind that `fillcmdline` shouldn't accept counts to begin with since `12 fillcmdline foo` doesn't make sense without a `<CR>` to interpret - unless someone somewhere really wants other functionality like `foofoofoo...foo` (12 times). I also feel like the committed change is currently too powerful, if someone starts filtering with a number (e.g., the section of the text they were looking at, the name of the song/movie/painting/game/etc., a particular issue or ticket, etc.) but doesn't want the tab at the index, they have to complete a second argument _and_ append another whitespace to start seeing other tabs. They could also immediately input any arbitrary non-number character, but that's hardly intuitive. If I could propose another solution, perhaps completions should select the tab at the index for single number queries, but don't exclude showing other matching tabs? I can begin looking into that, if desired.

That aside, would you still be interested in retaining the optional filtering of this patch? It diverges from Vimperator where, if the results are non-empty, it would enter a view of the commandline and you could `;y` to yank or `;o` to open links corresponding to the buffers via hinting with `;q` or `ESC` to exit the view. It only supported those operations for one successful hint before closing and think it would be worthy to expand on that, eventually, with the usual hinting conventions. Capability to open a link in the background with `F`, for example.

bovine3dom:
@Koushien I'm struggling to remember what this PR was trying to do. Is it still necessary?

PR #227, opened 2017-12-09
Implement ^,$ navigating to 0%,100% in x-direction

Koushien:
> This is made possible by also allowing the second
> argument to be "x" or "y". We supply "y" as the
> default to retain the current, more commonly
> desired behavior.

We can totally support `<num>[(x|y)]` for percentage scroll, but at that point I feel we should accept `,` as a delimiter, i.e., `scrollto <num>,<num>` -- I thought to KISS for now.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -44,6 +44,9 @@ const DEFAULTS = o({
         "l": "scrollpx 50",
         "G": "scrollto 100",
         "gg": "scrollto 0",
+        "$": "scrollto 100 x",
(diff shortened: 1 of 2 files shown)


cmcaine:
Thanks, Koushien!

cmcaine:
Are you still looking at improving count support? `10l` is an example of a key sequence that doesn't work properly with the current logic. Might make a good test case.

Koushien:
Definitely, I'll have a look-see.

PR #238, opened 2017-12-17
Add hint kill submode (;k)

johnbeard:
This submode deletes the selected hinted item. Useful for removing
overlay blockers or even unwanted/annoying content.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -73,6 +73,7 @@ const DEFAULTS = o({
         "F": "hint -b",
         ";i": "hint -i",
         ";I": "hint -I",
+        ";k": "hint -k",
         ";y": "hint -y",
         ";p": "hint -p",
         ";r": "hint -r",
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1301,6 +1301,7 @@ import * as hinting from './hinting_background'
         - -r read an element's text with text-to-speech
         - -i view an image
         - -I view an image in a new tab
+        - -k delete an element from the page
         - -; focus an element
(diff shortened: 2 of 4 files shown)


bovine3dom:
Works great, thanks!

Sorry that cmcaine and I have been a little more inattentive than usual; both of us are holidaying.

A nice addition to this might be a way to undo killing a hint (I managed to kill the wrong one in my testing and lost the entire body). An obvious extension would be storing the elements to kill and automatically killing them, but I suspect that could be slightly tricky; we'd need aucmds for a start. (Mostly thinking out loud here - not a demand for more work!)

cmcaine:
I think the best way to do this long term would be asking the ublock people
to expose an API for us to call, then just give the selector to them. I've
posted on their bug tracker.

iago-lito:
> asking the ublock people to expose an API for us to call [...] I've posted on their bug tracker.

@cmcaine Neat. That's the right thing to do I'm sure :) Would you cross-link to this post?

cmcaine:
Check the web view of this issue and it's linked above.

johnbeard:
I think deferring to a specialist app is a good idea.

Maintaining an undo stack of killed (and/or modified if we get new "page mutation" functions) elements would not be too hard, as long as they are unwound in order. Otherwise, if you kill an element, then its parent, and try to restore the original one, where does it go?

However, there a still issue like what happens if it's a SPA web app and the DOM has changed a lot since the killing, but we're on the same "page", etc.

Could also have (or redo this one) a "hide element" submode which just slaps `display: none !important` on the element?

Issue #109, opened 2017-11-22
Support for sites with content security policy


bovine3dom:
I've started work on this in the `cspbypass` branch, mostly in "requests.ts": https://github.com/cmcaine/tridactyl/blob/3fbef1cba518fdaea71a168ef23d596f2dcbaf24/src/requests.ts

Does anyone understand why line 7 works but line 8 does not? Ideally, we want to overwrite as little of the CSP as possible, and just add our own scripts to the whitelist.

(I am aware that this is a Mozilla bug, but they're moving pretty slowly and I thought this would be an easy fix...)

bovine3dom:
https://github.com/ueokande/vim-vixen/issues/189#issuecomment-346156869 - as this excellent bovine3dom fellow pointed out on the Vixen repository, this is partially fixed already in FF58, so we should make sure that we're testing our fixes to this in FF57.

bovine3dom:
Fixed if you do `set csp clobber` in e26f4a5 onwards.

PR #245, opened 2017-12-21
Cspbypass

cmcaine:


bovine3dom:
Related issue is #109.

bovine3dom:
I'm slightly confused as to what this branch is supposed to be doing.

On https://bitpoll.mafiasi.de/, nothing seems to change from the default Firefox behaviour (i.e, scrolling on the page etc works but showing the command line does not).

If you revert back to one of my first commits in this branch, the command line does work.

I thought the plan was to have a config setting that completely clobbered CSP so that Tridactyl would work?

bovine3dom:
I merged this because I noticed that it fixed `raw.github*` pages. It's now in `master`. Thanks, @antonva.

PR #248, opened 2017-12-22
Fix issue 247 - readme links

IsaacKhor:
Fixes issue 247; turns out that GFM issue links are not processed on README files. [SO Link](https://stackoverflow.com/questions/16539687/github-readme-reference-issue).

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -74,7 +74,7 @@ Mozilla doesn't give us access to them. See [issue #73](https://github.com/cmcai
 
 - Why doesn't Tridactyl work on websites with frames?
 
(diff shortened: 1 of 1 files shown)


IsaacKhor:
I'm an idiot; I think a search/replace went wrong. Anyways, its fixed now.

bovine3dom:
122 and 158 are now the wrong way around; 122 is the iframe one :(

IsaacKhor:
You must be kidding me. I don't know what happened, but its not good. I double checked this time -- its finally fixed corrently.

bovine3dom:
Thanks : )

Issue #230, opened 2017-12-12
Dark theme for commandline


KrazyCavin:
Current theme `/src/static/commandline.css` fits well FF's default light theme.
However, I prefer the dark side, the dark theme of FF. It would be nice if Tridactyl has a dark theme for commandline.

bovine3dom:
As with #126, feel free to make some CSS and we'll add `set darkmode true` or something.

johnbeard:
A very common way to do this for sites that have "light's out" modes is to add a class to top element and then have CSS that uses that class to trigger, overriding whatever the default CSS is. I tried a lights out button before (before we had config) and this worked on the new tab page (dark BG, greenish links to go with the logo):
```
/* Dark theme */
body.darkmode {
    background: #222;
    color: #CCC;
}

.darkmode a {
    color: #6ec071
}

.darkmode a:visited {
    color: #3d8b40;
}
```

I think a better command would be to allow any named colour scheme, rather than "dark/light" boolean. Would be more Vimmy (`colorscheme desert` etc), and also allow multiple themes, which might be user-settable in future? E.g. `set colorscheme darkmode`, where `darkmode` is the theme name and is used to construct the CSS class name. Maybe add a prefix to avoid stomping on internal names?

This is one of those things where we could do with a "change hook" for `set` to trigger and update the CSS. Perhaps some kind of event and observer?

KrazyCavin:
@bovine3dom here is the css file for dark theme https://gist.github.com/KrazyCavin/cebb807e961a4a32bb04aa0c1757bab8
It looks like:
![tridactyl-dark-commandline](https://user-images.githubusercontent.com/14293537/33907618-066f52b6-df86-11e7-9a31-6e384c92036f.jpg)

cmcaine:
How do nested hints look? Try `;p`

KrazyCavin:
@cmcaine Nested hints look ok-ish
![tridactyl-dark-hint-nested](https://user-images.githubusercontent.com/14293537/33934805-3fe4f072-dffa-11e7-95f2-ffab10149d56.jpg)

cmcaine:
Cool. Let's have this work via inherited attributes and variables from a top level class that we can apply to body/the hint host.

Then we just put a config line switch in the approriate code that adds the dark or light mode class.

cmcaine:
@KrazyCavin are you happy to do the CSS side of that?

PR #252, opened 2017-12-24
WIP: Attempt to apply theme to command line

furgerf:
Extract coloring of command line into default style and use dark style
from https://gist.github.com/KrazyCavin/cebb807e961a4a32bb04aa0c1757bab8
Relates to #230.
Reading current theme with `config.get("theme")` doesn't seem to work atm
(always returns default instead).


src/excmds.ts
```diff
@@ -922,7 +922,7 @@ export function composite(...cmds: string[]) {
 
 /** Please use fillcmdline instead */
 //#background
-export function showcmdline() {
```
furgerf: This function appears to be invoked locally only (which seems to be intended, according to the comment).

cmcaine:
You may already know, but FYI you can just force push to your branch to update the PR without closing and re-opening it.

Issue #253, opened 2017-12-25
Support %s magic string for search providers


bovine3dom:
If %s exists in a search provider URL, put the search string there, otherwise just add the search to the end of the URL as is current behaviour.

(Also discussed in #73).

furgerf:
Since https://github.com/cmcaine/tridactyl/pull/254 is merged, can this issue be closed?

PR #254, opened 2017-12-26
Support %s magic string for search providers

furgerf:
Also add reddit as search provider which uses this feature. #253

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -125,11 +125,6 @@ function searchURL(provider: string, query: string) {
         new URL(searchurlprovider.replace("%s", encodeURIComponent(query))) :
         new URL(searchurlprovider + encodeURIComponent(query))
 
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -112,7 +112,7 @@ const DEFAULTS = o({
         "mdn":"https://developer.mozilla.org/en-US/search?q=",
         "gentoo_wiki":"https://wiki.gentoo.org/index.php?title=Special%3ASearch&profile=default&fulltext=Search&search=",
         "qwant":"https://www.qwant.com/?q=",
-
+        "reddit":"https://reddit.com/r/%s",
```
cmcaine: I don't think most users typing `open reddit blah` expect to go to the blah subreddit. We could have a `subreddit` search provider, perhaps.

And, of course, you can add your own providers locally with `setsearchkeyword reddit whatever`

PR #255, opened 2017-12-26
WIP: Attempt to apply theme to command line

furgerf:
Extract coloring of command line into default style and use dark style
from https://gist.github.com/KrazyCavin/cebb807e961a4a32bb04aa0c1757bab8
Relates to #230.


cmcaine:
@furgerf Do you want more assistance?

furgerf:
@cmcaine I didn't have time these past few days but plan to have another go after New Year. My problem was the config, always reporting the default value for the theme like @bovine3dom commented. If you have ideas as to why that might be happening, that would be helpful :)

src/static/commandline.css
```diff
@@ -9,7 +9,6 @@ input {
   font-family: monospace;
```
cmcaine: You don't need to modify this file: default colors and so on can live in here and be overridden by a dark theme.

src/static/commandline.html
```diff
@@ -2,10 +2,12 @@
...
+        <link rel="stylesheet" href="commandline-default.css">
+        <link rel="stylesheet" href="commandline-dark.css">
     </head>
     <body>
         <div id="completions"></div>
-        <input id=tridactyl-input></input>
+        <input id="tridactyl-input"></input>
         <script src="../commandline_frame.js"></script>
     </body>
 </html>
```
cmcaine: Why have you changed the mode of userChrome-minimal.css?

src/commandline_frame.ts
```diff
@@ -51,10 +52,22 @@ function enableCompletions() {
...
 
 export function focus() {
     enableCompletions()
     document.body.classList.remove('hidden')
+
+    // update theme of command line
+    let theme = Config.get("theme")
+    if (theme !== lastTheme) {
+        if (lastTheme) {
+            document.body.querySelectorAll("*").forEach(child => child.classList.remove(lastTheme))
```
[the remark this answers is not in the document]
cmcaine: Applying the CSS in this way is the root of your troubles: not all the elements you want to style will exist to receive the dark or default class.

src/commandline_frame.ts
```diff
@@ -51,10 +52,22 @@ function enableCompletions() {
 /* document.addEventListener("DOMContentLoaded", enableCompletions) */
 
 let noblur = e =>  setTimeout(() => clInput.focus(), 0)
+let lastTheme = undefined;
```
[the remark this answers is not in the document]
furgerf: Yes - I like being explicit, should I change it?

src/commandline_frame.ts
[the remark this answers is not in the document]
furgerf: Ah yes, thanks! :)

src/static/commandline.html
furgerf: It was the only css that had 755, all others were 644, that looked like an accident to me. Is there a reason for the differing permissions?

furgerf:
Thanks for the lesson @cmcaine! Another thing, we don't want to actually add/remove the `default` class, should I handle this special case or leave it as-is in favor of readability?
I've also had a go at the hinting. 3 things:
- I couldn't use `body.dark` like for the command line so I've chained the classes, is that ok?
- The red box with the letter (`span.TridactylHint`) doesn't apply the dark theme. I've tried a few different combinations but I didn't get it to work.
- As I understood it, all hints are always re-created whenever they're shown, so there shouldn't be any need to keep track of (and remove) the previous theme. However, when going from dark to default theme on a page, the hints continue to show up dark.

src/hinting.ts
```diff
@@ -131,6 +131,7 @@ class Hint {
         `
         modeState.hintHost.appendChild(this.flag)
         target.classList.add('TridactylHintElem')
+        target.classList.add(config.get("theme"))
```
cmcaine: You're adding the theme class to the target element, which is the web content element that is getting hinted, not the hint label.

Check out the HintState destructor to see how hints are removed (Hint.hidden = true). You need to hook into that logic, maybe add a destructor to the Hint class, maybe just remove the theme when hidden = true.

Also, the theme shouldn't be a name likely to clash with web content CSS. Try "TridactylHintElemDark" or something.

src/static/commandline.html
cmcaine: There's no reason for the permissions to continue to differ.

src/static/hint-dark.css
```diff
@@ -0,0 +1,18 @@
+span.dark.TridactylHint {
```
cmcaine: Here you should consider applying the dark theme to div.TridactylHintHost and then inheriting from that.

src/commandline_frame.ts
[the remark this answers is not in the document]
cmcaine: Yes please. It should be obvious to JS/TS devs what value an uninitialised var has.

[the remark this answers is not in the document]
cmcaine: You could type it as a string, though.

cmcaine:
Design suggestions:

Apply a class to each of

 - The commandline root
 - The web content root (this one needs a name that is unlikely to clash with other people's styles)

N.b. You can get the root element with `document.querySelector(':root')`

Repeated colours should be set as variables in the CSS for the commandline. Maybe also in the webcontent, but need to make them non-clashing names there, so might be more effort than it's worth.

```css
#commandline-dark.css

:root.dark {
  --foreground: #ccc;
  --background: black;
}

:root.dark td {
  color: var(--foreground);
  background-color: var(--background);
}

#content-dark.css
:root.TridactylThemeDark {
  --tridactyl-variable-foreground: #ccc;
  /* ... */
}

:root.TridactylThemeDark div.TridactylHintHost {
}
```

furgerf:
I've adapted the changes to the command line per your comments, thanks! The hints are definitely not finished yet, though...
- About the name of the theme, you suggested `TridactylThemeDark` for the hints to avoid clashes, am I right to assume that you'd also want to use the same name for the command line? If so, where would be a good place to put such a `getThemeName` function? The functions in config.ts are all generic so I hesitated putting it there. Obviously we can't keep `if (theme === "dark") then add class "TridactylThemeDark"` in hinting.ts.
- I've put the theming of the hints into the `hidden` property and set that property to false in the `Hint` constructor since it was already doing that, implicitly. That looks best to me.
- It's strange, when I add the `content-dark.css` to the manifest, the theming of the command line no longer works (no matter what I put into the file), any ideas?

furgerf:
Hi @cmcaine I was busy with exams lately and just noticed this is still open. Do you maybe have some time to have a look and give feedback? Thanks :)

cmcaine:
I commit to reviewing this in the next 48h :)

IsaacKhor:
Any progress on merging this?

furgerf:
Still waiting for feedback :)

cmcaine:
Sorry it took me so long to get around to this. Fixed and merged, thank you :)

Issue #183, opened 2017-12-01
Vimperator Like Hint Mode


clov0:
Is there a way for a Vimperator-like hint mode to be available as an alternative to the Vimium approach?
I mean for the hint mode labeling to be done using some -maybe prefixed- numbers, rather than text labels, and instead using text content of the element as a hidden tag for filtering and narrowing down while typing.

bovine3dom:
I'd be overjoyed if someone would make a PR for this. We can have a `set hintmode vimperator | vimium`. Otherwise, it's reasonably far down my todo list.

johnbeard:
You could also keep the current alphabetic hints, and have some sort of "escape" to enter the hinted element text, which will narrow down from whatever hints are left. Perhaps `:`? Would need to be able to get back to hint-label mode in case there are multiple elements with identical text.

cmcaine:
Someone else suggested capital letters for the hints, lower case for searching text. I think vimperator just uses any key that doesn't match a hint label to search the hinted element text and link. That seems fine to me. There's some potential for confusion, but maybe we just show the current value of the search and hint select strings?

A bind to switch between searching hinted element text and hint label seems like a pretty good option for people who don't like to search often.

jstetten:
>You could also keep the current alphabetic hints, and have some sort of "escape" to enter the hinted element text, which will narrow down from whatever hints are left. Perhaps :? Would need to be able to get back to hint-label mode in case there are multiple elements with identical text.

Vimperator used `mapleader` to toggle from hint-matching to text-matching and back again (if needed). `;` is handy and rarely appears in links, so worked well for me. For those using alphabetic hints, it was a simple solution.

cmcaine:
@saulrh said on Riot that they'd have a go at this.

Circa here: https://matrix.to/#/!xmubtggllUtLrQiGyn:matrix.org/$15129502964031442GlZXO:matrix.org

Issue #234, opened 2017-12-13
Escape is passed through to website when Tridactyl interface elements are open


lehasb:
I expected to find this issue already listed, but I don't see it anywhere in the list, so:

When a piece of the tridactyl interface is open (hints or command bar), hitting escape not only exits the tridactyl interface, but also passes through to the main webpage. It's an issue on pages like http://ccca.biola.edu/advent/ where there's a separate popup that is closed when you hit escape. Expected behavior is that tridactyl will catch the escape when an interface element is open and not pass it on to the webpage.

chrlsnchrg:
I'd like to add that this bug seems to extend to _all_ keys in hint mode, not just ESC. When keys used for hinting in "hint-mode" are pressed, both tridactyl and the website receive the key press. 
Try for example pressing "s" in hint mode on https://4chan.org/ck/catalog. The catalog page's search option will be activated _and_ the hint will be activated.

Koushien:
Just naively, I noticed that `Escape` and `s` only activated the noted events in your respectively linked pages on keyup (the Biola U page seems to only listen for keyup), which Tridactyl never tampers with.

I've implemented a stupidly basic handler that suppresses keyup events outside of Ignore Mode in this branch: https://github.com/Koushien/tridactyl/commit/d29d2e19bc2ce9552b84c749e76cd84863c76707. I neither know if this is reasonably minimal or if the general method is better than the alternatives, but it's something you can use for now.

lehasb:
Is there a recommended way to check how a website is using key events? I've noticed some other oddities in keypress behavior around dropdown menus, but I'm not sure if it's related to this or a separate issue.

cmcaine:
You can try to reason it out from looking at the exclusion rules in src/keydown_content.ts and comparing that to what happens when you press keys or you can grep the source of the website for `addEventListener`

bovine3dom:
Also happens on https://www.adamsmith.org/ - pressing `<Esc>` sends you to a squarespace login page (let's just set aside how strange a bind that is anyway).

bovine3dom:
This has started annoying me more so I've made switched the labels to make myself feel better.

PR #258, opened 2017-12-29
Implement vimperator-style hint filtering

saulrh:
This is a first pass at addressing #183. It has the necessary flows for control and data to implement that functionality, but there are some rough edges to be smoothed out before the feature is actually ready.

Changes:

- Add a `hintfiltermode` config option that selects hint filter modes. It defaults to `simple`.
- Augment the `Hint` class with a `filterdata` field of type `any` that can be used to store information for filtering.
- Add a layer of indirection in `hinting.ts` to permit selection of hint building (in `hintPage`) and filtering (in `filter`) functions.
- Move current hint selection functionality into the `simple` setting for `hintfiltermode`, backed by a builder that doesn't use `filterdata` at all and a filter that's a straight copy-paste of current hint filtering code. This mode should be identical to current behavior and it is set as the default in `config.ts`.
- Add a `vimperator` setting for `hintfiltermode`. The builder for this mode populates the `name` of the hint as in `simple` and the `filterData` field of hints with a DWIM string extract from the text, content, title, or inner html of the hint's target. The filter function consumes the filter string one character at a time. Characters in `hintchars` are matched against the names of hints, and other characters consume from the head of filterdata. This is equivalent to partitioning the filterstring into two (hintchar and non-hintchar) subsequences and testing for name starts with hintchar portion and filterdata string contains non-hintchar portion as a subsequence.
- Add a `vimperator-reflow` setting for `hintfiltermode`. The builder is the same as for `vimperator`. The filter is almost the same as for `vimperator`, but it chooses new names for each hint after every non-hintchar character in the filterstring. This essentially "reflows" hints as the user filters them down.

Thoughts:

- I feel like I could still improve the pattern of data mutation in the vimperator-style hint filter function. The following thoughts went into the structure here:
  - Hints should have some kind of immutable unique ID so that they can be more easily debugged.
  - The filterable text needs to be a snapshot of the element from the moment the page was hinted, otherwise elements that change while the user is selecting a hint would be impossible for users to handle.
  - When consuming a hintchar from the filterstring, hints are matched using their names, which are a function of any previous non-hintchars in the filterstring, so the outermost loop must be over the filterstring, unlike the simple-style hint filter.
  - The code should reduce to simple-mode filtering if the filter string is entirely composed of hintchars.
  - The vimperator-style hint filter currently operates by storing temporary versions of the names and filterdatas for hints and consuming from them as it processes characters from the filterstring. The alternative was to keep pointers into the immutable names and filterdatas. I felt that this code was clearer and easier to work with, and since its only operations are to slice strings I *think* - but I'm not certain that - the browser should be able to entirely avoid string copies and performance should be the same as maintaining indexes. I haven't tested this.
- Switching the filtering mode halfway through a hinting operation could select almost any hint on the page essentially at random, so I determine the filter function when hints are built and store it in the mode state.
- Should I try to break the vimperator filter function into smaller pieces? It feels a bit big right now.

Things to do in the future:
- It looks like this may be the first configuration option that's an enum. I did not address that problem in this PR: `hintfiltermode` is an arbitrary string and unrecognized strings have undefined behavior (but really just blow up).
- In the vimperator-style filter function, there's a relatively easy spot to add scoring and sorting of hints based on the quality of the match between the hint and the filterstring. I decided that that functionality was out of scope for this PR, as it's more fine-tuning and shouldn't require any changes to control flow or data structures.
- `' '` is captured by the browser and navigates to the currently focused hint, which is *very* annoying when the instinctive approach for narrowing to a single hint is to type a couple words selected from its target element. Setting `' '`as a hintchar in the mode state didn't seem to do anything. This will certainly be something to investigate when tuning the filtering and adding scoring+sorting; "all `/[^[:space:]]+/` must be filterable text" turns out to be *far* nicer than "filterstr is subsequence of filterable text".
- Go read the code for every vimperator-descendant out there that has a feature like this and see what their DWIM is for an element's filterable text.
- If the user is selecting a hint by narrowing to it with non-hintchar inputs, they's probably be typing a short selection of actual text from the hint's target element. If they're typing full words and going quickly, they almost certainly won't react to the hint being selected fast enough to stop typing before tridactyl starts taking non-filter-string input, resulting in spurious input. The hard solution to this is to ignore keystrokes for a bit after a successful non-hintchar selection. The easy solution is to add a setting that suppresses the automatic selection when only a single hint remains active.

### DIFF
--- a/src/hinting.ts
+++ b/src/hinting.ts
@@ -243,6 +243,10 @@ function filterHintsSimple(fstr) {
 }
 
 function filterHintsVimperatorReflow(fstr) {
+    filterHintsVimperator(fstr, true)
+}
+
+function filterHintsVimperator(fstr, reflow=false) {
     let active = modeState.hints.map(h => {
         return {
             'hint': h,
@@ -274,11 +278,13 @@ function filterHintsVimperatorReflow(fstr) {
                     act.hint.hidden = true
                 }
             }
-            // ...and then reflow the remaining active hints
-            let names = hintnames(nextActive.length)
-            for (let [act, name] of izip(nextActive, names)) {
-                act.name = name
-                act.hint.flag.textContent = name
+            if (reflow) {
+                // ...and then reflow the remaining active hints
+                let names = hintnames(nextActive.length)
+                for (let [act, name] of izip(nextActive, names)) {
+                    act.name = name
+                    act.hint.flag.textContent = name
+                }
             }
         }
         active = nextActive
@@ -300,54 +306,6 @@ function filterHintsVimperatorReflow(fstr) {
     if (active.length == 1) {
         selectFocusedHint()
     }
-}
-
-/** Show only hints:
-    - prefixed by the subset of fstr in the hintchars config.
-    - containing the rest of fstr as a subsequence in a dwim-type
-      chunk of their html.
-    Focus the first match.
-**/
-function filterHintsVimperator(fstr) {
-    const active: Hint[] = []
-    let foundMatch
-    for (let h of modeState.hints) {
-        if (!filterHintsVimperatorPredicate(fstr, h)) {
-            h.hidden = true
-        } else {
-            if (! foundMatch) {
-                h.focused = true
-                modeState.focusedHint = h
-                foundMatch = true
-            }
-            h.hidden = false
-            active.push(h)
-        }
-
-    }
-    if (active.length == 1) {
-        selectFocusedHint()
-    }
-}
-
-function filterHintsVimperatorPredicate(fstr, h) {
-    let fstrName = ''
-
-    let curIdx = 0
-    for (let c of fstr) {
-        if (config.get("hintchars").includes(c)) {
-            fstrName = fstrName + c
-            if (!h.name.startsWith(fstrName)) {
-                return false
-            }
-        } else {
-            curIdx = h.filterData.indexOf(c.toLowerCase(), curIdx)
-            if (-1 == curIdx) {
-                return false
-            }
-        }
-    }
-    return true
 }
 
 /** Remove all hints, reset STATE. */
--- a/src/hinting.ts
+++ b/src/hinting.ts
@@ -144,7 +144,7 @@ type HintSelectedCallback = (Hint) => any
 
 /** Place a flag by each hintworthy element */
 class Hint {
-    private readonly flag = document.createElement('span')
+    public readonly flag = document.createElement('span')
 
     constructor(
         private readonly target: Element,
@@ -212,7 +212,7 @@ function buildHintsVimperator(els: Element[], onSelect: HintSelectedCallback) {
     const filterableTextFilter = new RegExp('[' + escapedHintChars + ']', 'gi')
     for (let [el, name] of izip(els, names)) {
         let ft = elementFilterableText(el)
-        // strip out non-alphanumeric characters and hintchars.
+        // strip out hintchars
         ft = ft.replace(filterableTextFilter, '')
         logger.debug({el, name, ft})
         modeState.hintchars += name + ft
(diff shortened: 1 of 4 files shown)


bovine3dom:
This is fantastic, thanks so much.

I have a couple of minor comments:

> ' ' is captured by the browser and navigates to the currently focused hint, which is very annoying when the instinctive approach for narrowing to a single hint is to type a couple words selected from its target element. Setting ' 'as a hintchar in the mode state didn't seem to do anything. This will certainly be something to investigate when tuning the filtering and adding scoring+sorting; "all /[^[:space:]]+/ must be filterable text" turns out to be far nicer than "filterstr is subsequence of filterable text".

This is probably just #234.

---

I think most people would appreciate a `:hintmode` (or better name) excmd that changes the hintfiltermode and sets up hintchars to something sensible: e.g, `hintmode vimperator` would change the hintchars to be numerical and change the hinting filter.

saulrh:
> This is probably just #234.

Ah, that would do it. I think you're right. It's hard to see it because the page navigates away and clears the console, but I see flashes of hint-filtering output before the navigation happens.

> I think most people would appreciate a :hintmode (or better name) excmd that changes the hintfiltermode and sets up hintchars to something sensible: e.g, hintmode vimperator would change the hintchars to be numerical and change the hinting filter.

Ah, that's an excellent idea. 

I should probably make it so that it swaps the user's hintchars in and out, rather than clobbering them completely. Which looks like it wouldn't even be difficult, since config settings can have type object and they behave sanely. That's perfect - I expect approximately nobody to want to have the same hintchars for both simple and vimperator hintfiltermodes, which means that the hintchars setting probably wants to be an object mapping hintfiltermodes to character-sets. On the other hand, backwards-compatibility - that seems like the kind of thing that would break people's config-files. Not optimal. It doesn't look like we have a system for migrating a user's config-file on tridactyl upgrade?

bovine3dom:
No, we don't currently have any migration procedure. config.ts:173 is how I handled moving the bind system - we could potentially do something similar.

However, I would rather we didn't have hint characters as an object because I still haven't thought of a nice way to allow users to set objects - everything that uses an object so far has an excmd helper (e.g. bind) - although really `set object key1 key2 value` -> object[key1][key2] = value would probably be fine, now that I've actually thought about it for a little bit...

So let's clobber them for now and we can think up a nicer way of doing it later once config is improved : ).

---

I've tested the code and so far it seems to work without any show-stopping bugs. I'll actually read it now, and, if all goes well, we can merge it.

saulrh:
> So let's clobber them for now and we can think up a nicer way of doing it later once config is improved : ).

Yup. :D

Is there a ticket for discussing config improvements? Seems like my brain is insisting on thinking about it and it'd be good to have a place to dump idle thoughts.

src/hinting.ts
```diff
@@ -157,8 +178,56 @@ class Hint {
...
+    for (let [el, name] of izip(els, names)) {
+        console.log({el, name})
+        modeState.hintchars += name
+        modeState.hints.push(new Hint(el, name, null, onSelect))
+    }
+}
+
+function buildHintsVimperator(els: Element[], onSelect: HintSelectedCallback) {
+    let names = hintnames(els.length)
+    const filterableTextFilter = new RegExp('[' + config.get('hintchars') + ']|[^[:alnum:]]', 'gi')
```
saulrh: Ah, I forgot about this. If someone tries to use `]` as a hintchar this will do strange things. That suggests a feature ("filter this text out the filterable text when in vimperator-style hintfiltermodes"), but right here I should probably do something about it.

...Is there a common wisdom for escaping strings for interpolation into regex character classes? o_o

bovine3dom: Not really: https://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex#answer-6969486

The general approach is not to use user-provided strings in a RegEx, but to use `find` and its friends instead.

bovine3dom:
> Is there a ticket for discussing config improvements? Seems like my brain is insisting on thinking about it and it'd be good to have a place to dump idle thoughts.

I've just made #259 and #260 - there's not a meta-bug, feel free to make one if you want.

bovine3dom:
I've read through the code and it has my seal of approval, almost:

@saulrh, would you mind escaping that RegEx you pointed out? Also Colin does have a little point; those if statements could be neater as switches.

saulrh:
> @saulrh, would you mind escaping that RegEx you pointed out? Also Colin does have a little point; those if statements could be neater as switches.

Will do both of those. May take me until monday-ish thanks to holiday stuff.

Looks like I also have some merge conflicts to deal with. How do you want me to deal with that? Rebase, merge master into feature branch before merging back to master?

cmcaine:
I just do `git fetch origin && git rebase origin/master` from the feature
branch.

bovine3dom:
Feature request - if multiple hints have the same href, and are the only hints remaining, follow that link.

Totally happy for it to go into a separate issue once merged. I just wanted to record it while I was thinking about it.

src/hinting.ts
```diff
@@ -195,7 +195,10 @@ function buildHintsSimple(els: Element[], onSelect: HintSelectedCallback) {
...
 function buildHintsVimperator(els: Element[], onSelect: HintSelectedCallback) {
     let names = hintnames(els.length)
-    const filterableTextFilter = new RegExp('[' + config.get('hintchars') + ']|[^[:alnum:]]', 'gi')
+    // escape the hintchars string so that strange things don't happen
+    // when special characters are used as hintchars (for example, ']')
+    const escapedHintChars = config.get('hintchars').replace(/^\^|[-\\\]]/g, "\\$&")
+    const filterableTextFilter = new RegExp('[' + escapedHintChars + ']', 'gi')
     for (let [el, name] of izip(els, names)) {
         let ft = elementFilterableText(el)
         // strip out non-alphanumeric characters and hintchars.
```
cmcaine: Why not omit all this regex stuff?

`Array.from(ft).filter(char => hintchars.indexOf(char) == -1).join('')`

cmcaine: My snippet returns `ft` with every character in hintchars removed, which I think is what the regex does, but I haven't read the regex closely to check.

saulrh: Yes, the intent is to remove from the filterable text all characters that are in the hintchars set.

I didn't look too hard at filter-join pipelines for this because I was worried about performance. I haven't done a real survey, but I know that I've run into a couple sites with elements with tens of k of inner html - it can happen when someone puts a target on an element that has a ton of stuff inside it. I suppose that we could just wait for profiling (or reports of slowness) instead of optimizing prematurely, or just truncate, or something else, though?

saulrh: I agree that I'm not happy with the regex silliness. :/

cmcaine: > Programmers waste enormous amounts of time thinking about, or worrying about, the speed of noncritical parts of their programs, and these attempts at efficiency actually have a strong negative impact when debugging and maintenance are considered. We should forget about small efficiencies, say about 97% of the time: **premature optimization is the root of all evil**. Yet we should not pass up our opportunities in that critical 3%.

\-- *Donald Knuth*

We should go with the simpler option and consider higher level optimisations such as truncation or avoiding precompution entirely, if optimisation is required.

Out of curiosity, I decided to benchmark and RegExp is a bit faster, but they're both pretty fast: https://jsperf.com/filter-chars-from-string

On my system, filtering a .7MB string takes ~62ms via the array method and ~50ms with RegExp. For reference, human reaction time is about 200ms.

bovine3dom: Human reaction time isn't a particularly good benchmark for UI - I'd aim for screen refresh (~16ms) for guaranteed imperceptibility.

cmcaine: Sure, but I don't care about getting it down that low here. Doubtful we could get the total down that far.

Looking at the code some more, I would have had vimperator mode accept any character and move the filter function to the Hint, avoiding all precomputing. I think what you've done is fitting the new implementation a bit too strongly to the old model.

We could just slap a truncation at 1K characters (in case of big child nodes (data URIs, script tags, etc) and use this if you don't fancy refactoring.

bovine3dom:
@cmcaine this is a really nice feature, are there any outstanding reasons we shouldn't merge this?

jstoone:
@bovine3dom hopefully I can get to test it out this weekend. I'll report back if I stumble upon any hiccups.

On the other hand if you merge it, alpha/beta users can easily give it a proper go, and then we can iterate if needed.

I am longing for this feature. ❤️

bovine3dom:
@jstoone @cmcaine eventually got around to reviewing it and found a bug or two. I think he fixed it but has the code locally.

cmcaine:
I'll push it tomorrow.

jstoone:
@bovine3dom @cmcaine nice work peeps!

cmcaine:
Merged. Sorry for the delay.

I changed the behaviour of the vimperator-mode hint filter when non-hintchars are typed: previously each character would be checked individually to see if it is included, now each run of characters (runs started by a non-hintchar character and ended just before a hintchar) is checked.

So if the hint text is "simplify hintnames", then previously `sphn` and `nasi` would match, but now they don't. `simp`, `ify`, `hint`, etc. all still work.

I don't know if this is a good change. I think the old way matches too much, but this is probably too restrictive? We should probably do some fuzzy matching instead (`spnh` would match, but `hnps` wouldn't).

In code:

```
regex = new RegExp(run.str.split('').map(RegExp.escape).join('.*'))
action = action.filter(hint => regex.test(hint.filterData)
```

Where RegExp.escape is bobince's version from here: https://stackoverflow.com/questions/3561493/is-there-a-regexp-escape-function-in-javascript/

PR #267, opened 2018-01-03
WIP: Native messenger

johnbeard:
This is a very basic sketch from a while back about how to do a native messenger, based on the FF documentation for the ping pong messenger.

You need to modify `native/tridactyl.json` to point to the Python file and copy/symlink it to one of the locations where Firefox can find it (look near the end of [this page](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Native_manifests)). I used `~/.mozilla/native-messaging-hosts`.

There are two ex-commands:
* `ping`, which is a command that pings the native messenger and gets its version
* `loadconfig`, which reads from the file `~/.config/tridactyl/tridactylrc` or `XDG_CONFIG_HOME/tridactyl/tridactyrc`

You should see output from both in the Browser Console (not the Web Console!).

I haven't had a lot of time to deal with this recently and I won't have for a little while - feel free to repurpose anything in any way!

### DIFF
--- a/native/native_main.py
+++ b/native/native_main.py
@@ -1,41 +1,71 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
-import json
 import sys
-import struct
+
+import nm_comms
+import user_config
+import nm_utils
 
 
-# Read a message from stdin and decode it.
-def get_message():
-    raw_length = sys.stdin.buffer.read(4)
-    if not raw_length:
-        sys.exit(0)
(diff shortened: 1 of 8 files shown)


IsaacKhor:
I'm merging this into my config overhaul PR

bovine3dom:
@johnbeard thanks for this. Some version of it lives on in Tridactyl, so I'm closing this.

I've added the `nativemessenger` label to make it a bit easier to find again if we decide we need the original code.

PR #286, opened 2018-01-21
Add urlmodify command

johnbeard:
Has a few modes:

    -t: Straight text replacement
    -r: Regexp replacement
    -q: Change a query's value to a new one
    -Q: Delete a given query (and value if present)
    -g: Graft a path onto the current URL (or a parent path of it)

These can be used direct on the command line, or bound to keybindings.
The idea is to allow commands to easily navigate around a website. For
example, navigating to a project's issues page on GitHub can be done
with a graft command.

As a URL modification is generally site-specific, binding will be much
more useful with aucmds.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -87,7 +87,7 @@ import * as keydown from "./keydown_background"
 //#background_helper
 import {activeTab, activeTabId, firefoxVersionAtLeast} from './lib/webext'
 //#content_helper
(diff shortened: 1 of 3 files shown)


bovine3dom:
This is great, thanks!

A couple of minor points - I think the help could probably do with a few examples to make what the command can be used for clear, and I think it should be mentioned that the text replacement only replaces the first occurence.

PR #298, opened 2018-01-28
Urlmodify documentation

johnbeard:
As requested in PR #286 

Also fixed a silly bug in the regex replace mode.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -463,9 +463,10 @@ export function urlmodify(mode: "-t" | "-r" | "-q" | "-Q" | "-g", ...args: strin
                     + "<regexp> <new> [flags]")
             }
 
-            if (args[2] && args[2].search('/^[gi]+$/') === -1)
+            if (args[2] && args[2].search(/^[gi]+$/) === -1)
             {
-                throw new Error("RegExp replacement flags can only include 'g', 'i'")
+                throw new Error("RegExp replacement flags can only include 'g', 'i'"
+                    + ", Got '" + args[2] + "'")
             }
 
             let regexp = new RegExp(args[0], args[2])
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -426,19 +426,52 @@ export function urlparent (count = 1){
...
+ *   Deletes the given query (and the value if any):
+ *      * `http://e.com?id=foo&page=1` -> (`-Q id`) -> `http://e.com?page=1`
+ *
+ * * Graft mode: `urlmodify -g <graft_point> <new_path_tail>`
+ *
+ *   "Grafts" a new tail on the URL path, possibly removing some of the old
+ *   tail. `graft_point` >= 0 counts path levels from the left, < 0 counts
+ *   from the right
+ *      * `http://e.com/issues/42` -> (`-g 0 foo`) -> `http://e.com/foo`
+ *      * `http://e.com/issues/42` -> (`-g 1 foo`) -> `http://e.com/issuesfoo`
```
bovine3dom: Another little typo, should be issues/foo

src/excmds.ts
```diff
@@ -426,19 +426,52 @@ export function urlparent (count = 1){
...
+ *
+ * * Query delete mode: `urlmodify -Q <query>`
+ *
+ *   Deletes the given query (and the value if any):
+ *      * `http://e.com?id=foo&page=1` -> (`-Q id`) -> `http://e.com?page=1`
+ *
+ * * Graft mode: `urlmodify -g <graft_point> <new_path_tail>`
+ *
+ *   "Grafts" a new tail on the URL path, possibly removing some of the old
+ *   tail. `graft_point` >= 0 counts path levels from the left, < 0 counts
```
bovine3dom: As I mentioned on Matrix, I think this needs a little more explanation. You could say that you are counting from starting from the left of the path, and perhaps have website.com/this/is/a/path?

Additionally, it might be worth mentioning that "-1" is a bit magic in that you append an extra level to the path before you start counting.

Issue #300, opened 2018-01-30
Make it possible to not URI-encode search query string?


zcesur:
I would love to be able to use the Wayback Machine as one of my search keywords, but my URIs are being percent-encoded (#131) which prevents their search engine from working properly. Perhaps we could add the option to disable this?

> https://web.archive.org/web/*/https://github.com/cmcaine/tridactyl
> https://web.archive.org/web/*/https%3A%2F%2Fgithub.com%2Fcmcaine%2Ftridactyl

bovine3dom:
This is quite a reasonable request.

I'll have to do some thinking about how to best expose it. Perhaps a flag on `searchsetkeyword`?

cmcaine:
We might be able to work it out automatically from the location that the insertion/replacement happens.

For all of our existing defaults it's safe not percent encode if the query string isn't actually in the query part of the URL. (cnrtl and wikipedia search are the only ones that aren't queries and they both accept non-encoded strings)

cnrtl doesn't like slashes, but it doesn't like them if you encode them either.

Anyway, this is probably a good enough heuristic and it means the users don't need to think about it.

We can test with

```
if (searchURL.contains("%s") && (new URL(searchURL)).search.contains("%s") || (new URL(searchURL)).search != "") {
    query = encodeURIComponent(query)
}
// ...
```

PR #301, opened 2018-02-01
Automatically percent-encode seach query only if in query string (and add tests)

johnbeard:
For issue #300 

Move search term encode/interpolate/append logic to UrlUtil and put some tests on it. This allows an easier interface to add explicit encoding later if guessing based on location in the URL doesn't cut it for some searches.

Use in `excmds.ts `is a follow-up commit as I a little unsure about the import method for a "hidden" helper.


cmcaine:
You don't need to import urlutil twice, we just need to make it available to both content and background, so I've adjusted the last commit slightly.

johnbeard:
Aha, I see how it's done, thanks for the fixup!

cmcaine:
Thanks very much! :)

PR #314, opened 2018-02-20
[RFC] Config overhaul

### LINKED ISSUE #79: configuration file with custom key bindings
First of all, thanks for writing this extension. I'm a long time vimperator user, but clearly that chapter is over with the latest firefox. 

Is there a way to have tridactyl read a config file with custom key-bindings, and options that one could specify? If so, what should it be called and where should it live? 

I typically have a zoo of *rc files that i share between machines and would ideally like to have one for tridactyl as well. 

i'm on linux. 
thanks!

### LINKED ISSUE #98: viewconfig --but-only-things-i've-changed
Hello!
* I think it would be very useful to have a way to list the active bindings.
This list should have a flag telling if the binding is custom or factory, and ideally if it is a custom rebind of a factory one.
The function should accept a parameter to filter on one key, so that users can easily know if the binding they are about to create will overwrite another one… It should prevent some confusions…
* Some sort of “reset ALL bindings to factory defaults” panic function should probably be a good idea, too…

### LINKED ISSUE #260: Migration between Tridactyl versions
Particularly, documented ways to deprecate commands / configuration options (blocked by user feedback probably).

Also, we want a way to move a user from some old configuration scheme to another - for example, if a string configuration option becomes an object.

### LINKED ISSUE #280: Config page
I think it's finally time that the much vaunted config page actually gets started. There are two-and-a-half approaches that I can think of to implement a config page:

1. Load .tridactylrc, or at least some in-browser equivalent until we figure out the native messenger
2. Direct JSON editing of the the userconfig (similar to how VSCode handles configuration) in a `config.html` page
3. Some sort of hybrid where we either load from option 1 (from either the fs or web-ext storage) and display the resulting user config in something similar to option 2

I'm inclined to go with the second option, despite the 1st being more 'vim-like', given typical user expectations (raise your hand if your extension wipes your config on startup), and just the convenience factor. Having a central location displaying all active config options is going to useful anyways. Option 3 is just more code for no good reason (and I don't feel like writing it). 

Any suggestions before I start working on an implementation of option 2? Do you mind if I write it in InfernoJS? Pulling in both CodeMirror and Inferno as dependencies seems like a bit much, but if we're going to be migrating the command line over to Inferno anyways, might as well start using it now.

### LINKED ISSUE #346: A way to use bépo layout ?
Hi,

I use bépo keyboard with this Vim "adaptation" --» http://bepo.fr/wiki/Vim
Is there is a way to do the same this with tridactyl.
Thanks for your works !

IsaacKhor:
It's finally here -- the long-awaited configuration page! Resolves #280.

The UI is pretty much completed, now I just have to restructure the config to use the rc file as the ultimate source of truth.

Fixes #79 by converting to rc-based config.
Fixes #260 by adding a migration engine.
Fixes #98 by adding a current settings page.
Fixes #346  by letting users make custom layout files for each keyboard layout.


bovine3dom:
This looks really promising, thanks so much!

A minor point: you should load the content script so that Tridactyl works (hinting etc) on the settings page, similarly to how we do it with the new tab page - https://github.com/cmcaine/tridactyl/blob/e122caa3262597bd563ce2a3ba1d2e21e21b9191/src/static/newtab.template.html#L13

IsaacKhor:
The RC functionality is now working. I just have older config migration and native integration left.

IsaacKhor:
Everything is done; I will try to tack on a error handling mechanism so the extension doesn't just fall over and die if the native messenger encounters some kind of problem, but all of the core functionality is now done.

behrmann:
I do hope this lands sometime soon, an rc-based config, where I finally have a source of truth, where I can look up the config would be awesome.

native/native_main.py
```diff
@@ -0,0 +1,115 @@
...
+    sys.stdout.buffer.flush()
+
+
+def findUserConfigFile():
+    """ Find a user config file, if it exists. Return the file path, or None
+    if not found
+    """
+    config_dir = _getenv("XDG_CONFIG_HOME",
+                         os.path.expandvars('$HOME/.config'))
+    candidate_files = [os.path.join(config_dir, "tridactyl", "tridactylrc")]
```
bovine3dom: I think we should permit ~/.tridactylrc and ~/.config/tridactylrc as well.

native/native_main.py
```diff
@@ -0,0 +1,115 @@
...
+    config_path = None
+
+    # find the first path in the list that exists
+    for path in candidate_files:
+        eprint("Checking file {}".format(path))
+        if os.path.isfile(path):
+            config_path = path
+            break
+
+    return config_path
```
bovine3dom: If no config is found, this gives the very helpful message "An unexpected error occured" in the web console...

native/install.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
+
+# To install, curl -fsSl 'url to this script' | bash
```
bovine3dom: Why not `| sh`?

readme.html
```diff
@@ -0,0 +1,313 @@
+<p><img src="src/static/logo/Tridactyl_200px.png" alt="Tridactyl logo"></p>
```
bovine3dom: As this page is generated it should be added to .gitignore :)

.gitignore
```diff
@@ -1,7 +1,8 @@
-build
-node_modules
+.vscode/
+build/
+generated/
```
bovine3dom: Please add the files you generate with marked to this list.

src/config.ts
```diff
@@ -1,361 +1,153 @@
...
     e.g.
         set("nmaps", "o", "open")
         set("search", "default", "google")
         set("aucmd", "BufRead", "memrise.com", "open memrise.com")
 */
-export function set(...args) {
+export function set(...args: any[]): void {
     if (args.length < 2) {
-        throw "You must provide at least two arguments!"
+        throw `Must provide > 2 arguments!`
```
bovine3dom: Why was this changed? The new error is wrong: you need >= 2 arguments.

src/config.ts
```diff
@@ -1,361 +1,153 @@
...
-//
+import DEFAULTS from './config_defaults'
+import * as Utils from './utils'
 
-const CONFIGNAME = "userconfig"
-const WAITERS = []
-let INITIALISED = false
+export const CONFIG_NAME = 'config'
+const asyncGetters: (() => void)[] = []
+let initialised = false
```
bovine3dom: Are GLOBAL_VARIABLES in all caps frowned on in JS?

src/config.ts
```diff
@@ -1,361 +1,153 @@
...
-    Proposed semantic for config versionning:
+/** Updates the config to the latest version. Proposed semantic for config
+    versionning:
      - x.y -> x+1.0 : major architectural changes
-     - x.y -> x.y+1 : renaming settings/changing their types
-    There's no need for an updater if you're only adding a new setting/changing
-    a default setting
-
-    When adding updaters, don't forget to set("configversion", newversionnumber)!
+     - x.y -> x.y+1 : renaming settings/changing their types There's no need for
```
bovine3dom: This comment has got mangled somehow.

src/config.ts
```diff
@@ -1,361 +1,153 @@
...
+            // Before we had a config system, we had nmaps, and we put them in the
+            // root namespace because we were young and bold.
+            let legacy_nmaps = await browser.storage.sync.get('nmaps')
+            if (legacy_nmaps)
+                cfgObj['nmaps'] = Object.assign(
+                    legacy_nmaps['nmaps'],
+                    cfgObj['nmaps']
+                )
+        } catch (e) {
+            console.error('Error while migration from 0.0 -> 1.0', e)
```
bovine3dom: I get this error with the message "can't convert undefined to object" on every page that loads (on a fresh profile - i.e. starting config version is 1.1).

src/excmds.ts
```diff
@@ -327,6 +340,14 @@ export async function help(excmd?: string) {
     }
 }
 
+/**
+ * Navigate to the settings page.
+ */
+//#background
+export async function settings(): Promise<void> {
+    return browser.runtime.openOptionsPage()
```
bovine3dom: It would be nice if we could open the page on specific tabs - e.g. :changelog

src/settings_page.ts
```diff
@@ -0,0 +1,67 @@
...
+const rctextarea = <HTMLTextAreaElement>document.querySelector('#rc-textarea')
+const isAutoloadCheckbox =
+    <HTMLInputElement>document.querySelector('#toggle-fs-rc')
+
+// Start by showing current settings
+activateTab('#current-settings')
+fillCurrentConfig()
+fillRc()
+
+// Switch tab when a its clicked on the tabbar
```
bovine3dom: *Switch tab when the tab bar is clicked

or just "Switch settings tab on click"

src/static/settings.html
```diff
@@ -0,0 +1,70 @@
...
+        <a class="nav-link" role="tab" href="#current-settings">Current Settings</a>
+      </li>
+      <li class="nav-item">
+        <a class="nav-link" role="tab" href="#rc-text">.tridactylrc</a>
+      </li>
+      <li class="nav-item">
+        <a class="nav-link" role="tab" href="#about">About</a>
+      </li>
+      <li class="nav-item">
+        <a class="nav-link" role="tab" href="#license">License</a>
```
bovine3dom: In British English this is spelt Licence. License is commonly understood to mean "legally questionable wall of text that nobody every reads" so we could perhaps allow it.

Apologies for the pedantry.

src/static/settings.html
```diff
@@ -0,0 +1,70 @@
...
+        <a class="nav-link" role="tab" href="#readme">Readme</a>
+      </li>
+    </ul>
+  </nav>
+  <div id="tab-container" class="tab-content">
+    <div class="tab-panel" role="tabpanel" id="current-settings"></div>
+    <div class="tab-panel" role="tabpanel" id="rc-text">
+      <p>
+        If the button to load from filesystem is not working, check the console
+        to see the error you are getting. Yes, we don't have an error reporting
```
bovine3dom: I think this is a little too cheerful - I'd lose the "Yes,".

src/static/settings.html
```diff
@@ -0,0 +1,70 @@
...
+        mechanism right now.
+      </p>
+      <button id="btn-save-rc">Save to browser</button>
+      <button id="btn-load-rc">Load from browser</button>
```
bovine3dom: perhaps "Read config from text box"?

src/static/settings.html
```diff
@@ -0,0 +1,70 @@
...
+      <button id="btn-load-fs-rc">Load from filesystem</button>
```
bovine3dom: "Read config file from $XDG_CONFIG_HOME/tridactyl/tridactylrc"?

I think it also needs much more explanation saying that it requires a native messenger. Perhaps we can check to see if we can talk to the messenger on page load and grey out the button if we cannot, with some red text saying "Native messenger not found", etc?

I was 100% convinced your "Load from file system" was going to open a file picker dialogue.

src/static/settings.html
```diff
@@ -0,0 +1,70 @@
...
+  </nav>
+  <div id="tab-container" class="tab-content">
+    <div class="tab-panel" role="tabpanel" id="current-settings"></div>
+    <div class="tab-panel" role="tabpanel" id="rc-text">
+      <p>
+        If the button to load from filesystem is not working, check the console
+        to see the error you are getting. Yes, we don't have an error reporting
+        mechanism right now.
+      </p>
+      <button id="btn-save-rc">Save to browser</button>
```
bovine3dom: On closer inspection - I have no idea what this button does.

I have an RC file with `bind q fillcmdline QQQ` which I can load in using the third button.

If I click the first button, I get "Setting new RC file: bind q..." but it doesn't seem to do anything.

And if the first button is to load the configuration into the browser to make all the binds etc, what on earth does the second button do?

src/config.ts
IsaacKhor: Generally, ALL_CAPS are for CONSTANTS. At least that's the convention I've always used. That's not a global variable; it's a top-level variable in the module.

IsaacKhor:
Error handling has been improved; the page will `alert()` the user to the error.

native/native_main.py
IsaacKhor: The error handling is further down here:

```python
        file_content = getUserConfig()
        if file_content:
            reply['content'] = file_content
        else:
            reply['error'] = 'File not found'
```

src/config_rc.ts
```diff
@@ -0,0 +1,115 @@
...
+import * as Native from './native_background'
+import * as Messaging from './messaging'
+import DEFAULTS from './config_defaults'
+import Logger from './logging'
+
+const RC_NAME = 'rc-text'
+const RC_LOC_NAME = 'rc-location'
+const RC_AUTOLOAD_FS_NAME = 'rc-autoload'
+const RC_DEFAULT = `
+# This is the default RC file.
```
bovine3dom: I could have sworn that someone else mentioned this, but I cannot see it: Comments in VimL are started with ", not #.

src/config_rc.ts
```diff
@@ -0,0 +1,115 @@
...
+            rcText = await getBrowserRc()
+            break
+        default:
+            throw `Location '${location}' is not valid. Please contact a dev
+            and tell them to use either 'filesystem' or 'browser'.`
+    }
+
+    // If we are autoloading and we actually loaded something from
+    // the filesystem, then set the rc as the thing
+    if (isAutoload && location === 'filesystem') setBrowserRc(rcText)
```
bovine3dom: This doesn't seem to work for me. Manually clicking load from file system and then run worked, but the tickbox below doesn't seem to do anything (tested on a fresh profile by loading Tridactyl as a temporary addon, ticking box, restarting Firefox, loading Tridactyl again. The tickbox is preserved but settings are not loaded (test binds are not bound).)

bovine3dom: Also, could this be turned on by default? I think really we want to load from an RC file if present unless told otherwise, so that the minimum amount of faff is required to set up a new computer with Tridactyl and all of your settings.

src/config.ts
```diff
@@ -1,361 +1,153 @@
...
-// Listen for changes to the storage and update the USERCONFIG if appropriate.
-// TODO: BUG! Sync and local storage are merged at startup, but not by this thing.
-browser.storage.onChanged.addListener(
-    (changes, areaname) => {
-        if (CONFIGNAME in changes) {
-            USERCONFIG = changes[CONFIGNAME].newValue
+async function init() {
+    try {
+        const storage = await Utils.getStorage(CONFIG_NAME)
+        const cfg = storage ? storage : DEFAULTS
```
bovine3dom: Does this deal with https://github.com/cmcaine/tridactyl/pull/93#issuecomment-345971554 at all?

As far as I can tell, it stores all of the defaults in with the user configuration.

bovine3dom:
Closing in favour of #502.

PR #407, opened 2018-04-21
[RFC] Nativeuserchrome

bovine3dom:
Ultimate vision: `set gui` alike. (We'd need to add a listener to config to execute this script every time gui was changed).

Currently just a stub: the cssfiddle.py would have to live inside native_main.py, and be called using excmds in the usual fashion.

Having thought about it a bit more, a more sensible approach might be to just use the native messenger to read the CSS, and then parse, edit and stringify the CSS in JavaScript, before writing the file with the native messenger, as that will be easier to document and update.

Todo:
- [x] check if CSS actually needs to live in dependencies in package.json. It seems to work ok without it? I think I don't understand how webpack works.
- [x] support profiles without chrome / userChrome.css
- [x] add more options?
- [x] particularly, work out how to support meta options like gui=none
- [x] sort out proper backup to make sure that when we inevitably mangle someone's beloved userChrome, it isn't lost forever.
- [x] OSX support
- [x] dig up Vimperator's documentation...
- [x] ... and try to make sure most of what could be done with `set gui` is replicated
- [ ] consider being the worst and hacking in a special case for `set gui` that sends it to `guiset`
- [x] documentation. Particularly the restart requirement.

Bonus:
- [x] figure out how addon debugging's browser toolbox lets you fiddle with userChrome without restart and make an issue


bovine3dom:
Todo:
- [ ] check if CSS actually needs to live in dependencies in package.json. It seems to work ok without it? I think I don't understand how webpack works.
- [ ] support profiles without chrome / userChrome.css
- [ ] add more options? particularly, work out how to support meta options like gui=none
- [ ] sort out proper backup to make sure that when we inevitably mangle someone's beloved userChrome, it isn't lost forever.

Bonus:
- [ ] figure out how addon debugging's browser toolbox lets you fiddle with userChrome without restart and do that.

bovine3dom:
Also: winblows and cackintosh support, esp for profile finding hack. Should probably allow config to override it.

bovine3dom:
I reckon this is pretty much ready to be merged into master.

@cmcaine, what do you think? It might be worth releasing the native messenger for real and then having this sit in the beta for a while.

PR #432, opened 2018-04-26
Changelog: add dates and move to the project root

chocolateboy:
See e.g.:

* https://keepachangelog.com/en/1.0.0/
* https://github.com/tech-angels/vandamme/#format


bovine3dom:
Thanks for this, I had no idea there was a convention.

https://github.com/tech-angels/vandamme/blob/master/README.md seems to say that you should have a / between the release and the date, and that the file name should be all caps excluding the extension.

I'm happy to make those changes as I merge it later today.

chocolateboy:
Yes, the basename is usually upper case. I didn't want to push my luck :-)

The separator varies. The [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) site uses a hyphen; some use a [space (or a hyphen)](https://metacpan.org/pod/distribution/CPAN-Changes/lib/CPAN/Changes/Spec.pod#Required-Elements); some parenthesize the dates. Whatever you prefer. I don't mind as long as there's a date :-)

cmcaine:
LGTM :)

cmcaine:
Yep. Did you use a script to generate the headers from git tags? If so, can
you paste it here, please.

Issue #461, opened 2018-04-30
Close all tabs to the left/right


maxigaz:
I'd like to be able to close all tabs left/right of the current tab with a key combination, just like in Vimium.

(In Vimium, this feature can be accessed by pressing `gx0`, which closes all tabs to the left, and `gx$`, which closes all tabs to the right.)

bovine3dom:
I'd be happy to help anyone work on this. It should be quite simple.

The trickiest part will be deciding which tabs are to the left, and which are to the right. I suspect that the tabs array on Window https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/windows/Window has the tabs in order of their position, but I could be wrong.

cmcaine:
Tab objects come with an index, so we can just sort by that. Nae bother.

caputchinefrobles:
In vimperator rcfile I had something like:
js <<EOF

closeAllToRight = function () {
    var current = tabs.getTab();
    var currentIx = tabs.index(current);
    var nexttab = current.nextElementSibling;
    var N = tabs.count;
    var numToClose = N - (currentIx + 1);
    tabs.remove(nexttab, numToClose);
}

closeAllToLeft = function () {
    var current = tabs.getTab();
    var currentIx = tabs.index(current);
    var firsttab = tabs.getTab(0);
    var N = tabs.count;
    var numToClose = currentIx;
    tabs.remove(firsttab, numToClose);
}

EOF

command! closealltoright :js closeAllToRight()
command! closealltoleft :js closeAllToLeft()

" close tabs to left
map x. :js closeAllToRight()<Return>
" close tabs to right
map x, :js closeAllToLeft()<Return>

Maybe It could help with a future feature.

caputchinefrobles:
Is there anyone with this already? If not I'd like to take it if possible.

PR #472, opened 2018-05-04
Tabclose all right left

caputchinefrobles:
Implements #461 with commands tabcloseAllToLeft and tabcloseAllToRight. No default binding included as sugested (gx0, gx$).

Could be used like ":bind gx0 tabcloseAllToLeft".


bovine3dom:
Merged in 24d4563. I added binds (although I think maybe gd0 and gx$ might make more sense now I think about it...) and made the functions lowercase to fit in with the rest of the ex commands.

Thanks!

PR #476, opened 2018-05-04
Update tab page and other documentation to reflect new ignore mode binding

terminalmage:


terminalmage:
Also, because the `shiftKey` attribute is being used, this means that right-shift doesn't work. Might want to find a better way of doing this.

bovine3dom:
Thanks. I thought I'd found all of these... apparently I wasn't even close.

Issue #406, opened 2018-04-20
`hint -b` doesn't always open a new tab in the same container as current tab


WorldCodeCentral:
Just realized that when hitting a hint it's not always the same as a middle click. For example I use the Firefox Multi-Accounts Containers extension and it has always worked the same for me, middle click and hint hitting, but just now I found that in the one tab I have hint -b does not open in the current container, whereas just hint works in the current container and middleclick works as well (opens in a new background tab in the same container)

It's the first time it's happened to me though and I can't reproduce it :( So it's not really annoying

cmcaine:
Our code is not currently container aware. Your expected behaviour only occurs when the hint code successfully simulates a click on the hinted element and the firefox code handles all that jazz.

The code to change is `src/lib/webext.ts:openInNewTab()`

WorldCodeCentral:
Was able to reproduce :-) The problem was a toplevel domain prefix that I didn't see which took me to the same site though. But my container was only set up to activate automatically for the toplevel domain without the prefix so it didn't activate.

How containers work though is that you're never leaving the container if you click any links, even in new tabs. That is where the behavior differs if you use Tridactyl.

It's just one more thing on the list that can be striked off once Tridactyl can just reproduce the browser's behavior/methods instead of use it's own/web extension specific things :-)

antonva:
This is fixed as of (at least) 1.11.2-1.11.2.

bovine3dom:
@antonva really? I thought I hadn't released it to stable yet. I was expecting it to appear in version 1.12.0.

antonva:
Yeah, it's working on that version most definitely.

bovine3dom:
There's some minor issue with beta versioning, but it's definitely not in the stable release yet.

PR #499, opened 2018-05-08
container logic

antonva:
This is mainly a fix for the hint -b issue as well as adding some container logic to libwebext and openInNewTab.
As per discussion on matrix, I'm keeping the tabopen behaviour but behind a setting in the config.
Addresses #406

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -254,6 +254,10 @@ const DEFAULTS = o({
         "curl -fsSl https://raw.githubusercontent.com/cmcaine/tridactyl/master/native/install.sh | bash",
     profiledir: "auto",
 
+    // Container settings
+    // If enabled, tabopen opens a new tab in the currently active tab's container.
+    tabopencontaineraware: "false",
+
     // Performance related settings
 
     // number of most recent results to ask Firefox for. We display the top 20 or so most frequently visited ones.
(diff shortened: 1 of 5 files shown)


bovine3dom:
Do not merge until b35ed85 is released onto the AMO.

bovine3dom:
This works great, @antonva. Since it will be a while before I can merge this, would you like to have a go at implementing:
- changing the container of a tab via an excmd
   - updating help with an example of autocmds that do this automatically
- allowing `{,win,tab}open` to accept container arguments
- creating new identities via excmds?

Cheers.

antonva:
Yeah, no problem at all. I had already started working on implementing the basic context identities api as excmds.
I'll keep you posted here.

antonva:
Upon looking at the implementation of the multi account container addon from mozilla, there are some peculiarities as to how it deals with containers. I was expecting there to be a unique constraint on the combination of name, color and icon but there isn't. Meaning that it's impossible to address a container by anything but it's id.

I'm of the thought that there should be a unique constraint on all three even if we support containers that are managed elsewhere, in such that **containerCreate** and **containerUpdate** in our implementation would not allow you to create the exact same container with the only discernable difference (the cookieStoreId) being hidden from the end user. 

Case in point:
example

bovine3dom:
The warning is out in the wild in version 1.11.0, so this is no longer blocked. : )

PR #520, opened 2018-05-11
provide hostname hint in editor temporary files

larsks:
when calling an external editor, include the hostname of the remote site in
the generated temporary filename.

This is built on top of #519.

### DIFF
--- a/native/native_main.py
+++ b/native/native_main.py
@@ -154,7 +154,7 @@ def handleMessage(message):
 
     elif cmd == 'temp':
         (handle, filepath) = tempfile.mkstemp()
-        with open(filepath, "w") as file:
+        with os.fdopen(handle, "w") as file:
             file.write(message["content"])
         reply['content'] = filepath
 
--- a/native/native_main.py
+++ b/native/native_main.py
@@ -4,11 +4,14 @@
 import sys
 import os
 import json
+import re
 import struct
 import subprocess
 import tempfile
+import unicodedata
 
 VERSION = "0.1.3"
+
 
 class NoConnectionError(Exception):
     """ Exception thrown when stdin cannot be read """
@@ -24,6 +27,7 @@ def eprint(*args, **kwargs):
 def getenv(variable, default):
     """ Get an environment variable value, or use the default provided """
     return os.environ.get(variable) or default
+
 
 def getMessage():
     """Read a message from stdin and decode it.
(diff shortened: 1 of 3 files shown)


native/native_main.py
```diff
@@ -153,7 +158,12 @@ def handleMessage(message):
             file.write(message["content"])
 
     elif cmd == 'temp':
-        (handle, filepath) = tempfile.mkstemp()
+        prefix = message.get('hostname')
+        if prefix is not None:
+            prefix = sanitizePrefix(prefix)
```
bovine3dom: This should be in src/native_backgrounds.ts I think.

It is probably a good idea to sanitise things. A quick search found this which looks sensible: https://stackoverflow.com/questions/295135/turn-a-string-into-a-valid-filename?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa#answer-295466

native/native_main.py
```diff
@@ -153,7 +158,12 @@ def handleMessage(message):
...
+
+        (handle, filepath) = tempfile.mkstemp(
+            prefix='moz-edit_{}_'.format(prefix))
```
bovine3dom: should probably default to "'tri-tmp" as it might be used for stuff other than editing

src/native_background.ts
```diff
@@ -238,8 +238,8 @@ export async function mkdir(dir: string, exist_ok: boolean) {
     return sendNativeMsg("mkdir", { dir, exist_ok })
 }
 
-export async function temp(content: string) {
-    return sendNativeMsg("temp", { content })
+export async function temp(content: string, hostname: string) {
+    return sendNativeMsg("temp", { content, hostname })
```
bovine3dom: Please do the sanitisation here.

bovine3dom:
Thanks for these PRs, @larsks.

If you would make the changes I've suggested, I'll merge both PRs at once and bump the native script version number.

native/native_main.py
```diff
@@ -153,7 +158,12 @@ def handleMessage(message):
             file.write(message["content"])
 
     elif cmd == 'temp':
-        (handle, filepath) = tempfile.mkstemp()
+        prefix = message.get('hostname')
```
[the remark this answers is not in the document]
larsks: I agree with making it more general, but left to my own devices I would be tempted to leave the sanitizing in `native_main.py`, since this is the code that is actually opening the file.

larsks:
Second attempt, which partially implements your comments.  It seems to me that the right place to perform the filename sanitizing is in the code that is actually opening the file, but if you feel strongly otherwise let me know and I'll try to figure out the python `->` javascript translation of that routine and stick it into `native_background.ts`.

larsks:
It looks as if the editor support doesn't clean up the temporary files.  Looking to see if that's an easy fix right now.

bovine3dom:
> the right place to perform the filename sanitizing is in the code that is actually opening the file

I see where you're coming from, but the native script is only ever accessed via native_background, so you can consider that to be the code that's opening the file if you want.

My reasoning is that it's a reasonably complicated bit of code that we might well find some nasty bug in, so it should go in the main extension code where we can get an update out to people reasonably quickly.

bovine3dom:
> It looks as if the editor support doesn't clean up the temporary files. Looking to see if that's an easy fix right now.

We've currently left that as an exercise for the operating system. I like being able to get back to the files for a bit if I need to.

larsks:
I am not all that familiar with Javascript. It looks as if it may be stupidly difficult to do the equivalent of this python:

```
fn = unicodedata.normalize('NFKD', fn).encode('ascii', 'ignore').decode('ascii')                                      
```

Am I wrong?

bovine3dom:
I have no idea. We can leave it in the python script :)

Issue #479, opened 2018-05-04
Windows support for native messenger (i.e, open-in-Vim and friends)


bovine3dom:
I haven't bothered looking into this yet (mostly because our build chain doesn't work on Windows, so it would be a minor faff to test).

If anyone feels up to the task, I'd be happy to help.

gsbabil:
Hi @bovine3dom ,

I have access to both Linux and Windows 10 machines, and would like
to take the task.

Currently I have a Windows set-up with Python 3.6, and can install
the `install.sh` script with some minor modifications around `mkdir`
using [MinTTY][0].

The Python messenger script i.e. `native_main.py` also seems to run
fine. Although, Firefox (I'm on Beta `61.0b4 (64-bit)`) can't
interact with it on Windows, I think due to restrictions coming from
within Tridactyl.

I think on Windows we should aim at using [Powershell][1] as the
installation script, as compared to the current Bash script, or a
old-school Windows Batch script. I don't think we want to move away
from Python as the messenger script, although on Windows we might be
able to provide compiled Windows binary straight out of the Python
messenger script, i.e. `messenger_main.exe`.

Please share some initial pointers so that I can start implementing
and testing the Installation script under Windows. I'm sure I'll
have more questions, and I'll get back to you with them as I
progress.

Thanks.


[0]: https://github.com/mintty/mintty
[1]: https://en.wikipedia.org/wiki/PowerShell

bovine3dom:
@gsbabil great!

MDN has a guide here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Native_messaging

I'd suggest you install the manifest mentioned in that guide and the python script manually before we worry about automating that procedure. I agree that we should keep the python script the same.

Tridactyl has a few places that check if native is trying to run on windows and then quits. I think they're all in native_background.ts. There's some OS specific stuff dotted around but we can't worry about that later. For now, just try to get `:native` working.

gsbabil:
Hi @bovine3dom,

> MDN has a guide here: https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Native_messaging

> I'd suggest you install the manifest mentioned in that guide and
> the python script manually before we worry about automating that
> procedure. I agree that we should keep the python script the same.

Thanks. Will do.

A very immediate step could be to just patch the `install.sh` to add
support [MinTTY][1]/[MSYS2][2], which seems to ship with Bash
`4.4.19`.

[...]

Issue #48, opened 2017-11-09
Support for building on Windows


bovine3dom:
- [ ] change all shell scripts to pure JS

- [ ] make all paths OS independent

- [ ] fix git commit hooks

Building in Cygwin used to work until recently, but it would be good if pure Windows would work too.

cmcaine:
Relative paths with forward slashes will work fine, I think.

Cygwin will probably work except for the symlinks?

bovine3dom:
There's no Node for Cygwin so I was using the Windows binary, meaning that it can't find some things such as bash... but some stuff did work.

bovine3dom:
Nobody has asked for this for a while, so I'm going to close this issue.

bovine3dom:
```
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules\fse                                                               vents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@                                                               1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"}                                                               )
```

These aren't errors - they're warnings. You can safely ignore them : )

```
    - PowerShell major version >= '3' required, '2' found, quitting ...
```

is because the build process also installs the native messenger, which uses a PowerShell script on Windows. Please update your version of PowerShell (although it probably isn't technically needed for a local installation).


Could you open a new issue to discuss these poorly documented dependencies? I don't quite understand what you mean about Python.

reversebreak:
I'm the kind of person that compiles with -Wall, y'know?  If those are ignorable, then that's OK.

I learned something today!  Apparently the version of the powershell shell and powershell engine are different and independent!  You can have the fancy newer shell installed and yet still be running on the old one internally...
I updated and that's fixed now.

I'm not sure I've got enough to open an issue on in regards to the Python dependency.  I didn't think to write down which package or run time thing required it, and now that it's fixed, I'm not sure I can easily find it again.  
All I can say is that it was looking by default for a specific sub-version of 3.6 Python exe in a default Program Files location, then when that failed looked for the PYTHON env variable.  
I might try removing the PYTHON environment variable and see what happens then, once I'm done with #711

gsbabil:
@reversebreak @bovine3dom

Off the top of my head, the building Tridactyl on Windows seems to
require the following:

1. [PowerShell][0]
    - Recommended version >= `3`

[...]

PR #532, opened 2018-05-14
Extend native-messenger support to Windows

gsbabil:
Dear Maintainers,

As discussed with @bovine3dom on #479, the patches below would allow
building Tridactyl with [MinGW][0]/[MSYS2][1] on Windows, and extend
native-messenger support to Windows.

Both `:installnative` and `:updatenative` commands currently seems
to work fine on Windows 10, while the install/update internals are
handled via the new Powershell script added in `native/win_install.ps1`.

Hope this helps. Thank you.

[0]: http://www.mingw.org
[1]: https://www.msys2.org


native/win_install.ps1
```diff
@@ -0,0 +1,214 @@
...
+# Constants
+#
+$win_tridactyl_dir_name = ".tridactyl"
+$batch_file_name = "native_main.bat"
+$manifest_file_name = "tridactyl.json"
+$manifest_replace_str = "REPLACE_ME_WITH_SED"
+$native_messenger_binary_name = "native_main.py"
+
+$native_messenger_files_http_base = `
+    "https://raw.githubusercontent.com/cmcaine/tridactyl/master/native"
```
bovine3dom: We should allow for a "local" installation that copies the native messenger from the ./native directory, so that people can develop the native messenger on Windows, or at least exit the script with a warning that the stable release is installed.

scripts/build.sh
```diff
@@ -2,25 +2,52 @@
...
 scripts/make_docs.sh &
-nearleyc src/grammars/bracketexpr.ne > src/grammars/.bracketexpr.generated.ts
-native/install.sh local
 
-(webpack --display errors-only && scripts/git_version.sh)&
+nearleyc src/grammars/bracketexpr.ne \
+  > src/grammars/.bracketexpr.generated.ts
+
+if [ "$(isWindowsMingw)" == "True" ]; then
+  powershell native/win_install.ps1
```
bovine3dom: Please make this a local installation as I have mentioned on the PowerShell script.

scripts/build.sh
```diff
@@ -2,25 +2,52 @@
```
bovine3dom: We should update the documentation on readme.md to say that #48 is fixed if you use MINGW.

native/win_install.ps1
```diff
@@ -0,0 +1,214 @@
...
+#
+$batch_file_path = [string]::Format("{0}\{1}",
+    $win_tridactyl_dir_path,
+    $batch_file_name)
+Write-Host "[+] Preparing $batch_file_path ..."
+
+$batch_file_content = @"
+@echo off
+
+call python -u $win_tridactyl_dir_path\native_main.py
```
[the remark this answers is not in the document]
bovine3dom: Also, we should check that Python is on the path in the installer.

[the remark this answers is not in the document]
gsbabil: Hi @bovine3dom ,

> Does this always call python 3?

Things seem a bit different on Windows around Python 2/3 (as
compared to on Unix systems). In short, even `Python3` tends to
install _only_ as `Python.exe` in default location such as
`C:\Python36`.  There does not seem to be any `Python3.exe` after a
vanilla installation process.

There are various instructions on [their official documentation][1]
to point to the desired version of Python on Windows by setting
`PYTHONPATH`, `PYTHON_HOME` environment variables etc. They also
seem to have this thing called [Python Launcher for Windows][0],
which is meant to _"automagically"_ pick the correct version of
Python for the user.

What I have found is that [Python-Launcher][0] tends to install
itself as `%WINDIR%\py.exe`, which allows setting `-2` or `-3`. If
Python scripts are invoked by `py -3 script.py`, as compared to more
common `python3 script.py`, they will run _only_ as Python 3. So in
the [latest revision][2] of the Powershell script, I have [gone with
`py -3`][4].

> Also, we should check that Python is on the path in the installer.

In the [latest revision][2] of the Powershell installer script, I
have [enforced checks for Python 3][3].

Thanks.

[0]: https://docs.python.org/3/using/windows.html#launcher
[1]: https://docs.python.org/3/using/windows.html
[2]: https://github.com/gsbabil/tridactyl/commit/062137b54f3d99d7541fca681036ff3bcc12f256
[3]: https://github.com/gsbabil/tridactyl/blob/062137b54f3d99d7541fca681036ff3bcc12f256/native/win_install.ps1#L444
[4]: https://github.com/gsbabil/tridactyl/blob/062137b54f3d99d7541fca681036ff3bcc12f256/native/win_install.ps1#L238

native/win_install.ps1
```diff
@@ -0,0 +1,214 @@
...
+
+#Get-ItemProperty `
+#    -Path $manifest_registry_path `
+#    -Name "(Default)"
+
+if ((Get-ItemProperty `
+    -Path $manifest_registry_path)."(Default)" = $manifest_file_path)
+{
+    Write-Host "    - Registry key looks good!"
+}
```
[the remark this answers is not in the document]
gsbabil: [Done!][0]

[0]: https://github.com/gsbabil/tridactyl/blob/f301f6549c43aeace44e7e7e2daa1b2e70ce7af5/native/win_install.ps1#L493

native/win_install.ps1
gsbabil: [Done!][0].

If the script is invoked as shown below, it will copy
`native_main.py` and `tridactyl.json` from the specified directory.

This is somewhat similar to `bash install.sh local`, but allows
picking any target directory, as well as `native` to read files
from.

Thanks.

```
  PS C:\Users\<user-name>\tridactyl> .\win_install.ps1 -DebugDirBase .\native
```

[0]: https://github.com/gsbabil/tridactyl/blob/f301f6549c43aeace44e7e7e2daa1b2e70ce7af5/native/win_install.ps1#L180

scripts/build.sh
gsbabil: I think this is fixed as mentioned above.

The script also currently supports an additional [flag called
`-InstallDirBase`][0] to drop the installation files to any target
directory. The default is `%USERPROFILE%\.tridactyl`.

For example, the command below will install `native_main.py`,
`tridactyl.json` etc. in `C:\Users\<user-name>\Desktop\.tridactyl`
directory.

```
PS C:\Users\<user-name>\tridactyl> .\win_install.ps1 `
  -InstallDirBase C:\Users\<user-name>\Desktop
```

There's also [an `-Uninstall` flag][1], that removes these files and
the registry.

Example:

```
PS C:\Users\<user-name>\tridactyl> .\win_install.ps1 -Uninstall
```

Or,

```
PS C:\Users\<user-name>\tridactyl> .\win_install.ps1 `
  -Uninstall -InstallDirBase C:\Users\<user-name>\Desktop
```

[0]: https://github.com/gsbabil/tridactyl/blob/f301f6549c43aeace44e7e7e2daa1b2e70ce7af5/native/win_install.ps1#L111
[1]: https://github.com/gsbabil/tridactyl/blob/f301f6549c43aeace44e7e7e2daa1b2e70ce7af5/native/win_install.ps1#L42

bovine3dom: I think installing it locally from source should be done at this step i.e. `  PS C:\Users\<user-name>\tridactyl> .\win_install.ps1 -DebugDirBase .\native`.

bovine3dom:
Thanks for this, @gsbabil.

It will take me a little while longer to review this than I had hoped as I actually managed to brick my computer by bungling an install of MSYS2 yesterday... So testing this is blocked by me fixing my Windows machine :(

scripts/build.sh
```diff
@@ -2,25 +2,52 @@
...
+fi
+
+export PATH
+
+mkdir -p build
+mkdir -p build/static
 mkdir -p generated/static
 mkdir -p generated/static/clippy
+
 scripts/excmds_macros.py
```
bovine3dom: This needs to be `py -3 scripts/excmds_macros.py`, else I get

```
olie@OLIE-PC MINGW64 ~/projects/tridactyl (gsbabil-master)
$ npm run build

> tridactyl@0.1.0 build C:\Users\olie\projects\tridactyl
> sh scripts/build.sh

/usr/bin/env: ‘python3’: No such file or directory
```

So we should just put in a little guard like you have done on line 16, or perhaps just `WINPYTHON="py -3"` in that if statement and then `$WINPYTHON scripts/excmds_macros.py`.

scripts/build.sh
bovine3dom: I get

```
$ npm run build

> tridactyl@0.1.0 build C:\Users\olie\projects\tridactyl
> sh scripts/build.sh

native/win_install.ps1 : File C:\Users\olie\projects\tridactyl\native\win_install.ps1 cannot be loaded because running
scripts is disabled on this system. For more information, see about_Execution_Policies at
https:/go.microsoft.com/fwlink/?LinkID=135170.
At line:1 char:1
+ native/win_install.ps1
+ ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : SecurityError: (:) [], PSSecurityException
    + FullyQualifiedErrorId : UnauthorizedAccess
```

We should add a warning to the README or run `Set-ExecutionPolicy -ExecutionPolicy Unrestricted` in an elevated prompt.

gsbabil:
@bovine3dom 

> So testing this is blocked by me fixing my Windows machine :(

Good luck! :-)

I'm assuming that you are not a regular Windows user, and the
easiest in this case might be to install Windows under Virtual
Box/VMware, and keep a snap-shot after the fresh install so that you
can always get back to the clean state whenever needed.

Thanks.

gsbabil:
Hi @bovine3dom,

> We should add a warning to the README or run Set-ExecutionPolicy
> -ExecutionPolicy Unrestricted in an elevated prompt.

I have added a new commit [here][0] to fix this during the
_installation_ phase by setting the `ExecutionPolicy` to `Bypass`
with the `-Scope` set as `Process` only.

From development point of view, the easiest for us would indeed be
to suggest the user to open up an elevated Powershell prompt _i.e._
"Run as Administrator", and then copy/paste the line found via
`:installnative`. All we'll have to do is to update the `-Scope`
parameter from [this patch][0] to achieve this, and both
`:installnative` and `:updatenative` should work fine from there
onward. However, I feel a bit guilty to advise this since it'd
weaken user's system security along the process.

In my understanding, currently there are two challenges around this
on Windows:

  1. During _installation_ of native messenger (`:installnative`)
  2. During _update_ native messenger (`:updatenative`)

We have solved the challenge at [1] by [this][0]. 

Regarding [2], is it possible
to modify the Python script (`native_main.py`) so that it will
download a newer version of itself if it receives an `updatenative`
message?

My rationale for this is that we need to drop the manifest JSON, the
wrapper Batch script and set up registry only as part of the
_installation_ process. The _update_ process don't really need to
touch these files, except the core Tridactyl files (which can
auto-update), and the `native_main.py` Python file (which I believe
Tridactyl core can signal after the auto-update).

I'm somewhat new to this code-base. So please let me know what you
think.

Also, in `excmds.ts` I see that we have a line like the following as
part of the update process:

```Javascript
    await Native.run(await config.get("nativeinstallcmd"))
```

Is `Native.run` executing a real system shell i.e. Sh, Bash?

Thank you.

[0]: https://github.com/gsbabil/tridactyl/commit/1b0e7ce14f881dd1688f042852f4a21d3114bb69

bovine3dom:
@gsbabil `installnative` and `updatenative` work fine with `-ExecutionPolicy Restricted`. The problem only comes about when you're doing `npm run build`.

bovine3dom:
>Also, in excmds.ts I see that we have a line like the following as part of the update process:
>   `await Native.run(await config.get("nativeinstallcmd"))`
>   Is Native.run executing a real system shell i.e. Sh, Bash?

It uses subprocess.check_output with Shell=True: https://docs.python.org/3.6/library/subprocess.html#id1

On Windows, that means cmd.exe, which is probably our problem: we need to change the `nativeinstallcmd` on Windows to tell PowerShell to launch.

bovine3dom:
> Regarding [2], is it possible to modify the Python script (native_main.py) so that it will download a newer version of itself if it receives an updatenative message?
> 
> My rationale for this is that we need to drop the manifest JSON, the wrapper Batch script and set up registry only as part of the installation process. The update process don't really need to touch these files, except the core Tridactyl files (which can auto-update), and the native_main.py Python file (which I believe Tridactyl core can signal after the auto-update).

Probably, but I think we just need to make sure that `updatenative` works when called from cmd.exe (i.e, it just needs to have the word `powershell ` prepended to it).

src/config.ts
```diff
@@ -254,8 +254,10 @@ const DEFAULTS = o({
     browser: "firefox",
     nativeinstallcmd:
         "curl -fsSl https://raw.githubusercontent.com/cmcaine/tridactyl/master/native/install.sh | bash",
-    win_nativeinstallcmd:
+    win_powershell_nativeinstallcmd:
         "Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/gsbabil/tridactyl/master/native/win_install.ps1'))",
```
bovine3dom: I don't think you need to change the ExecutionPolicy here, as the `Invoke-Expression` means that this is basically like copying and pasting in the script line by line interactively. ExecutionPolicy applies only to scripts that are written to files, I think.

src/config.ts
```diff
@@ -254,8 +254,10 @@ const DEFAULTS = o({
...
+    win_cmdexe_nativeinstallcmd:
+        '@"%SystemRoot%\\System32\\WindowsPowerShell\\v1.0\\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "iex ((New-Object System.Net.WebClient).DownloadString(\'https://raw.githubusercontent.com/gsbabil/tridactyl/master/native/win_install.ps1\'))"',
```
bovine3dom: I don't think ExecutionPolicy is needed here, either.

scripts/build.sh
```diff
@@ -2,25 +2,57 @@
...
 scripts/make_docs.sh &
-nearleyc src/grammars/bracketexpr.ne > src/grammars/.bracketexpr.generated.ts
-native/install.sh local
 
-(webpack --display errors-only && scripts/git_version.sh)&
+nearleyc src/grammars/bracketexpr.ne \
+  > src/grammars/.bracketexpr.generated.ts
+
+if [ "$(isWindowsMinGW)" == "True" ]; then
+  powershell native/win_install.ps1 -DebugDirBase native
```
bovine3dom: This is where we need the `Set-ExecutionPolicy` bit : )

gsbabil:
@bovine3dom 

> Probably, but I think we just need to make sure that updatenative
> works when called from cmd.exe ...

Just submitted the fix for `:updatenative` on Windows [here][0]. Now
we have both Powershell and Cmd.exe variant of the commands
available. Currently we have Powershell command for
`:installnative`, and Cmd.exe version for `:updatenative`.

Please have a look. Thanks.

```Javascript
{
  ...
  win_powershell_nativeinstallcmd: "Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/gsbabil/tridactyl/master/native/win_install.ps1'))",
  win_cmdexe_nativeinstallcmd: '@"%SystemRoot%\\System32\\WindowsPowerShell\\v1.0\\powershell.exe" -NoProfile -InputFormat None -ExecutionPolicy Bypass -Command "iex ((New-Object System.Net.WebClient).DownloadString(\'https://raw.githubusercontent.com/gsbabil/tridactyl/master/native/win_install.ps1\'))"',
    ...
}
```

[0]: https://github.com/gsbabil/tridactyl/commit/dbeab7deb3e58621f666b15694410123728197a1#diff-541faaaa3b5d3979be7ea476ef8a1cf2

src/config.ts
gsbabil: This is weird :-S. But I think you are right.

The `-Scope Process` part I think could only do damage to the currently running PowerShell instance. Would you still like to have the following part removed?

```Powershell
  Set-ExecutionPolicy Bypass -Scope Process -Force;
```

P.S.

I see that [Chocolatey][0], a new popular package manager for Windows, recommends [non-administrative installation][1] with `Set-ExecutionPolicy Bypass`.

[0]: https://chocolatey.org
[1]: https://chocolatey.org/install#non-administrative-install

gsbabil:
@bovine3dom

> I don't think we have any bash dependency for Windows, other than for building.

Right. I actually meant Bash for Unix systems, and Powershell on Windows. I put them in one category since we are using these as installer-helpers.

> PowerShell dependency should be a non-issue as Firefox requires Windows 7+, and PowerShell is built-in to all Windows O/Ses after that.

I didn't know that PowerShell was default on all new Windows. Good to know! Thanks.

> What is causing you concern?

I guess this leaves me with only one _"concern"_ - Python :-).

I know most Unix systems are shipping with Python 3, and most Tridactyl users should know how to get Python on Windows. So this is nearly a non-issue, but in my ideal world, I guess I would have preferred a _"single-click"_ installer for Windows users with all dependencies. I'm currently looking at [PyInstaller][1] and [CxFreeze][2]. Will update if I make progress with these.

To be honest, I use Python regularly for work and other things. So I don't mind keeping it up-to-date on all my systems. But an average Windows user in this case would need the whole Python 3 distribution just to run `native_main.py`. Then again, an average Windows user will most likely _not_ use Tridactyl. I hope I am wrong. :-)

```Markdown

> PS: GitHub has a strange markdown variant in the reply boxes where
> linebreaks are rendered as linebreaks; you don't need to put them
> in yourself unless you're replying by email.

Sorry about that. I edit most of my texts in Vim (then copy/paste
where it must go). Fixed-width mono-space is my preferred font
setting, and I have `:set textwidth=68` default for most things I
write, which is [supposed to improve readability][0]. Apparently
short lines a good for focused reading, while long lines are useful
for skimming. :-)

[0]: https://en.wikipedia.org/wiki/Line_length

```
Thank you.

[1]: https://www.pyinstaller.org
[2]: https://anthony-tuininga.github.io/cx_Freeze/

src/config.ts
gsbabil: Looked around a bit, and it seems that this execution-policy restriction in PowerShell is a sham. I have removed the unneeded ones in [this patch][0]. Thanks.

[0]: https://github.com/gsbabil/tridactyl/commit/1e572c5ba5726f2789457bf15c695ff973a5d24d

src/config.ts
gsbabil: Fixed [here][0].

[0]: https://github.com/gsbabil/tridactyl/commit/1e572c5ba5726f2789457bf15c695ff973a5d24d

bovine3dom:
```
Sorry about that. I edit most of my texts in Vim (then copy/paste
where it must go). Fixed-width mono-space is my preferred font
setting, and I have `:set textwidth=68` default for most things I
write, which is [supposed to improve readability][0]. Apparently
short lines a good for focused reading, while long lines are useful
for skimming. :-)
```

Just inject `p {max-width: 30rem; text-align: justify; -moz-hyphens: auto}` with Stylish into every page and you're golden :)

(an aside: an `ex` is the width of a single character, and `2.3ex == 1em`, hence `68em \approx 30em`, and `rem` is a magic unit that scales with the current font).

---

This PR looks good now. I'll do some final testing && merge it when I'm next by a Windows box.

gsbabil:
Hi @bovine3dom,

> Edit: p {max-width: 68ex; text-align: justify; -moz-hyphens: auto; margin: auto} works better. I forgot which of rem and em was relative to the current font. It's em, obviously, the r in rem stands for not Relative, I guess.

Thanks for the hint on `rem`. I never used it before. I have been using `ch` instead under Stylish so far. I'll definitely give it a go.

> This PR looks good now. I'll do some final testing && merge it when I'm next by a Windows box.

Great! Thanks.

P.S.

I have done some more reading on [CSS font styling and lenghts][0], and [testing with `ex` vs. `ch`][1]. It looks like even `ch` is a lie :-S, i.e. it's not the number of characters, rather the width of the character zero (`0`)! Example screenshot [here][2]. Test code [here][2].

Thanks.

[0]: https://css-tricks.com/the-lengths-of-css
[1]: https://codepen.io/gsbabil/pen/rvqwML
[2]: https://imgur.com/a/XBKyxZr

bovine3dom:
@gsbabil rebased and merged. Thanks so much for this. It would have been done much more shoddily and slowly if it had been left to myself and @cmcaine. :heart:

gsbabil:
@bovine3dom @cmcaine 

Thanks for this wonderful add-on. Like many others, I'm a longtime Vim and Vimperator user. So I guess you'd know what an add-on like this could mean to users like us. :-)

I'm happy to have been able to contribute a little bit. Thank you!

PR #553, opened 2018-05-17
Fix native_main for python < 3.5.0

Werkov:
We are passing None value to mkstemp which is only possible since python
3.5.0 [1]. Let's use our own mask with empty string as the prefix when
None is given. That works with both python API versions.

[1]
https://github.com/python/cpython/commit/ad577b938b367da53ee19d6d5021b19e50b92873


bovine3dom:
Rebased and merged. Thanks!

PR #559, opened 2018-05-17
Handle (well, ignore) an extra colon before any excmd

ELLIOTTCABLE:
In Vim proper, initial `:` entered before a command-line are entirely ignored. This has a few benefits:

 - When new users see an example command on the web (“Use the command `:set searchengine duckduckgo` to …”), they may not understand that the initial colon is intended as an activation-key, instead as a part of the command that they are to paste into their command-line. With this change, the command will still work as-intended.

 - When I'm trying to construct a more complex command (at least in Vim-proper), I often open the command-line *while thinking*; then, occasionally, I forget that I'd done so, when I finally settle on an approach to the problem at hand, and thus reflexively begin my complicated ex-command with a colon, intended to open the command-line.

This patch, my first to y'all's project, is a simple one to address this. Initial colons are ignored in the ex-mode command-line.

I tried to fairly-thoroughly test this by hand, after building the extension. Let me know if you want me to exercise it in some automated fashion.

As this is my first, I'd love some feedback before you merge — similarly, my commits were small and granular; but this is a very small work, so I'm likely to push these squashed into a single commit once you give me the go-ahead! `<3`

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -279,7 +279,7 @@ function process() {
 
     hide_and_clear()
 
-    const [func, ...args] = command.trim().split(/\s+/)
+    const [func, ...args] = command.trim().split(/ +/)
 
     if (func.length === 0 || func.startsWith("#")) {
         return
--- a/src/aliases.ts
+++ b/src/aliases.ts
@@ -3,8 +3,9 @@ import * as config from "./config"
 /**
  * Expands the alias in the provided exstr recursively. Does nothing if
  * the command is not aliased, including when the command is invalid.
+ * Returns another string.
  *
- * @param exstr :exstr typed by the user on the commantd line
+ * @param exstr :exstr typed by the user on the command line
  */
 export function expandExstr(
     exstr: string,
@@ -12,11 +13,37 @@ export function expandExstr(
     prevExpansions: string[] = [],
 ): string {
     // Split on whitespace
-    const [command, ...args] = exstr.trim().split(/\s+/)
+    let [command, ...args] = exstr.trim().split(/ +/)
+
+    // Handle initial (extra) colon
+    if (command[0] === ":") {
+        command = command.substring(1)
+    }
+
+    return expandExarr([command, ...args]).join(" ")
+}
+
+/**
+ * Expands the alias in the provided exstr recursively. Does nothing if
+ * the command is not aliased, including when the command is invalid.
+ * Expects, and returns, an array of `command, ...args`.
+ *
+ * @param exstr :exstr typed by the user on the command line
+ */
+export function expandExarr(
+    exstr: string | string[],
+    aliases = config.get("exaliases"),
+    prevExpansions: string[] = [],
+): string[] {
+    const exarr: string[] =
+        typeof exstr === "string" ? exstr.trim().split(/ +/) : exstr
+
+    const command = exarr[0],
+        expansion = aliases[command]
 
     // Base case: alias not found; return original command
-    if (aliases[command] === undefined) {
-        return exstr
+    if (expansion === undefined) {
+        return exarr
     }
 
     // Infinite loop detected
@@ -27,10 +54,10 @@ export function expandExstr(
     // Add command to expansions used so far
     prevExpansions.push(command)
 
+    // Remove the current command being expanded, and replace it with
+    // the command expansion (possibly exploded into multiple words.)
+    exarr.splice(0, 1, ...expansion.trim().split(/ +/))
+
     // Alias exists; expand it recursively
-    return expandExstr(
-        exstr.replace(command, aliases[command]),
-        aliases,
-        prevExpansions,
-    )
+    return expandExarr(exarr, aliases, prevExpansions)
 }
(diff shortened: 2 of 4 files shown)


ELLIOTTCABLE:
(On an unrelated note, but not really worthy of a separate Issue — I spent half-an-hour or so *extremely* confused by your git tooling repeatedly committing *every* change in my working-directory, instead of just those that I'd staged for committing. While I can understand the rationale behind such automated and silent linting, as a fellow maintainer and tooling dev, might I suggest you instead simply error-out and deny the commit, instructing the committer to manually run `npm lint` and solve any errors before committing? That leaves new contributors like me a little less confused, especially if they're new. Just a thought!)

bovine3dom:
@ELLIOTTCABLE thanks for this PR. I agree that the problem that you're trying to solve is real.

I am kind of of the opinion that we should solve this with CSS and a ::before selector, which we can use to show a pretend colon. @cmcaine, what do you think?

If that's not possible then I think this approach is OK.

bovine3dom:
@ELLIOTTCABLE half an hour of your time is definitely worth a separate issue. Sorry that we confused you.

bovine3dom:
I think 7add357 is a nicer solution to your first problem - new users misunderstanding what the colon is for.

I'm open to merging this anyway as (to my surprise) Vim does let you just put in any number of colons before your command and I don't want to break your workflow :)

ELLIOTTCABLE:
Sounds good! Rebasing and pushing.

It's not actually critical to my workflow, or anything — I had just been keeping my eye out for low-hanging fruit, so I could get started contributing to Tridactyl. This is the first / easiest thing I saw. `:3`

ELLIOTTCABLE:
Rebased! This should be good to go. `<3`

bovine3dom:
This doesn't work for me.

`:help` -> `# Error: Not an excmd: :help`.

Can you confirm?

Also, I think we should allow any number of colons in front of the command, like Vim.

src/commandline_frame.ts
```diff
@@ -279,7 +279,7 @@ function process() {
 
     hide_and_clear()
 
-    const [func, ...args] = command.trim().split(/\s+/)
+    const [func, ...args] = command.trim().split(/ +/)
```
bovine3dom: I think the "colon ignoring" should go here, really. Colon is a message to the command line that we want the line to be parsed as if it was an ex-command. I don't think it's reasonable to expect aliases to have colons ignored.

cmcaine: Prepended colons become part of the excmd language with this PR. Their handling should go in src/parsers/exmode.ts:parser().

In vim prepended colons are ignored in maps, composite statements, etc: `:echo "hi" | :::echo "bye"`

cmcaine: It's not really anything to do with aliases, IMO. Aliases should be expanded, then colons should be removed, then other parsing things.

ELLIOTTCABLE:
Taking a look at this now — looks like I might need to do a little more spelunking to get this Right; but that's the point, nah? To find my way around y'all's work for the first time, prior to moving on to bigger and better things? 🤣

Another tooling-nitpick, by the way; every time I `git pull cmcaine master`, your `hooks/post-merge` tries to execute `npm install`, which repeatedly causes `package-lock.json` to be updated … which in turn causes my original `git pull` to fail with “Cannot rebase: You have unstaged changes.” every time. Gotta keep remembering to notice that, `git reset --hard cmcaine/master`, etc!

Re: the location, I'll take one more stab. Feel free to just dump my work and add what is, I'm sure, a one-liner, if one of y'all decides to knock it out yourself, if my next submission still nags atcha!~

bovine3dom:
> Another tooling-nitpick, by the way; every time I git pull cmcaine master, your hooks/post-merge tries to execute npm install, which repeatedly causes package-lock.json to be updated … which in turn causes my original git pull to fail with “Cannot rebase: You have unstaged changes.” every time. Gotta keep remembering to notice that, git reset --hard cmcaine/master

It should only run npm install when there are changes in package.json, which shouldn't be very often :S. Sounds like a bug.

ELLIOTTCABLE:
I finally got around to this properly (new puppy's keeping me very busy!). So, the reason I originally used `aliases.ts`, was that multiple callsites that didn't seem to call into `parsers/exmode.ts` used it, giving me one central place to make the change.

I can easily move the relevant line to `parsers/exmode.ts`; but now that I've done a little more in-depth look through your code, it looks like this would leave the client-side completion code in `commandline_frame.ts` out in the cold? It uses `aliases.ts#expandExstr` directly, without detouring through the parser. If I understand correctly, this would leave ex completion broken in the edge case where the user both accidentally prepends a colon, and has an alias defined?

Seems like enough of an edge-case in a non-critical system, though. Pushing your requested changes, which you can cherry-pick as you please!

bovine3dom:
@ELLIOTTCABLE were you eaten by a puppy before you got a chance to push it? : )

bovine3dom:
Closing due to inactivity, and because it will probably come for free with #789.

ELLIOTTCABLE:
YIKES totally fell through the cracks

welp have photo of puppy as substitute for working code. 😇

[rocky](http://ell.io/iegBlwK)

PR #560, opened 2018-05-18
Compiled native messenger binary (native_main.exe) support for Windows

gsbabil:
Dear Maintainers,

The following patch(es) will add compiled native binary support on Windows. Compilation turns `native_main.py` into `native_main.exe` using [PyInstaller][0] for Windows systems.

By default, the Powershell installer script would use the compiled EXE version for Windows. However, the old behaviour can achieved by using the `-UsePython` flag to `win_install.ps1`.

Currently, the `native_main.exe` is built every time `scripts/build.sh` is invoked. However, this behavior can be adjusted by invoking the build script as shown below:

```
PYINSTALLER="0" npm run build
```

Hope this helps. Thank you.

[0]: https://www.pyinstaller.org


native/win_install.ps1
```diff
@@ -1,5 +1,6 @@
 Param (
     [switch]$Uninstall = $false,
+    [switch]$UsePython= $false,
```
bovine3dom: Aren't .exe files more likely to trigger antivirus / corporate "don't trust foreign executables" policies than the Python script?

In which case, I think it would make more sense to default to using Python unless `py -3` isn't on the path, in which case we should fall back to the executable.

What do you think?

gsbabil: Hi @bovine3dom 

> Aren't .exe files more likely to trigger antivirus / corporate "don't trust foreign executables" policies than the Python script?

IMHO, this is more of a "perception" issue, and not completely unbiased. We have pre-compiled binaries even on popular [Linux][1] and [BSD][2] distributions these days. So, "compiled-binary" I think is becoming a norm even outside of Windows now.

Besides, the "trust-chain" has to bootstrap somewhere, and very few people actually download and compile their own [CPython][3] or the OS kernel. So, essentially the argument becomes somewhat like "we don't trust _'your'_ EXE (_i.e._ native_main.exe), but we trust _'their'_ EXE (_i.e._ python.exe)".

AFAIK, the best way to break this "logjam" is via [code signing][7]. I have just checked in binary signing and verification support with OpenPGP [here][4]. Please have a look. 

Windows has their own signing implementation via ["sigtool"][8]. But I think [OpenPGP][9] is more trustworthy, open-source and cross-platform friendly.

-------------------------------------------------------------------

> In which case, I think it would make more sense to default to using Python unless py -3 isn't on the path, in which case we should fall back to the executable.

My rationale for preferring EXE over PY on Windows was that we'll not only get to remove the Python dependency, we'll also get to avoid getting blames for Tridactyl problems when Python installation on Windows changes, breaks _etc._

Please note that Python on Windows is _not_ a very common set-up. I mean that it might be common within the Tridactyl community, considering that we agree that Tridactyl/Vim users are slightly more technical than the average, and are more likely to have Python on their systems. But given prior observation my guess is that a vast majority of Windows users are not likely to have Python. So I think it really boils down to who we define to be our target user-base.

With that said, I'm okay to go with the following strategy:

  - As default, look for `py -3` when invoking `win_install.ps1` on Windows

      - If Python is detected, we'd install `native_main.py`
      - if Python is _not_ detected, we'd install `native_main.exe`

As said, the risk here is that we remain dependent on the Python installation. Anytime, Python breaks, Tridactyl would also break.

Please confirm. I can submit another patch to realise the above strategy. Thanks.

-------------------------------------------------------------------

_P.S._

I did some _very quick_ research on extensions on Firefox/Chrome using native-messenger, and I find that most of them guide users to install a binary messenger, even on Mac. Here's a list of bunch of extensions using [binary native messenger][5]. They seem to be using [this][6] native-client written in Node.

Hope this helps. Thanks.

[1]: https://www.debian.org/distrib/packages
[2]: https://www.openbsd.org/faq/ports/ports.html
[3]: https://github.com/python/cpython
[4]: https://github.com/gsbabil/tridactyl/commit/2f8689cf3d78c85e104a243801813c8231751b09
[5]: https://add0n.com/open-in.html
[6]: https://github.com/andy-portmen/native-client
[7]: https://en.wikipedia.org/wiki/Code_signing
[8]: https://msdn.microsoft.com/en-us/library/windows/desktop/aa387764(v=vs.85).aspx
[9]: https://www.openpgp.org/software

bovine3dom: Thanks for such a comprehensive reply. I don't think we need to bother with code signing - people can just build the exe themselves and compare the checksums, which seems like fewer steps. Do you disagree?

My concern wasn't really security so much as wanting people to be able to use Tridactyl in as many places as possible. I know lots of workplaces block the running of unknown binaries (it's built into lots of antivirus software these days) so I think a python script is more likely to bypass that as, as you said, python.exe is trusted.

A related point is that we should tell users that if the native exe is blocked by their work, they should try installing python and then running our install script again.

I'm not worried about python breaking - Tridactyl will just tell people to reinstall, and that should fix it.

So, yes, please let's go with falling back to the exe.


PS: re Tridactyl users being representative of the general population - 2/3rds of them use Linux where I think the actual prevelance is 0.5%... so we're a weird bunch :)

gsbabil: Hi @bovine3dom,

Apologies. This is another long reply. Please take a deep breath and see below. :-)

#### ## Prefer Python over Compiled EXE

> - As default, look for py -3 when invoking win_install.ps1 on
>   Windows
>
>     - If Python is detected, we'd install native_main.py
>     - if Python is not detected, we'd install native_main.exe

I just committed the default strategy discussed above [in this commit][0].  Please have a look.

[0]: https://github.com/gsbabil/tridactyl/commit/a940453958babe2932f8e07b9b355aab6b0c8458

-------------------------------------------------------------------

#### ## EXE Distribution and Code-Signing

> I don't think we need to bother with code signing - people can just build the exe themselves and compare the checksums, which seems like fewer steps. Do you disagree?

I'm sorry, but I'd have to disagree with this given the following. Please see details below. 

1. [Circular Dependency][2]

    - [PyInstaller][1] is a Python module, and usually installed via `python3 -m pip install --upgrade pyinstaller`

    - So, if the user wanted to compile and then verify the `native_main.exe`, they would need a functional Python 3 environment first. And if the user had Python 3 installed already (ready for PyInstaller compilation), they wouldn't need the `native_main.exe`.

2. **Non**-[Reproducible Build][3]

    - PyInstaller compiled binaries seem to be non-reproducible, meaning that if we compiled the exact same `native_main.py` at two different points in time even on the same machine, the EXE will have the same effective functionalities as the unchanged `native_main.py`, but will have slight variations in the output EXE, leading to drastically different SHA-256 digest.

    - The "non-reproducible" behavior can be verified on Windows by running the following command:

        ```
         PYTHONHASHSEED=1 pyinstaller \
          --clean \
          --console \
          --onefile \
          --noupx \
          --noconfirm \
          --workpath native \
          --distpath native 'native/native_main.py' 2> /dev/null \
          && sha256sum.exe native/native_main.exe
        ```

    - Please note that I have tried using the same `PYTHONHASHSEED=1` value advised in the [PyInstaller documentation][4], but it does not seem to work as described on PyInstaller version `3.3.1` _i.e._ SHA-256 hashes are still different.

In summary, if we agree to provide compiled `native_main.exe` for the Tridactyl users, I think we should _really_ adhere to code-signing. It would be the same reason AMO requires us to submit signed ZIP before official release. The GPG signature and SHA-256 sums should be released together with `native_main.exe`, and it should be up to the user to perform the verification.

[1]: https://www.pyinstaller.org
[2]: https://en.wikipedia.org/wiki/Circular_dependency
[3]: https://en.wikipedia.org/wiki/Deterministic_compilation
[4]: https://pyinstaller.readthedocs.io/en/stable/advanced-topics.html#creating-a-reproducible-build

-------------------------------------------------------------------

#### ## Windows Support and Development Directions

> Tridactyl users being representative of the general population - 2/3rds of them use Linux where I think the actual prevelance is 0.5%... so we're a weird bunch :)

I think we are getting pretty close to the hard questions below:

  1. How much support would we like to provide for Tridactyl users on Windows?
  2. Would Windows users get the same functionalities as the Linux/Mac users with Tridactyl?

Right now, I don't think Windows is the same first-class citizen that a Linux/Mac system is with Tridactyl. I think a clear decision around this will make future development directions much more coherent.

I think the active owners/maintainers of this project (I believe you and @cmcaine) should set the course for this sooner rather than later. I'm happy to listen and contribute where I can!

Hope this helps. Thank you.


_P.S._

I have submitted a bug report with PyInstaller [here][5] regarding reproducible-build. It seems that their current PyPI version `3.3.1` is buggy, but their [latest master][6] works fine. However, even if we eliminate the "reproducible-build" issue, the "circular-dependency" issue in my opinion still stands, and I'd still strongly vote in for code-signing. :-)

Thank you.

[5]: https://github.com/pyinstaller/pyinstaller/issues/3526
[6]: https://github.com/pyinstaller/pyinstaller/commit/1033a8770ab96cc8fa0f2d3047b2db3db9dcda43

bovine3dom:
> I think we are getting pretty close to the hard questions below:
> 
> #. How much support would we like to provide for Tridactyl users on Windows?
> #. Would Windows users get the same functionalities as the Linux/Mac users with Tridactyl?
> 
> Right now, I don't think Windows is the same first-class citizen that a Linux/Mac system is with Tridactyl. I think a clear decision around this will make future development directions much more coherent.
> 
> I think the active owners/maintainers of this project (I believe you and @cmcaine) should set the course for this sooner rather than later. I'm happy to listen and contribute where I can!

I was saying that 2/3rds of Tridactyl users use Linux to point out that we're strange people - I think Python will be much more prevalent on our Windows machines than it is in the general population.

But Windows is definitely a second-class citizen. @cmcaine and I develop using Linux so that will be what is tested. The reason it works on OSX is mostly because very little had to be changed rather than a lack of focus; neither of us have access to a real Mac. I actually use Windows a fair bit because it has much better font hinting than Linux, and more video games - so I can test stuff on that.

I briefly toyed with the idea of having a Windows build bot so that we would at least see if we broke the Windows build procedure, but I can't find any free ones. It would be cool to set one up with geckodriver...

We're very happy to support Windows as much as possible, though, and very happy to merge PRs that help it :)

---

I'll read through and respond to the rest tomorrow. My gut reaction to the "circular dependency" is that it is indeed one, but your alternative solution is to install an even less common piece of software (at least, on Windows machines), so even though your dependency isn't circular, it's still circuitous. I'm not against merging it, though, provided we can think of a sensible way of getting the right answer to people? I'm not sure what threats we're defending against:

- if cmcaine or I have breached accounts on GitHub, the buildbot will merrily sign anything that is pushed to this repository as a beta, so signing the executable is no help there.
- releases to AMO are done manually and would require that the build bot itself is breached - so I guess they would be kept safe?

gsbabil:
Hi @bovine3dom,

> I'm not sure what threats we're defending against:
>

I'd answer this differently: by providing a signed binary, you would be vouching about the authenticity and integrity of the specific release of `native_main.exe`. The idea is the same as the signed ZIP files for AMO submission. Anything that goes against the spirit of verifying the authenticity and integrity of a binary-blob received from the trusted sender (in this case you and @cmcaine) is a potential threat.

So, an unsigned binary is a threat as in I can never 100% be sure if I have it exactly the way you had shipped, _bit by bit_.

Frankly, the similar insecurity applies to the "curl-bash" combo, but it is quickly becoming a dangerous norm. Running the "curl-bash" combo is little different than downloading and running an unsigned binary.

>   if cmcaine or I have breached accounts on GitHub, the buildbot
>   will merrily sign anything that is pushed to this repository
>   as a beta, so signing the executable is no help there.

The GPG signing _private_-key should _never_ leave your's or @cmcaine's personal computers (same as the AMO keys), and should never be put on Github. I'm not entirely sure how your current build-bot is set-up. But I'd never prefer Github to have access to _any_ of my private keys. 

The signing should be done on one of the maintainer's machines _i.e._ your or @cmcaine's laptop. After that, when you'd perform `git push`, only the signature and _public_-key would be put on Github. From that point onward, users can download the `native_main.exe`, `native_main.exe.sig`, and `native_main.exe.sha256` as needed, and verify authenticity/integrity before use.

I have all these steps integrated into the new `build.sh::signWindowsCompiledNativeBin()`. You can see my GPG generated [signature][3], [hash][2], [public-key][1], and the rest of the files [here][0]. If you want to verify the signature of the compiled `native_main.exe` from my repo, you can run the following series of commands. I've already added the Powershell version of the commands below in [README.md][4] for Windows users.

```
$ gpg2 --armor --import native/gsbabil-pub.asc
$ gpg2 --armor --verify native/native_main.exe.sig native/native_main.exe
$ sha256sum --check native/native_main.exe.sha256
```

> We're very happy to support Windows as much as possible, though, and very happy to merge PRs that help it :)

I think that would be great, and would bring parity to the original Vimperator project which we all loved dearly, and currently seeking refuge in Tridactyl. I'd be happy to help test/patch Windows/Linux specific issues where I can.

Thank you!

[0]: https://github.com/gsbabil/tridactyl/tree/gsbabil/windows-compiled-native-messenger-support/native

[1]: https://github.com/gsbabil/tridactyl/blob/gsbabil/windows-compiled-native-messenger-support/native/gsbabil-pub.asc

[2]: https://github.com/gsbabil/tridactyl/blob/gsbabil/windows-compiled-native-messenger-support/native/native_main.exe.sha256

[3]: https://github.com/gsbabil/tridactyl/blob/gsbabil/windows-compiled-native-messenger-support/native/native_main.exe.sig

[4]: https://github.com/gsbabil/tridactyl/commit/2f8689cf3d78c85e104a243801813c8231751b09

bovine3dom:
> The GPG signing private-key should never leave your's or @cmcaine's personal computers (same as the AMO keys), and should never be put on Github. I'm not entirely sure how your current build-bot is set-up.

The build bot is a physical machine that we have access to. It tries to pull from GitHub every five minutes or so and if there are any changes it makes a new beta release and publishes it on its web server. So my point is that if you trust the signed betas, you also trust that our GitHub accounts are not breached and there is nothing malicious on the repository, so you also trust the executable.

bovine3dom:
Tangential to our chat about code signing - I'm not sure compiling the executable should go in the npm run build step, as that's not actually the proper build - that's done in scripts/sign.

cmcaine:
We take security quite seriously, but one compromise we make in the name of ease is explicitly trusting github and our buildbot's connection to github for the beta releases. We also obviously have to trust Mozilla.

Currently the .xpis are primarily protected by the AMO key and the native messenger script is just us trusting github and your connection to github (so DNS system, your CA database, global CA system, github's ISP (I don't think curl does certificate pinning), github's machines).

If we don't sign the .exe and just host it on our buildbot then it gets a similar level of protection to files hosted on github and so our security situation is not really degraded (the only extras are our ISP and our machine).

We could remove our reliance on the whole https:// security system with gpg or whatever, but if your https:// security is broken you've got bigger problems than possibly installing a bad tridactyl binary.

We can talk about further securing the delivery of tridactyl artefacts in another issue, but we consider this to be 1) fairly low priority, 2) worth doing properly if we're going to do it at all.

We don't consider it a blocker on this PR.

cmcaine:
I think the way forward is to take the signing stuff out and we'll consider it in a different PR.

gsbabil:
@bovine3dom

> Tangential to our chat about code signing - I'm not sure compiling the executable should go in the npm run build step, as that's not actually the proper build - that's done in scripts/sign.

I'm fine with moving `signWindowsCompiledNativeBin()` to `scripts/sign`. Please see further comments below.

@cmcaine

> I think the way forward is to take the signing stuff out and we'll consider it in a different PR.

The native binary compilation and signing can currently be turned off with the following. The default invocation of `build.sh` currently would perform both compilation and signing if `PYINSTALLER=0` is not set.

```
$ npm run build ## compiles and signs `native_main.exe`
$ PYINSTALLER=1 npm run build ## compiles and signs `native_main.exe`

$ PYINSTALLER=0 npm run build ## does _not_ compile/sign `native_main.exe`

```

Would you prefer it as `PYINSTALLER=1 npm run build` only to perform compile/signing, and anything else _i.e._ `npm run build` or `PYINSTALLER=0 npm run build` to skip compilation/signing? It'd be an easy change to `build.sh`, and I think the code here can still be useful later.

I can submit another patch to this PR to achieve the following strategy:

  1. We don't set `PYINSTALLER=1`, call `npm run build` as before, and nothing changes given the previous behavior.

  2. If `PYINSTALLER=1` is set, both the compilation and signing are performed.

      - **Compilation**: `native_main.exe` is produced by `scripts/build.sh`

      - **Signing**: `native_main.exe.sig` and `native_main.exe.sha256` are produced (by `scripts/sign` as per @bovine3dom's comment above)

Please confirm. Thanks.

gsbabil:
@cmcaine @bovine3dom 

> If we don't sign the .exe and just host it on our buildbot then it gets a similar level of protection to files hosted on github and so our security situation is not really degraded (the only extras are our ISP and our machine).

Apologies, I'm not entirely familiar with buildbot - is this a physical machine you own, or is this something Github provides in the cloud? If this is a machine you own and _**you trust it**_, I think the _signing step_ is fine to be done on this machine. As said, I would not trust Github with any of the private keys.

> Edit: my thinking is that the executable should be hosted on tridactyl.cmcaine.co.uk/betas which will happen magically if you just plonk it in web-ext-artifacts/

I think _hosting_ of the beta builds, or other build artifacts can be anywhere. However, _**signing**_ must be done _**on only the machine you can fully trust**_, because that machine is going to need access to your private-key(s).

> We could remove our reliance on the whole https:// security system with gpg or whatever, but if your https:// security is broken you've got bigger problems than possibly installing a bad tridactyl artefact.

I think we could have both - HTTPS would ensure security at the transport layer, while the GPG signature can ensure security (authenticity/integrity/verification) off-line and end-to-end without relying on any trusted 3rd-party _i.e._ the certification authorities, _even after_ the transportation of data has taken place.

Currently, I'm signing `native_main.exe` on my personal machine, but hosting the signed binary, signatures etc. on Github.

> We can talk about further securing the delivery of tridactyl artefacts in another issue, but we consider this to be 1) fairly low priority, 2) worth doing properly if we're going to do it at all.

I think we _should_ do it and _do it properly_ :-) Love your project, and happy to help where I can!

Thank you.

gsbabil:
@bovine3dom @cmcaine

> The build bot is a physical machine that we have access to. It tries to pull from GitHub every five minutes or so and if there are any changes it makes a new beta release and publishes it on its web server. So my point is that if you trust the signed betas, you also trust that our GitHub accounts are not breached and there is nothing malicious on the repository, so you also trust the executable.

Sorry I think I missed @bovine3dom's comment above.

It sounds like you own this machine and it can be trusted. I think signing is safe to be done on this machine in that case.

As said above, signing and hosting can be done independently.  Signing _must_ be done on a _trusted_ machine, _hosting_ can be done on _any_ machine, even on Github.

Thanks.

gsbabil:
@cmcaine @bovine3dom 

Just to be clear, could you please confirm the following:

  1. We are okay to have the EXE _i.e._ `native_main.exe`, but we don't want to get into GPG signing business of the EXE right now.

  2. We do not want the whole EXE business right now.

- If it is **`[1]`**, I think the least we should do is to officially distribute the SHA-256 hash _i.e._ `native_main.exe.sha256` together with the EXE. We may even publish the hash via `README.md`.

- If it is **`[2]`**, I think this can be end of the conversation on this thread, and this PR can be trashed.

As said, the ZIP/XPI is signed. So it offers verifiability outside of the HTTPS trasnport loop. The native-messenger binary is the questionable piece, and it deserves end-to-end verifiability. I'm inclined to say the same for the Python script, but I think the risk is somewhat "softened" by the plain-text nature of the file.

Either way, please confirm the direction you'd like to take. Thanks.

bovine3dom:
It's 1. We appreciate your efforts with this. Happy to consider signing in a different PR now but at a slower pace than this PR.

Re this PR - please move generation of executable to scripts/sign or a new script as we only want it to run on the build bot.

gsbabil:
@bovine3dom 

> It's 1. We appreciate your efforts with this. Happy to consider signing in a different PR now but at a slower pace than this PR.

Sounds great. Thanks, and no problem. 

> Re this PR - please move generation of executable to scripts/sign or a new script as we only want it to run on the build bot.

With [this commit][0], currently I have the following. Basically, if the build-bot calls `npm run build`, no EXE compilation or GPG signing would be performed. 

```
  $ # following does _not_ compile/sign `native_main.exe`
  $ npm run build
  $ PYINSTALLER=0 npm run build

  $ # following _does_ compile `native_main.exe` (no GPG signing)
  $ PYINSTALLER=1 npm run build
```
_**This is basically the same behavior as we had before this PR.**_ So, the buildbot should not have any problem to auto-compile/build stuffs.

Regarding generating EXE with buildbot, I do _not_ think you can do it considering the buildbot is a non-Windows machine. I don't think PyInstaller supports cross-compilation. 

So, in order to generate the EXE, you'd have to run the following on a MinGW/MSYS2 system:

```
$ py -3 -m pip install pyinstaller
$ PYINSTALLER=1 npm run build
```
Let me know if this is okay with you. Thanks.

[0]: https://github.com/gsbabil/tridactyl/commit/ad3e2d51f7753c56836fb351d2dcbe13d0bfe548

cmcaine:
Thank you very much :)

The [PyInstaller FAQ](https://github.com/pyinstaller/pyinstaller/wiki/FAQ) suggests we can build under wine, which would be much more preferable to us than adding a manual build step or another machine.

Email thread from 2011 with code to cross compile: https://groups.google.com/forum/#!topic/pyinstaller/veq3BlA_Bns

Probably way overcomplicated project from 2011 that provides a web service for generating .exes from python files: https://bazaar.launchpad.net/~jacksonc/python-packager/devel/files/head:/Build-Server/

I guess I'll have a look at this tomorrow if no one else plays with it.

bovine3dom:
For clarity - this is blocked until we figure out how to use pyinstaller under Wine.

gsbabil:
Hi @bovine3dom,

I just added [`scripts/wine-pyinstaller.sh`][0] that prepares the required PyInstaller environment under Wine. The current requirements to run the script are as shown below. These are available both at Debian Testing or Ubuntu Bionic repositories.

  1. p7zip-16.xx
  2. Wine-3.xx
  3. WinPython-3.5.4

Thanks.

[0]: https://github.com/gsbabil/tridactyl/blob/507cd0515865225290ce3492f8205bf854c9d7c9/scripts/wine-pyinstaller.sh

cmcaine:
Hero :D

gsbabil:
> Hero :D

:smiley:

scripts/wine-pyinstaller.sh
```diff
@@ -0,0 +1,172 @@
...
+    --distpath "${OUTDIR}" \
+    "${DLDIR}/native_main.py"
+
+  ## Test the compiled EXE
+  colorEcho "[+] Checking compiled binary ...\n"
+  OUTFILE="${OUTDIR}/native_main.exe"
+
+  if [ -f "${OUTFILE}" ]; then
+    python3 \
+      "${DLDIR}/gen_native_message.py" version \
```
bovine3dom: There's a few problems with this:

- gen_native_message.py needs `chmod +x` 
- gen_native_message.py has Windows line endings which breaks env:
```fish
wine-pyinstaller/downloads/gen_native_message.py native
/usr/bin/env: ‘python3\r’: No such file or directory
```
- Once I fixed all that, gen_native_message.py gives
```fish
wine-pyinstaller/downloads/gen_native_message.py native
Traceback (most recent call last):
  File "wine-pyinstaller/downloads/gen_native_message.py", line 39, in <module>
    val = sys.argv[i].strip().split(separator)[1]
IndexError: list index out of range
```
I think because you've forgotten your own syntax: If I change this to `"${DLDIR}/gen_native_message.py" cmd..version` it works:

```fish
wine-pyinstaller/downloads/gen_native_message.py cmd..version | wine wine-pyinstaller/dist/nativ
e_main.exe
{"version": "0.1.6"}0030:fixme:ver:GetCurrentPackageId (0x32fe24 (nil)): stub
```

scripts/wine-pyinstaller.sh
```diff
@@ -0,0 +1,172 @@
+#!/bin/bash -e
```
bovine3dom: Please mark this file as executable (`chmod +x`).

scripts/wine-pyinstaller.sh
```diff
@@ -0,0 +1,172 @@
...
+  colorEcho "[+] Installing Pip ...\n"
+  wine python.exe "${DLDIR}/get-pip.py"
+
+  colorEcho "[+] Installing PyInstaller ...\n"
+  wine python.exe -m pip install --upgrade pyinstaller
+
+  ## Compile with PyInstaller
+  colorEcho "[+] Compiling with PyInstaller under Wine ...\n"
+  rm -rf "${OUTDIR}"
+  PYTHONHASHSEED=1 wine Scripts/pyinstaller \
```
bovine3dom: Why the capital S on scripts?

Issue #276, opened 2018-01-06
]] doesn't work on hacker news


bovine3dom:
https://news.ycombinator.com

Koushien:
I asked on the gitter, but is there any issue adding `more` to the existing pattern?

bovine3dom:
Should be fine provided it is low priority compared to the other patterns.

Also, I didn't see that message - we should consider saying that matrix is the canonical chat and all others are somewhat unreliable.

Koushien:
It shows up on the Riot which should robustly be 1:1 to the Matrix?
https://matrix.to/#/!xmubtggllUtLrQiGyn:matrix.org/$1515551891395580VxyvQ:matrix.org
Noted otherwise.

The current implementation looks for the last anchor matching either of the `next` patterns. Firing in the dark, I suppose a concern would be a comment or post section that has a "more" link further below an actual page advancement. Does anyone have an example of such a page? Otherwise I think it's worth it to do a quick patch until someone encounters a bad case (and speaks up about it).

chocolateboy:
> According to the help text, Vimperator checks patterns against links in order and uses the first match.
> Would that behavior be preferred if we wanted to introduce priority?

Vimperator's documentation doesn't match its [implementation](https://github.com/vimperator/vimperator-labs/blob/d3c61fa63e7924686e2598514db661cf6b753643/common/content/buffer.js#L552-L615) (which does match the [last](https://github.com/vimperator/vimperator-labs/blob/d3c61fa63e7924686e2598514db661cf6b753643/common/content/buffer.js#L591) fallback link), and its implementation is broken (it doesn't support multiple `rel` values (e.g. `rel="next nofollow"`) and thinks `rev` is a synonym for `rel` (it's the opposite).

Having wasted a lot of time trying to get the pager documentation and bugs fixed in Vimperator (and Pentadactyl), I'm pretty confident there's nothing in that hard-to-follow function that hasn't been learned from and improved here.

> I suppose a concern would be a comment or post section that has a "more" link further below an actual page advancement. Does anyone have an example of such a page?

Yes, Google has a "[Learn more](https://i.imgur.com/J3gdv4F.jpg)" link at the bottom of the page, which is why I [removed it from the list](https://github.com/cmcaine/tridactyl/pull/120).

Incidentally, in general I recommend fixing the site in cases like this, which is [easy to do](https://github.com/chocolateboy/userscripts#pagerizers). Trying to fix it in the fallback pattern just moves the breakage around.

PR #120, opened 2017-11-23
Improve pagination

chocolateboy:
a) Prefer the canonical forms over fallback patterns (fixes rockpapershotgun.com, Discourse etc.):

1. link[rel]
2. a[rel]
3. a::text(pattern)

b) Tighten fallback patterns (fixes Google, GitHub etc.)

c) Remove site-specific patterns

PR #569, opened 2018-05-19
[WIP/RFC] Improve pagination by fixing the websites themselves

bovine3dom:
Related: #120 #276

I'm wondering whether we should have a separate "site-specific fixes" part to Tridactyl. Things might get a bit messy if we just have a bunch of autocmds.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -140,7 +140,15 @@ const DEFAULTS = o({
     }),
     autocmds: o({
         DocStart: o({
-            // "addons.mozilla.org": "mode ignore",
+            /** chocolateboy's pagination fixes */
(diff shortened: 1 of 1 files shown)


src/config.ts
```diff
@@ -140,7 +140,15 @@ const DEFAULTS = o({
     }),
     autocmds: o({
         DocStart: o({
-            // "addons.mozilla.org": "mode ignore",
+            /** chocolateboy's pagination fixes */
```
bovine3dom: @chocolateboy points out that we should add a link back to https://github.com/chocolateboy/userscripts#pagerizers

PR #571, opened 2018-05-19
Fix path to docs in :help command

tbroadley:
I just installed Tridactyl and noticed that, when I use the `:help` ex-command, I'm taken to a page with with this error message:

```
Firefox can’t find the file at moz-extension://<snip>/static/docs/modules/_excmds_.html#.
```

This PR fixes the issue by opening `_src_excmds_.html` instead. I've tested it manually and it seems to work!


bovine3dom:
Thanks!

PR #574, opened 2018-05-19
Call get when there's no value to set Fixes #573

### LINKED ISSUE #573: make :set key (without value) print the current value
In vim, saying `:set key ` without saying a value will cause vim to show the current value.  For example, I can open a file and say `:set enc` and it will print the current encoding.  Of course, I can also say `:set encoding=utf-8` to set the value of `encoding` to `utf-8`.

In vimperator (and, I think in pentadactyl), set worked the same way.  I could say `:set searchengine` and see what `searchengine` was currently set to.  I'm on tridactyl 1.12.0-9-g82c4348 and I find that saying `set $KEY` without providing a value just causes tridactyl to complain to me:

```
# Error: Both key and value must be provided! 
```

jcrowgey:
This is simple, but untested (I didn't see any test code in this repo).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1894,8 +1894,11 @@ export function searchsetkeyword(keyword: string, url: string) {
 */
 //#background
 export function set(key: string, ...values: string[]) {
-    if (!key || !values[0]) {
-        throw "Both key and value must be provided!"
+    if (!key) {
+        throw "Key must be provided!"
+    } else if (!values[0]) {
+        get(key)
+        return
     }
 
     const target = key.split(".")


bovine3dom:
@jcrowgey please change the issue number to the right one, #573, otherwise this will close a random PR when I merge it :)

jcrowgey:
@bovine3dom Oh goodness, good catch!  I edited the PR title.  Is that enough or should I do some git magic to rewrite history with a new commit message?

bovine3dom:
Rebased, commit message edited, and merged in 71e1ed7. Thanks :)

PR #579, opened 2018-05-20
[WIP] New theme structure

bovine3dom:
Tidied up version of #537.

Todo:

- [x] new tab page, tutor, mode indicator should all get themed automatically
- [x] document how to add a theme on contributing.md


cmcaine:
@caputchinefrobles, @bovine3dom, @atrnh 

Great work everybody!

cmcaine:
If a theme doesn't just use variables then injecting it into web content as we do is problematic: the class names in the commandline are likely to collide with web content.

I guess we'll just have to namespace in the commandline as well? We can give the root a special class on the commandline page, for example.

With all the required prefixes it would be nicer to use SASS or some other CSS pre-processor.

Or if `@-moz-document` we maybe use that, but we'd have to dynamically add the UUID, which is a pain.

Issue #451, opened 2018-04-28
Opening GUI applications from the terminal in OSX is weird


bovine3dom:
1. `firefox` doesn't accept --new-tab arguments (even though it claims to). The solution is to write some apple script probably. [Relevant BMO issue](https://bugzilla.mozilla.org/show_bug.cgi?id=393645).

2. Most terminals don't accept commands to run from the command line. The solution again is to write some applescript, probably.

bovine3dom:
The first point would fix #538: `nativeopen` needs to point to some applescript that accepts some of the same arguments that `firefox` does on Unix.

sirn:
For 1, I could get Firefox to open `about:` pages in macOS with (one-liner, argsify):

```shell
$ osascript -e 'on run argv' -e 'tell application "Firefox" to open location item 1 of argv' -e 'end run' about:addons
```

If you don't mind, I can try working on 1.

bovine3dom:
@sirn that would be great :) I heard tales of people just using `open -a firefox [link]` and it working, so you might want to try that too.

sirn:
@bovine3dom I tried that, but apparently `open -a Firefox <link>` has special handling for `https?://`, otherwise `open` will treat the argument as file and error with something among the line of:

```
$ open -a firefox about:addons
The file /Users/sirn/about:addons does not exist.
```

So I think AppleScript is the only way… I'll make a PR!

barnabee:
In response to your point (3) @bovine3dom, [Kitty](https://sw.kovidgoyal.net/kitty/) terminal works really nicely for me. It's also the terminal I use day-to-day, so I guess I got lucky!  

The simplest `editorcmd` that I think works is this one:
```
:set editorcmd /Applications/kitty.app/Contents/MacOS/kitty --wait-for-single-instance-window-close vim %f "+call cursor(%l,%c)"
```

I like to customise the window size and give it a nice title (though a `%t` for the page title would make this better), as below. You can temporarily override any config option you like on the command line so it's really configurable. 
```
:set editorcmd /Applications/kitty.app/Contents/MacOS/kitty -o initial_window_width=720 -o initial_window_height=480 -o remember_window_size=no --wait-for-single-instance-window-close -T "Web input" vim %f "+call cursor(%l,%c)"
```

Screenshot:
<img width="832" alt="image" src="https://user-images.githubusercontent.com/1179257/162596159-5b9cc00f-af20-4017-bd70-d9070806309d.png">

PR #592, opened 2018-05-21
Support macOS in nativeopen

sirn:
Make `nativeopen` works with macOS by leveraging AppleScript to open a new tab in currently active Firefox window. See also #451

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -284,14 +284,17 @@ export async function fixamo() {
  * **Be *seriously* careful with this: you can use it to open any URL you can open in the Firefox address bar.**
  *
  * You've been warned.
- *
- * Unsupported on OSX unless you set `browser` to something that will open Firefox from a terminal pass it commmand line options.
  */
 //#background
 export async function nativeopen(url: string, ...firefoxArgs: string[]) {
-    if (firefoxArgs.length === 0) firefoxArgs = ["--new-tab"]
     if (await Native.nativegate()) {
-        Native.run(config.get("browser") + " " + firefoxArgs.join(" ") + " " + url)
+        if ((await browser.runtime.getPlatformInfo()).os === "mac") {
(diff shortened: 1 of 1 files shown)


bovine3dom:
Does this deal with
- new tabs
- new windows
- and new private windows?

If not, is there any way of doing that?

sirn:
* new tabs: the AppleScript opens in a new tab
* new windows: t will open a new tab in a currently active window
* new private windows: I just tested and it doesn't seems to open in private window 😞

I'll look if there's a way to workaround private window issue.

bovine3dom:
> new windows: t will open a new tab in a currently active window

I'm asking about `winopen` :)

sirn:
Oops, sorry I misunderstood.

I just tested with `winopen` and it doesn't seems to open a new window. It's possible to workaround using keystroke simulation to open a new window before opening a URL (pressing Cmd+N), but detecting that new window is already native is a bit tricky, but I'll look into it.

Another issue is for private window, which doesn't seems to response to `open location`. Using keystroke simulation can workaround it… but there's a slight flash in the UI while the key is being typed:

output

bovine3dom:
Nah, that's too janky, I think.

Having `tabopen` work most of the time is win enough for me.

If you feel like it, you could add a `nativegate` in a sensible place to `winopen` and `tabopen` if it's a private window, and then use `fillcmdline_notrail("# nativeopen isn't supported. Please install Linux or Windows.")` to politely inform the user why what they just tried to do didn't work.

sirn:
Sorry for taking really long. I tried to get `:winopen` to work but turns out detecting the window is already activate is really tricky. The only somewhat reliable way to detect window require the user to allow Firefox access macOS accessibility feature, so I decided to tell the user to install Linux in cmdline instead (for now).

bovine3dom:
I improved the warnings a little bit (to explain that `nativeopen` works in some circumstances), and merged.

Thanks : )

PR #593, opened 2018-05-21
FIx/improve greenmat theme on newtab/tutor

caputchinefrobles:


src/static/themes/greenmat/greenmat.css
```diff
@@ -35,11 +38,8 @@
...
   --tridactyl-hint-bg: rgba(0, 0, 0, 0.25); 
   --tridactyl-hint-outline: 2px solid var(--tridactyl-hintspan-bg);
-}
-
-
 
-/*custom rules over tridactyl default*/
+}
 
 :root.TridactylThemeGreenmat table {
```
cmcaine: These should all get `.TridactylOwnNamespace` too or they'll apply to webcontent (the rule above applies to all tables in all pages, for example).

src/static/clippy/tutor.template.html
```diff
@@ -1,4 +1,4 @@
-<html lang="en">
+<html lang="en" class="TridactylOwnNamespace">
```
cmcaine: Please also add `<!doctype html>` to disable quirks mode.

PR #601, opened 2018-05-23
[RFC] Containers

antonva:
lib/containers:
- enforces uniqueness on container creation and update as per comment in #499
- containerUpdate needs positional arguments / flag parsing imo. I'm willing to hear other ideas on how to implement it.

excmds:
- containercreate: could probably do with a success/failure return value.
- tabopen: takes a -c argument now, implementation is a bit messy, I'm open to suggestions here.
- containerclose: closes all tabs in specified container.
- ideas for other excmds?


src/excmds.ts
```diff
@@ -1309,6 +1328,7 @@ export async function tabopen(...addressarr: string[]) {
 
     activeTabContainerId().then(containerId => {
         if (containerId && config.get("tabopencontaineraware") === "true") openInNewTab(url, { active: active, cookieStoreId: containerId })
+        else if (container) openInNewTab(url, { active: active, cookieStoreId: "firefox-container-" + container })
```
bovine3dom: This should take priority over the line above it; `tabopen -c blah` should always open a tab in blah.

src/excmds.ts
```diff
@@ -1542,6 +1562,30 @@ export async function qall() {
...
 // }}}
 
+// {{{ CONTAINERS
+
+/** Closes all tabs open in the same container across all windows.
+ * @param containerId
+ *      The string represtation of the container id.
+ */
+//#background
+export async function containerclose(containerId: string) {
```
bovine3dom: What are these IDs? Are they just the container names? If they aren't, we should accept names as well as IDs and convert the names into IDs.

We should also have an excmd that lists all of your containers.

src/excmds.ts
antonva: Nice catch, I'll fix that real quick.

src/excmds.ts
antonva: The container ids are cookieStoreIds from firefox' contextual identities api.  They look like this: `firefox-container-1` and just increment that last digit by 1. 
I went with close by id digit instead of name because of how other addons handle containers. Most allow multiple containers with the same name, color and icon, the constraint placed in this is on (name, color, icon) meaning that even in this implementation you can create containers with the same name.

bovine3dom: I don't think I care about protecting people from themselves. I think

`tabopen -c blah` should open the container `blahrgbizness`, but throw an error if there is also `blahb` or indeed another `blahrgbizness`.

Related: how do containers get colours with the Firefox addon? is that something built in? It would be nice to add to the mode indicator, like we do for private mode now.

antonva: So, essentially a fuzzy match on the container name and error it in case of ambiguity? I'll see what I can do.

Re: colours
The colours are built in and easily accessible. I'll take a stab at adding it to the mode indicator.

bovine3dom: > So, essentially a fuzzy match on the container name and error it in case of ambiguity? I'll see what I can do.

Something like that. I'd probably check for exact matches, then fall back to `containername.includes(requestedname)`.

---

Colours - great :)

Don't sweat it too much if you can't but it should be pretty easy to copy the code that changes it for private.

bovine3dom:
@antonva is this ready to be merged, or do you want to colour the mode indicator?

antonva:
This is ready to be merged now. I'll have to take a look at colouring the mode indicator later.

src/lib/containers.ts
```diff
@@ -0,0 +1,154 @@
...
+/** Creates a container from the specified parameters. Unlike the official Mozilla addon, does not
+ *  allow multiple  containers with the same (name, color, icon) constraint.
+ *  @param containerName  The container name.
+ *  @param containerColor  The container color, limited to browser.contextualIdentities.IdentityColor.
+ *  @param containerIcon  The container icon, limited to browser.contextualIdentities.IdentityIcon.
+ */
+export async function containerCreate(
+    name: string,
+    color: string,
+    icon: string,
```
bovine3dom: Icon & colour should be optional. Colour should be picked at random; icon should probably just default to "fingerprint".

src/lib/containers.ts
```diff
@@ -0,0 +1,154 @@
+import * as Logging from "../logging"
+const logger = new Logging.Logger("excmd")
+
+/** Creates a container from the specified parameters. Unlike the official Mozilla addon, does not
+ *  allow multiple  containers with the same (name, color, icon) constraint.
+ *  @param containerName  The container name.
+ *  @param containerColor  The container color, limited to browser.contextualIdentities.IdentityColor.
+ *  @param containerIcon  The container icon, limited to browser.contextualIdentities.IdentityIcon.
```
bovine3dom: The help should list the valid colours / icons, or at least link to the MDN page.

src/lib/containers.ts
```diff
@@ -0,0 +1,154 @@
...
+    color: string,
+    icon: string,
+) {
+    let container = containerFromString(name, color, icon)
+    if (await containerExists(name)) {
+        logger.error("[containerCreate] Container already exists, aborting.")
+        logger.debug(container)
+    } else {
+        browser.contextualIdentities
+            .create(container)
```
bovine3dom: You should make sure that this throws an error if the container can't be created (e.g, if an invalid colour is asked for.)

bovine3dom:
@antonva could you try to tidy up this PR's history? It's quite hard to work out what you're trying to do.

I can't find my previous review comments but one thing that jumps out is that the help on the container commands needs a lot of improvement; e.g. there's no explanation of valid colours or icons for `containercreate`.

Cheers!

antonva:
@bovine3dom oh yeah this looks like a hot mess. Sorry about that!

I must admit I totally forgot to document the excmds properly, I'll get on that rn.

PR #606, opened 2018-05-24
Prettier compliance for documentation

sumnerevans:
I ran `prettier` on all of the `.md` files in the repository. Pretter appears to have made the following changes:

- Remove all hard-wraps
- Change list formatting
- Fix spacing around headers/paragraphs

This should make all of the docs consistently formatted.

Is there anywhere that I should mention that people should run `prettier` on their documentation before committing?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,355 +2,374 @@
 
 ## Release 1.13.0 / Unreleased
 
-- `fixamo` added: you can now use Tridactyl on addons.mozilla.org. Requires a `restart`.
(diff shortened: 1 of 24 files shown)


bovine3dom:
> Is there anywhere that I should mention that people should run prettier on their documentation before committing?

Yeah. Just add "*.md" to `cachedJS` in `scripts/common`. We should probably add "*.css" to it, too, and run that.

sumnerevans:
> We should probably add "*.css" to it, too, and run that.

I'll do that in another PR. There are a lot of changes that need to happen to the CSS files.

Issue #594, opened 2018-05-21
[Feature Request] style/delstyle/styletoggle/styleenable/styledisable commands


MoSal:
### In rc file:

``` .vim
style! -name=visited http* :visited { color:#f070a0 !important; }
```

### In command-line:

``` .vim
:styletoggle -name=visited
```

-----------

## Vimperator documentation for reference:

```
:sty[le][!] [-name={name}] [-append] {filter} [{css}]
        
            Add CSS styles to the browser or to web pages. {filter} is a comma separated
            list of URLs to match. URLs ending with * are matched as prefixes, URLs not
            containing any : or / characters are matched as domains. {css} is a full
            CSS rule set (e.g., body { color: blue; }).
                
            If {name} (short option: -n) is provided, any existing style with the same
            name is overridden, and the style may later be deleted using {name}. If
            -append (short option: -a) is provided along with -name, {css} and
            {filter} are appended to its current value.

           If {css} isn't provided, matching styles are listed.
    

:dels[tyle] [-name={name}] [-index={index}] [{filter}] [{css}]
        
            Delete any matching styles. If {filter} is provided, only matching elements of
            the filter are disabled. For instance, a filter mozilla.org, given a
            style for www.google.com,mozilla.org, will result in a style for
            www.google.com. The available options are:
        
            -name:  The name provided to :style (short option: -n)
            -index: For unnamed styles, the index listed by :style
                (short option: -i)
        

:styled[isable] [-name={name}] [-index={index}] {filter} {css}
        Enable any matching styles. Arguments are the same as for :delstyle.
    

:stylee[nable] [-name={name}] [-index={index}] {filter} {css}
        Disable any matching styles. Arguments are the same as for :delstyle.
    

:stylet[oggle] [-name={name}] [-index={index}] {filter} {css}
        Toggle any matching styles. Arguments are the same as for :delstyle.
```

bovine3dom:
This is probably pretty easy to implement, but Stylish is good enough for my simple needs of
`p {text-align: justify; -moz-hyphens: auto}` everywhere, so I'll leave this for someone else.

Happy to help anyone else who wants to have a crack at this. `browser.tabs.insertCSS` is a good place to start looking.

bovine3dom:
It turns out Stylish is a pretty major security hole and I would advise against anyone else using it: https://news.ycombinator.com/item?id=17447816

PR #609, opened 2018-05-24
Styletoggle

caputchinefrobles:
Implementing #594 

- [x] style, delstyle
- [x] "isable, "nable
- [x] "oggle
- [x] Docs
~~Cmdline listing of saved styletoggles (not sure about that)~~


caputchinefrobles:
This is the first commit set for sketch presenting the feature

src/excmds.ts
```diff
@@ -2171,6 +2171,100 @@ export function unset(...keys: string[]) {
...
+//#background
+export async function style(mode: string, ...args: string[]) {
+    let name: string
+    let css: string
+    let append: boolean = false
+    //let cssregex = /(\S*\s+)\{((?:.|\n)*?)\}\s*$/gm
+    if (args.length > 0) {
+        if (args.includes("-name")) {
+            name = args[args.indexOf("-name") + 1]
+        } else throw 'Missing "-name" argument.'
```
bovine3dom: Minor point here - I think we probably want `--name` style arguments rather than `-name`. (I know `winopen -private` doesn't follow that rule... that's my fault :) ).

caputchinefrobles: That's totally ok with me since it follows the way Linux environment is and which I'm used to. If you want I could change -private too but that probably has big impact on users and would have to be in another PR.

bovine3dom: > If you want I could change -private too but that probably has big impact on users and would have to be in another PR.

No - you're right, please don't do that yet. I'll probably hack in a special case later once we have the arg parser.

caputchinefrobles:
The listing style toggles probably worth waiting the new completions. If that's ok, at least for now, `get styletoggles` would fit the purpose.

caputchinefrobles:
Ready for review.

src/excmds.ts
```diff
@@ -2171,6 +2171,128 @@ export function unset(...keys: string[]) {
...
+ *  Here are some examples of usage:
+ *
+ * - Include a new css style matching all https pages:
+ *
+ *  `styleinclude --name stylemod --filter https://* --css body {border: 3px solid green;}`
+ *
+ * - Turn off the style included:
+ *
+ *  `styletoggle --name stylemod`
+ *
```
bovine3dom: Could you document how to view your current style settings? i.e, `viewconfig styletoggles` :)

src/content.ts
```diff
@@ -143,3 +144,26 @@ config.getAsync("modeindicator").then(mode => {
         if (config.get("modeindicator") !== "true") statusIndicator.remove()
     })
 })
+
+// loading style toggles on page load
+config.getAsync("styletoggles").then(styletoggles => {
```
bovine3dom: Is there a reason you can't use `DocStart` autocmds?

bovine3dom (changes requested):
Just a couple of things I've spotted after skimming through :)

src/excmds.ts
caputchinefrobles: Yes. For sure.

src/content.ts
caputchinefrobles: Yes.. first loading the stylemod on page has a different logic than `style -t` (toggle mode). To use autocmd on DocStart I would have to include a hidden load mode on excmd `style`, which I was not sure about as it was already quite long. 

Also, I'd have to deal only `string.search` on DocStart to match  the filter criteria instead of just working on `url.href`, `url.host` and `url.hostname`. 

I'd be happy to hear any suggestions. 🙂

caputchinefrobles:
Don't worry @bovine3dom, I'm not following that close because of a very busy time at work, but also I know this is not priority list. That's ok.

src/excmds.ts
```diff
@@ -2171,6 +2171,132 @@ export function unset(...keys: string[]) {
...
+ * @param mode
+ *         ALIASES          MODES
+ *      styleinclude   : style -i
+ *      styledelete    : style -x
+ *      styledisable   : style -d
+ *      styleenable    : style -e
+ *      styletoggle    : style -t
+ *
+ *  - `-i` includes new rule
```
bovine3dom: @caputchinefrobles: Could you try to explain this a little better? What does "include" mean? Where is it included?

Does it just mean "create"?

src/config.ts
```diff
@@ -177,6 +177,11 @@ const DEFAULTS = o({
         h: "help",
         authors: "credits",
         openwith: "hint -W",
+        styleinclude: "style -i",
+        styledelete: "style -x",
+        styledisable: "style -d",
+        styleenable: "style -e",
+        styletoggle: "style -t",
```
bovine3dom: Could we try to match the Pentadactyl (and I presume Vimperator) commands?

```
:sty[le] [-name={name}] [-group={name}] [-agent] {filter} {css}:sty[le] -name={name} -append {filter} {css}:sty[le] [-name={name}] [-group={name}] [filter]
        
            Add CSS styles to the browser or to web pages. {filter} is a
            comma-separated list of site-filters for which the style will
            apply. The ! character may not be used to invert the sense
            of the match. [css] is a full CSS rule set (e.g., body {
            color: blue; }).
        

        The following options are available:
        -appendIf provided along with -name, {css} and
                {filter} are appended to its current value. (short name
                -a)-agentIf provided, the style is installed as an Agent sheet, which
                applies to contents user interface widgets as well as normal
                elements. (short name -A)-group={group}The group to which to add this style. Please note that
                this grouping is for semantic and cleanup purposes only. No
                additional site filtering is applied.-name={name}If provided, any existing style with the same name is
                overridden, and the style may later be deleted using
                {name}. (short name -n)

        If [css] isn't provided, matching styles are listed.
    

:dels
:delstyle
:dels[tyle][!] [-name={name}] [-index={index}] [filter] [css]
        
            Delete any matching styles. With [!], delete all styles.
            If [filter] is provided, only matching elements of the
            filter are disabled. For instance, when run with a filter
            mozilla.org, an existing style with a filter
            www.google.com,mozilla.org, will result in a style for
            www.google.com.
        

        The available options are:

        -nameThe name provided to :style (short name -n)-indexFor unnamed styles, the index listed by :style
                (short name -i)
    

:styleenable
:stylee
:styenable
:stye
:stylee[nable][!] [-name={name}] [-index={index}] [filter] [css]
        Enable any matching styles. With [!], enable all styles. Arguments are the same as for :delstyle
    

:styledisable
:styled
:stydisable
:styd
:styled[isable][!] [-name={name}] [-index={index}] [filter] [css]
        Disable any matching styles. With [!], disable all styles. Arguments are the same as for :delstyle
    

:styletoggle
:stylet
:stytoggle
:styt:
stylet[oggle][!] [-name={name}] [-index={index}] [filter] [css]
        Toggle any matching styles. With [!], toggle all styles. Arguments are the same as for :delstyle
```

Thanks.

bovine3dom (changes requested):
I found the commands quite hard to understand from a user's perspective.

I haven't looked at how the code actually works, so please don't take this as approval or disapproval of that : ).

src/config.ts
caputchinefrobles: I'll try to do initial css load on autocmd as well...

src/content.ts
bovine3dom: I don't see the problem with DocStart:

String.search() allows you to use a regex, so you can easily match multiple URLs with pipes, e.g. `autocmd DocStart /www.(reddit|google).com/`. Do you just think this is less user friendly?

> to use autocmd on DocStart I would have to include a hidden load mode on excmd style

I think `styleapply [name]` is a totally reasonable thing to have. One could use it to roll one's own reader mode, for example.

src/url_util.ts
```diff
@@ -370,3 +370,26 @@ export function interpolateSearchItem(urlPattern: URL, query: string): URL {
...
+/**
+ * Check if url URL filter criteria matches the current page URL
+ *
+ * @param part      a URL part string or full URL be matched with the @param url
+ * @param url       a URL object of the current page
+ *
+ * @return          `true` if url `part` matches the full URL href, and `false` otherwise
+ *
+ */
+export function partMatchesURL(part: string, url: URL): boolean {
```
bovine3dom: I don't understand the need for this over `String.prototype.search`.

Did Vimperator have something similar?

src/config.ts
bovine3dom: Please do make these match the Vimperator commands : )

src/styling.ts
```diff
@@ -61,3 +62,38 @@ browser.storage.onChanged.addListener((changes, areaname) => {
         retheme()
     }
 })
+
+export async function insertstyle(csscode: string) {
+    let tab = await webext.activeTab()
+    try {
+        //insert css to the current active tab
+        await webext.browserBg.tabs.insertCSS({ code: csscode })
```
bovine3dom: You don't need any of this code here. Please see my comments in excmds.ts : )

src/excmds.ts
```diff
@@ -2171,6 +2171,132 @@ export function unset(...keys: string[]) {
...
+    }
+
+    // applying config changes
+    if (mode !== "-x") {
+        if (styletoggles[name].enabled && styletoggles[name].toggled) {
+            if (styletoggles[name].filters != undefined) {
+                let match = styletoggles[name].filters.some(f => UrlUtil.partMatchesURL(f, current_url))
+                if (match)
+                    styletoggles[name].css.forEach(css => {
+                        messageActiveTab("styling_content", "insertstyle", [css])
```
bovine3dom: `browser.tabs.insertCSS({ code: css })` would work fine here.

src/excmds.ts
```diff
@@ -2171,6 +2171,132 @@ export function unset(...keys: string[]) {
...
+                    })
+            } else
+                styletoggles[name].css.forEach(css => {
+                    messageActiveTab("styling_content", "insertstyle", [css])
+                })
+        } else
+            styletoggles[name].css.forEach(css => {
+                messageActiveTab("styling_content", "removestyle", [css])
```
bovine3dom: `browser.tabs.removeCSS({ code: css })` : )

src/excmds.ts
bovine3dom: Please do improve this documentation : )

src/content.ts
bovine3dom: Using `DocStart` instead of this would also save you from having to use messages explicitly.

bovine3dom (changes requested):
I finally got around to doing a proper review...

This code is mostly fine, so thanks!

There are just a few places where I feel like we could do things a little better.

If you don't feel up to doing all or any of it yourself, please just say and I'll get around to doing it myself sometime this millennium.

src/content.ts
caputchinefrobles: Thanks for the review. I'm currently busy with some studio recording stuff but it will not take much longer.

src/styling.ts
caputchinefrobles: ok.. that's something I should make a point. I don't know if this is important or not, but wouldn't it be more semantically concise if we call insertCSS from styling.ts than spreading it all over the code (in excmds or any other file)?

bovine3dom: No, I think insertCSS is simple enough that we can have it inline in excmds : )

src/config.ts
bovine3dom: I don't see them, but that would be fine.

MoSal:
Any updates on this?  
Is there a way to help moving this along?

bovine3dom:
Waiting on @caputchinefrobles. If he is willing to let you have a go, all my comments still apply I think : )

caputchinefrobles:
@MoSal, please feel free to go on this. I'm not finding the proper time to contribute. I'm really sorry about that and would be of much appreciation if you could help moving this along.

PR #611, opened 2018-05-25
[WIP]: Completions

IsaacKhor:
Completely rewrite the completion engine to have real support for options and documentation and the parser with a formal grammar.

Goals:
1. Implement support for gnu-style `-f --foo` arguments
2. Let devs make a formal spec of each command and the options/arguments that are accepted
3. Completions generation based on the spec mentioned in 2

Why?
- We need a real parser
- Our current completions system is ... a thing that exists and is in danger of falling apart at the slightest touch
- Instead of one command with several options, we have multiple command that do their own options parsing all over the place and do it poorly

What would this solve?
- Command discovery
- Speed of command input (via tab completions)
- Error reporting (when the parser detects an error)

### DIFF
--- a/src/command_background.ts
+++ b/src/command_background.ts
@@ -1,15 +1,16 @@
 import { List, Map } from "immutable"
-import { MayList, toList, HistItem, RArray } from "./util"
+import { BrowserTab, MayList, toList, HistItem, RArray } from "./util"
 import minimist from "./minimist"
 import { get as levenshtein } from "fast-levenshtein"
 import * as I from "immutable"
 import * as Util from "./util"
 import * as Alias from "./aliases"
 import * as Config from "./config"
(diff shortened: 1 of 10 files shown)


src/command_background.ts
```diff
@@ -0,0 +1,323 @@
+import { List, Map } from "immutable"
+import { MayList, toList, HistItem, RArray } from "./util"
+import minimist from "./minimist"
```
bovine3dom: I can't remember why, but in #523 we found that minimist and yargs didn't do quite what we wanted, hence why we went with command-line-args, which still doesn't quite do what we wanted, but seemed less insurmountably so at the time.

bovine3dom:
What's left to do on this? Are you still intending to work on it, @IsaacKhor?

PR #620, opened 2018-05-28
Add build with Windows support on AppVeyor

gsbabil:
Dear Maintainers,

This PR will add support for building Tridactyl with MSYS2/MinGW on Windows on the AppVeyor platform.

Thank you.

PR #625, opened 2018-05-29
Add additional fixes, clean-ups etc. to AppVeyor Windows build support

gsbabil:
Dear Maintainers,

This PR add some additional fixes and clean-ups on top of #620. Please have a look.

Thank you.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -86,8 +86,8 @@ build_script:
   - python --version
 
   # Install Python modules under Python 3.6
-  - ps: python -m pip install --upgrade pip
-  - ps: python -m pip install --upgrade pyinstaller
+  - ps: python -m pip install --upgrade pip>=10.0.1
+  - ps: python -m pip install --upgrade pyinstaller>=3.3.1
 
   # Change to build directory and verify CWD
   - ps: Set-Location -Path $env:APPVEYOR_BUILD_FOLDER
@@ -115,8 +115,8 @@ test_script:
   - python --version
 
   # Install Python modules under Python 3.6
-  - ps: python -m pip install --upgrade pip
-  - ps: python -m pip install --upgrade pyinstaller
+  - ps: python -m pip install --upgrade pip>=10.0.1
+  - ps: python -m pip install --upgrade pyinstaller>=3.3.1
 
   # Change to build directory and verify CWD
   - ps: Set-Location -Path $env:APPVEYOR_BUILD_FOLDER
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -47,10 +47,19 @@ install:
   #
   # Credits: JuliaLang developers.
   #
-  - ps: if ($env:APPVEYOR_PULL_REQUEST_NUMBER `
-             -and $env:APPVEYOR_BUILD_NUMBER `
-             -ne ((Invoke-RestMethod `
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -59,28 +68,30 @@ install:
...
   - ps: $env:PATH = "C:\Python36-x64;$env:PATH"
-  - ps: Copy-Item -Path "C:\Python36-x64\Python.exe" -Destination "C:\Python36-x64\Python3.exe"
+  - ps: >-
+        Copy-Item `
+        -Path "C:\Python36-x64\Python.exe" `
+        -Destination "C:\Python36-x64\Python3.exe"
+  - python --version
+
+  # Install Python modules under Python 3.6
+  - ps: python -m pip install --upgrade pip
```
bovine3dom: Python 3.4+ should include pip by default - do you have reason to believe that it doesn't?

.appveyor.yml
```diff
@@ -59,28 +68,30 @@ install:
...
+  - ps: python -m pip install --upgrade pyinstaller
```
bovine3dom: As I mentioned in #620, this won't be used for anything unless we add pyinstaller to build.sh.

bovine3dom: (these two comments apply to npm test too, obviously :) );

.appveyor.yml
gsbabil: > do you have reason to believe that it doesn't?

No. 

I was doing the `--upgrade` mostly to get the latest `10.0.1` release. That's the version I have locally installed. 

We should probably "pin" the version numbers in here to be more clean and pedantic. Let me add them in. In case something breaks in the future, it'll be easier to spot.

Thanks.

.appveyor.yml
gsbabil: Well, I don't think it'll hurt anything. But I'm okay to comment them out for now if you want. Let me know.

BTW, did you want to have the `build.sh` from #560 here under #625? As said below, I think building EXE on Windows would be the right approach long term.

Thanks.

bovine3dom: No, as a separate PR, please :)

bovine3dom:
> So, I think it would be good to start building the EXE on Windows when we are ready, and move away from Wine at some point.

We'd have to work out how to get the executable out of AppVeyor and accessible via a "static" URL. A problem for another day, I think :)

gsbabil:
> A problem for another day, I think :)

I had actually looked into it. The simpler I think is to set-up a FTP server and let AppVeyor upload the file to us. This could go into the same machine where the current beta builds are getting stashed.

If this isn't possible or desired, AppVeyor alternatively supports web-hooks using which it can "signal" one our machines/bots to fetch the freshly built EXE.

So yes, I think you need to decide how you'd like to handle the delivery/deployment steps, on which server _etc._ before we can tackle this part.

Thanks for merging!

_P.S._

Let me know how you'd like to receive the omitted `build.sh` from the [gsbabil/windows-compiled-native-messenger-support][0] branch. Did you leave out any other files from the #560 PR?

I don't need it to get merged immediately. But I'd like to send the PR before I moved on to other things.

Thanks.

[0]: https://github.com/gsbabil/tridactyl/tree/gsbabil/windows-compiled-native-messenger-support

PR #683, opened 2018-06-14
Containers 2: documented boogaloo

antonva:
**New excmds:**
- containerclose: closes all tabs using specified container
- containercreate: creates a new container
- containerupdate: updates an existing container
- containerremove: removes an existing container (calls containerclose before removing container)

**Edits to tabopen:**
- Now takes the `-c` flag, allowing you to open a new tab in a specified container.
- Both `-b` and `-c` can be supplied in any order via argparse, in lieu of a proper flag parsing mechanism.

**Requested changes from old PR:**
- [x] `tabopen -c` takes precedence
- [x] improve excmd documentation wrt new container cmds
- [x] improve error handling 

**Closing thoughts:**
- I'm not very happy about how the `containerupdate` function works but I can't see any way around that as long as there's no flag parsing available.
- I should probably add more debug points in `src/lib/containers.ts` but I'm not sure what the convention is/should be.
- `open -c` is not supported in this PR since it's not possible to "move" a tab into a container. The container must be specified on tab creation meaning that the only way to support `open -c` would be to close and reopen the current tab.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -88,7 +88,7 @@
 // Shared
 import * as Messaging from "./messaging"
 import { l, browserBg, activeTabId, activeTabContainerId } from "./lib/webext"
-import { containerCreate, containerExists, containerFuzzyMatch } from "./lib/containers"
+import * as Container from "./lib/containers"
 import state from "./state"
 import * as UrlUtil from "./url_util"
 import * as config from "./config"
@@ -1488,7 +1488,7 @@ export async function tabopen(...addressarr: string[]) {
         } else if (args[0] === "-c") {
             // Ignore the -c flag if incognito as containers are disabled.
             let win = await browser.windows.getCurrent()
-            if (!win["incognito"]) container = await containerFuzzyMatch(args[1])
+            if (!win["incognito"]) container = await Container.fuzzyMatch(args[1])
             else logger.error("[tabopen] can't open a container in a private browsing window.")
 
             args.shift()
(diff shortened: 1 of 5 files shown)


src/excmds.ts
```diff
@@ -1456,7 +1457,9 @@ export async function tablast() {
 
 /** Like [[open]], but in a new tab. If no address is given, it will open the newtab page, which can be set with `set newtab [url]`
 
-    Use the `-b` flag as the first argument to open the tab in the background.
+    Use the `-c` flag followed by a ccontainer id to open a tab in said container.
```
bovine3dom: 1. Container is spelt with a single c :)
2. Please mention that the match is fuzzy.

antonva:
Container colour on the mode indicator in the latest commit. I'm having a hard time noticing the colours on the 1.5px border in dark mode though.
As a side node, should `src/lib/containers.ts` exist or should it be a part of `src/lib/webext.ts` or some other file?

bovine3dom:
@antonva if you feel like it, a `containerlist` or `viewcontainers` command would be nice if you're feeling up to it. I'd suggest looking at how `viewconfig` is done.

Merged as what is here is good. Thanks!

bovine3dom:
>open -c is not supported in this PR since it's not possible to "move" a tab into a container. The container must be specified on tab creation meaning that the only way to support open -c would be to close and reopen the current tab.

Is there a reason why there isn't a `winopen -c`?

antonva:
@bovine3dom it's simply because that `browser.windows.create doesn't take a cookieStoreId, there was no simple way for me to do it. 
I envisioned I could open a tab in the background, grab the tabId of that and move it to the new window via the `createData` object but I wanted to have a discussion on a feature that would behave like that before implementing.

bovine3dom:
It's probably something so uncontroversial that we should probably ask Mozilla for it on the BMO :)

antonva:
Sounds good to me. Is that something you have xp with?

I'll take a stab at creating `viewcontainers`. I'll create an issue for it.

bovine3dom:
>  Is that something you have xp with?

Nope. Bugzilla is a scary place.

PR #685, opened 2018-06-14
predictable numeric hints

peff:
One of the things I missed moving from pentadactyl to tridactyl is the way the hints are generated. I really like them to be predictable between visits to a page, since I tend to develop muscle memory (e.g., clicking through to the article of a Hacker News page is always "f13"). Switching `hintchars` to numeric isn't quite enough to get there, because the hint-generator tries to avoid prefixes. Which means that the "early" hints are impacted by the exact number of hints later in the page.

This PR introduces a new mode that provides sequential numeric hints. It's triggered by a new config variable rather than `hintfiltermode` because it's really orthogonal to the filtering itself (though I recommend using it with `vimperator-reflow`). I've been using it for a few months and have been pretty happy.

I had to do a little bit of refactoring along the way; see the individual commit messages for details.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -232,6 +232,7 @@ const DEFAULTS = o({
     homepages: [],
     hintchars: "hjklasdfgyuiopqwertnmzxcvb",
     hintfiltermode: "simple", // "simple", "vimperator", "vimperator-reflow"
+    hintnames: "short",
 
     // Controls whether the page can focus elements for you via js
     // Remember to also change browser.autofocus (autofocusing elements via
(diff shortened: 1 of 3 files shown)


peff:
> Any reason this couldn't support letters instead of numbers? i.e. still use hintchars?

I actually tried that at first, but it was awkward to create numerically sequential hints just by pasting the characters together. (E.g., `9` is followed by `0`, then `11`). That's not the end of the world, but I found it vaguely displeasing.

> Really, if you want reproducible interaction via the keyboard for a given website, a deterministic algorithm based on count isn't enough. 

You're right, but it opens up such a large number of questions that I'd rather take the "good enough" approach. In addition to the questions you raised, there's the issue of maintaining the hintdb. When I move to a new machine, I have to copy it or risk losing all of my memorized hints. Do I start version-controlling it with my other dotfiles? :grimacing:

peff:
> E.g., 9 is followed by 0, then 11

That's with `1234567890`, of course. You can do `0123456789`, but then you get `01`, etc.

bovine3dom:
Thanks!

Issue #447, opened 2018-04-27
"Your new tab has changed!" steals focus every ****ing time Tridactyl updates


bovine3dom:
We probably need to ask nicely on bugzilla for this not to pop up every time the extension is updated.

bovine3dom:
We should probably just hide this with userChrome.

bovine3dom:
Closing this as there is an upstream bug about it (see #760).

PR #699, opened 2018-06-18
guiset: added none option to navbar

Snaipe:
This guiset navbar option completely removes the bar (except during
customization), rather than auto-hiding it when not in focus.

Since add-on panels are appended as children of #nav-bar, care has been
taken not to hide them to keep functionality.

Signed-off-by: Franklin "Snaipe" Mathieu <me@snai.pe>

### DIFF
--- a/src/css_util.ts
+++ b/src/css_util.ts
@@ -79,6 +79,22 @@ export const potentialRules = {
             show: ``,
         },
     },
+    // All children except add-on panels
+    navbarnonaddonchildren: {
+        name: `:root:not([customizing]) #nav-bar > :not(#customizationui-widget-panel)`,
+        options: {
+            hide: `display: none !important;`,
+            show: ``,
+        },
+    },
+    // Set navbar height to 0
+    navbarnoheight: {
+        name: `:root:not([customizing]) #nav-bar`,
+        options: {
+            hide: ``,
+            show: `max-height: 0; min-height: 0 !important;`,
+        },
+    },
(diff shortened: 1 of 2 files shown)


bovine3dom:
This looks nice, thanks! I've got a bit of a backlog but this is reasonably simple so I'll probably review it properly tomorrow.

---

Have you thought about how you might warn people that one of their add-ons needs new permissions before it is allowed to update? Currently, Mozilla do this with a small muted yellow exclamation mark in the hamburger menu in the top right of the navbar. If there's no navbar, the only way to find out is to go to about:addons and check yourself.

Additionally, since you were playing with the browser CSS, would you mind looking into #447? It could be as simple as sticking `display: none` on the right element.

Snaipe:
> Have you thought about how you might warn people that one of their add-ons needs new permissions before it is allowed to update? Currently, Mozilla do this with a small muted yellow exclamation mark in the hamburger menu in the top right of the navbar. If there's no navbar, the only way to find out is to go to about:addons and check yourself.

Right; I think this is somewhat acceptable behaviour when specifying `none` explicitly. In my pentadactyl days, I didn't mind compromising from the lack of navbar, so I wouldn't mind checking from time to time the addons page; but to be fair, that's my own opinion.

> Additionally, since you were playing with the browser CSS, would you mind looking into #447? It could be as simple as sticking display: none on the right element.

Sure. I haven't payed too much attention though, so I can't recall seeing that dialogue, but I can try reproducing and nuking the element.

PR #704, opened 2018-06-19
Container commands exists and fuzzyMatch are now case insensitive.

antonva:


bovine3dom:
Please document this somewhere :)

PR #729, opened 2018-06-23
Improve robustness of Get-PythonVersionStatus

vlamacko:
The current way to get Python version is volatile, e.g. Anaconda bundled Python 3 won't get recognised as such.

As of now the `--version` flag is used. This results in inconsistent behaviour, e.g. reference Python will print the version to stderr but Anaconda Python will print to stdout. Thus the version is found using the sys module. The output is more standardised.

Regex matching is now employed.

### DIFF
--- a/native/win_install.ps1
+++ b/native/win_install.ps1
@@ -13,7 +13,7 @@ $global:MessengerBinPyName = "native_main.py"
 $global:MessengerBinExeName = "native_main.exe"
 $global:MessengerBinWrapperFilename = "native_main.bat"
 $global:MessengerManifestFilename = "tridactyl.json"
-$global:PythonVersionStr = "Python 3"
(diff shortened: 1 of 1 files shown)


bovine3dom:
This looks good, thank you!
In Linux we just check to see if python3 is on the path: https://github.com/cmcaine/tridactyl/blob/db780f2d692c4212fce24f4dc83ef67b181b9cd5/native/install.sh#L54

Is that also fragile?

vlamacko:
I don't think so, at least not to this particular issue. 

Now I am wondering whether it is even necessary to perform this check. We could just let the python script execution fail and handle afterwards.

I guess checking the version/installation like this (ahead of time) ensures more deterministic results but with possible tradeoff of false positives.

All considered I think this pull request is still fine as it reduces the occurrence of such false positives.

PR #734, opened 2018-06-25
Fix bug introduced in #733

vlamacko:
Fix bug introduced in #733. Also use `.\` for explicit relative path.


vlamacko:
To be clear the bug is always using PyName for generic BinName which can be either PyName or ExeName.

PR #750, opened 2018-07-03
add autocmd! to remove specific autocmds

friedm:
I had some trouble with an autocmd and couldn't find an easy way to remove it, since that state is persistent. Adds `rmautocmd <event> <url>` to clear that state.  Aliased to `autocmd!` and `au!` to attempt to be more vim-like.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -161,6 +161,8 @@ const DEFAULTS = o({
     exaliases: o({
         alias: "command",
         au: "autocmd",
+        "au!": "rmautocmd",
+        "autocmd!": "rmautocmd",
         b: "buffer",
         o: "open",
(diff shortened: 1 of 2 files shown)


bovine3dom:
Thanks so much for this, @friedm. I'm sorry it took so long - this PR crept past me. Next time I'll be quicker : )

I've renamed `rmautocmd` to `autocmddelete` because that's more like Vim - commands are pseudo-namespaced with the convention `object|action`.

I also removed your `au!` alias as the exclamation usually means overwrite in Vim, so I'd expect `au!` to behave exactly like `au` does but overwrite already existing auto commands (which we do anyway because I'm lazy).

PR #754, opened 2018-07-03
Auto Containers

antonva:
* Add `reopenincontainer` excmd.
* Add documentation example of container aucmd in DocStart section of `src/config.ts`

Current implementation uses the DocStart hook to reopen the tab. It's not the smoothest transition and not nearly as smooth as creating an eventListener for webRequests like the mozilla/contain-facebook extension does but that's more effort.


bovine3dom:
> but that's more effort.

#tridactyllife 

What's the problem with the webRequests approach? Can we not add an autocmd event for them?

antonva:
We can, I just haven't gotten around to it and this was a first naive approach I had a few weeks back. Thought I'd polish it a bit and see if it would do the job.

Also, as the webRequests approach is dealing with stuff before a tab is created, it'd be a lot of code that couldn't be a simple excmd like the current approach.

src/excmds.ts
```diff
@@ -1869,6 +1869,38 @@ export async function viewcontainers() {
             .replace(/#/g, "%23")
             .replace(/ /g, "%20")
 }
+
+/** Reopens current tab in specified container.
```
bovine3dom: Please mention that the tab is not duplicated at the history is lost. (That is what happens, right?).

antonva: Will do, that is what happens.

bovine3dom: It might be worth navigating back in the tab before closing it, otherwise `u` will get you trapped in a loop.

antonva:
Introduces the autocontain directives: `autocontain youtube.com google` contains youtube.com and all subdomains in the `google` container.

The `auconcreatecontainer` option allows for auto creation of containers if they're specified in tridactylrc but don't exist in the profile.
`auconpreservetab` is something I think should be default behaviour but I'm not quite sure if people want to be able to turn it off.

I haven't fully handled/cleaned up the webRequests but it works on the most basic level.

bovine3dom:
> Should we make the code interoperable with Multi Account Containers and/or Contain Facebook?

Yes, please.

> Currently, once a tab is opened in a certain container with autocontain, all links followed will be in that container as well, which is opposite of what plugins like contain-facebook do, where only the domain specified is contained and everything else is thrown back to the firefox-default container. Not quite sure which direction to take here. Would appreciate some input if anyone's interested.

I think `autocontain` should work like Contain Facebook but Contain [insert domain here] - so yes, please leave the container. It's probably sensible to look at their code (and please give credit in the code / commit messages).

src/lib/autocontainers.ts
```diff
@@ -0,0 +1,175 @@
...
+            delete this.cancelledRequests[tabId]
+        }
+    }
+
+    // Parses autocontain directives and returns valid cookieStoreIds or errors.
+    parseAucons = async (details): Promise<string> => {
+        let aucons = Config.get("autocontain")
+        const ausites = Object.keys(aucons)
+        const aukeyarr = ausites.filter(
+            e => details.url.search("^https?://.*" + e + "/") >= 0,
```
bovine3dom: I think the `autocontain` documentation is slightly misleading. It isn't clear that your "URL" is actually parsed as a regex.

E.g, `autocontain www.google.co.uk Work` would also match on `www#google#co.uk`and maybe `"https://www.duckduckgo.com/q=?www.google.co.uk/".search("^https?://.*" + "www.google.co.uk" + "/")`

It's also a useful property as it means you can do e.g. `autocontain www.google.co(.uk|com)`. (Forgive the pidgin regex).

src/excmds.ts
```diff
@@ -1821,7 +1821,7 @@ export async function containercreate(name: string, color?: string, icon?: strin
   @param name The container name.
  */
 //#background
-export async function containerremove(name: string) {
+export async function containerdelete(name: string) {
```
bovine3dom: Secret breaking changes are double-plus-ungood :(

Please leave an alias for the old command.

src/excmds.ts
```diff
@@ -2352,6 +2353,21 @@ export function autocmd(event: string, url: string, ...excmd: string[]) {
     config.set("autocmds", event, url, excmd.join(" "))
 }
 
+/** Automatically open a domain and all its subdomains in a specified container.
+  
+ @param domain The domain which will trigger the autoContain directive. Includes all subdomains.
```
bovine3dom: As I mention in the other file, please be more explicit about how the domain is matched.

src/lib/autocontainers.ts
antonva: I agree, I tried to make it clear by waffling on about domains. I'll write something to absolutely make it clear.

The regex bit is indeed useful and should probably also be documented better. 

I had already encountered similar situations as: `"https://www.duckduckgo.com/q=?www.google.co.uk/".search("^https?://.*" + "www.google.co.uk" + "/")` and had made some changes to the matching. I guess I could take another stab at writing a better regex there.

src/excmds.ts
antonva: Duly noted. (mainly changed it because of the typo in the changelog though ;))

PR #763, opened 2018-07-05
Make guiset hoverlink compatible with firefox 61

bodograumann:
In firefox 61 there has been a medium change to how the status panel
(that shows the link the user hovers over) is created. Compare firefox
changeset 18bbbdc02213:
https://hg.mozilla.org/releases/mozilla-release/rev/18bbbdc02213

In particular there is no more statuspanel tag. Instead there is an hbox
with id="statuspanel".


bovine3dom:
Thanks!

Issue #657, opened 2018-06-07
Failed installnative on Windows does not leave power shell window open


bovine3dom:
Related to #656 - user said in Riot that

> ok, Im just trying to figure out a way to make powershell not close after running the script so i can read the output

Permalink https://matrix.to/#/!xmubtggllUtLrQiGyn:matrix.org/%2415283898804338516FQCyr:matrix.org

budomino:
OK So after trying to figure out why I also had this problem, I realized the real problem is that there's no error message to let you know you need Python3 installed. After I installed python3 via Chocolatey the process completed without a hitch.

bovine3dom:
There is an error message, you just can't see it because the window closes :(

dokpm0:
I tried installing Python3.  It didn't help me.  If I issue $host.enternestedprompt() before trying to install native it leaves the PowerShell window open so I can see the error.  With or without python I always get a download failed, quitting error.  Yet I can manually download what PowerShell fails to download.  Is there a way to install native without using PowerShell?

bovine3dom:
You have to set some registry keys, so it's a bit of a faff.

I suggest downloading the files that the PowerShell script uses and then editing the script to use your local files, and then running that.

Could you copy and paste the errors here?

bovine3dom:
@gsbabil regarding this issue (PowerShell not remaining open), do you have any ideas?

dokpm0:
This is what I'm getting:

Select-Object : Property "Source" cannot be found.
At line:499 char:28
+             | Select-Object <<<<  -ExpandProperty "Source"
    + CategoryInfo          : InvalidArgument: (py.exe:PSObject) [Select-Object], PSArgumentException
    + FullyQualifiedErrorId : ExpandPropertyNotFound,Microsoft.PowerShell.Commands.SelectObjectCommand

    - Python 3 found at:
[+] Preparing C:\Users\dokpm0\.tridactyl ...
    - C:\Users\dokpm0\.tridactyl exists, moving on ...
    - Testing C:\Users\dokpm0\.tridactyl was successful!
    - Creating C:\Users\dokpm0\.tridactyl was successful!
[+] Downloading https://raw.githubusercontent.com/cmcaine/tridactyl/master/native/native_main.py?1531167545.20697 ...
Invoke-WebRequest StatusCode:
Invoke-WebRequest StatusDescription:
[+] Downloading failed, quitting ...

[...]

PR #786, opened 2018-07-11
Fix native_main.exe download failure, PowerShell auto-close etc.

gsbabil:
Dear Maintainers,

This set of patches will fix the following issues. These are
currently being discussed at #657.

1. Fix `native_main.exe` download due of TLS version mismatch
   between `Invoke-WebRequest` and `tridactyl.cmcaine.co.uk`

2. Prevent PowerShell window from auto-closing if native-messenger
   download or installation fails

3. Allow the output of `:installnative` to be executed on any of the
   following shells on Windows:

    - MinTTY/MSYS2
    - Windows Command Prompt (`cmd.exe`)
    - Windows PowerShell (`powershell.exe`)
    - Cygwin

Hope this helps. Thanks.


bovine3dom:
Ah, point 1 explains a lot! I was about to file an issue asking if native_main.exe was ever used as according to our logs, it was only downloaded twice in the last week...

PR #839, opened 2018-07-19
Ctrl-Y should scroll up

chrispickard:
`<c-y>` in vim scrolls up, so it should in tridactyl as well

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -60,7 +60,7 @@ const DEFAULTS = o({
         j: "scrollline 10",
         "<C-e>": "scrollline 10",
         k: "scrollline -10",
-        "<C-y>": "scrollline 10",
+        "<C-y>": "scrollline -10",
         h: "scrollpx -50",
         l: "scrollpx 50",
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks for this.

Does the opposite (scroll down) bind get stolen by Firefox? I'd rather bind both at once, or neither.

bovine3dom:
FWIW, your git config and GitHub settings disagree. You could add some email addresses to both until they agree (otherwise you won't be counted as a contributor properly).

chrispickard:
`<C-e>` scrolls down for me on linux firefox 61, does it work for you?

Thanks for the tip about my committer email

chrispickard:
`<C-e>` scrolls down for me on linux firefox 61, does it work for you?

Thanks for the tip about my committer email

bovine3dom:
Ah, you're right.

I didn't realise we'd already tried to implement this feature - it's just a typo you're fixing.

Cheers!

PR #862, opened 2018-07-28
Default bookmark name when bookmark function is not provided a title

reversebreak:
RE #711 

Fixed excmds.ts to take the current tab's title if the bookmark function isn't provided with a title.

Still a couple of improvements that could be made (as mentioned in the code comments).

Policy decision needs to be made about what we want to happen when someone uses the colon to call the bookmark maker function and bookmarks a URL that's not the current one and doesn't give a title.  Right now with these changes it will take the title of the current page.  
I wasn't comfortable making that call myself, so any input is appreciated.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4066,6 +4066,7 @@ export async function perfhistogram(...filters: string[]) {
  */
 //#background
 export async function bmark(url?: string, ...titlearr: string[]) {
+    const auto_url = (url == undefined)
     url = url === undefined ? (await activeTab()).url : url
     let title = titlearr.join(" ")
     // if titlearr is given and we have duplicates, we probably want to give an error here.
@@ -4076,7 +4077,7 @@ export async function bmark(url?: string, ...titlearr: string[]) {
     // if title is blank, get it from the current page.
     // technically could race condition if someone switched tabs REALLY quick after
     // bookmarking, but too unlikely to bother with for now
-    if (title == "") {
+    if (title == "" && auto_url) {
         //retrieve title from current tab
         title = (await activeTab()).title
     }
@@ -4097,7 +4098,7 @@ export async function bmark(url?: string, ...titlearr: string[]) {
         if (pathobj === undefined) pathobj = validpaths.find(p => p.path.includes(path))
         //technically an initial title string like `Firefox/` can give us a blank title
         //once we remove the path, so let's fix that
-        if (title == "") {
+        if (title == "" && auto_url) {
             //retrieve title from current tab
             const currTitle = (await activeTab()).title
             title = currTitle
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4073,8 +4073,15 @@ export async function bmark(url?: string, ...titlearr: string[]) {
     dupbmarks.forEach(bookmark => browser.bookmarks.remove(bookmark.id))
     if (dupbmarks.length !== 0) return
     const path = title.substring(0, title.lastIndexOf("/") + 1)
-    // TODO: if title is blank, get it from the page.
-    if (path !== "") {
+    // if title is blank, get it from the current page.
+    // technically could race condition if someone switched tabs REALLY quick after
+    // bookmarking, but too unlikely to bother with for now
+    if (title == "") {
+        //retrieve title from current tab
+        title = (await activeTab()).title
+    }
+
+    if (path != "") {
         const tree = (await browser.bookmarks.getTree())[0] // Why would getTree return a tree? Obviously it returns an array of unit length.
         // I hate recursion.
         const treeClimber = (tree, treestr) => {
@@ -4088,6 +4095,14 @@ export async function bmark(url?: string, ...titlearr: string[]) {
         let pathobj = validpaths.find(p => p.path === path)
         // If strict look doesn't find it, be a bit gentler
         if (pathobj === undefined) pathobj = validpaths.find(p => p.path.includes(path))
+        //technically an initial title string like `Firefox/` can give us a blank title
+        //once we remove the path, so let's fix that
+        if (title == "") {
+            //retrieve title from current tab
+            const currTitle = (await activeTab()).title
+            title = currTitle
+        }
+
         if (pathobj !== undefined) {
             browser.bookmarks.create({ url, title, parentId: pathobj.id })
             return


src/excmds.ts
```diff
@@ -3072,7 +3072,38 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
+    var titleSourceTab = browser.tabs.query({active: true, currentWindow: true});
+    //TODO: define and implement sane behaviour for if the title is blank, but the URL
+    //      that is being bookmarked is not actually the current tab. just using 
+    //      await activeTab().title doesn't make sense for that, we'll need to change the tab query
+    //      immediately above to search for the url.  I guess then just go with arr[0] for the winner
+    //      The possibility also exists that someone tries to bmark a URL thats not a current tab anywhere
+    //      so sane behaviour needs to be defined for that too.  Load the page headless and extract URL?
+    //      Make the title the URL?
+    titleSourceTab.then(getCurrTabTitle);
+    await titleSourceTab;
```
bovine3dom: This is really hard to read - the `currTitle` side-effect is quite buried.

A clearer way of doing all of this would simply be:

```js
let currTitle = (await activeTab()).title
```

or, pretending you don't know about our own `activeTab`

```js
let currTitle = (
    await browser.tabs.query(
        {active: true, currentWindow: true}
    )
)[0].title
```

: )

src/excmds.ts
```diff
@@ -3072,7 +3072,38 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
+
+    // if title is blank, get it from the page.
+    if (title == "") {
+        title = currTitle
```
bovine3dom: Please only execute the title finding code if `title` is blank.

src/excmds.ts
```diff
@@ -3072,7 +3072,38 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
+        //requires [tabs] permission which tridactyl should have grabbed anyway
+        if (tabs[0]) {
+            var currentTab = tabs[0];
+            currTitle = currentTab.title
+        }
+    }
+    var titleSourceTab = browser.tabs.query({active: true, currentWindow: true});
+    //TODO: define and implement sane behaviour for if the title is blank, but the URL
+    //      that is being bookmarked is not actually the current tab. just using 
+    //      await activeTab().title doesn't make sense for that, we'll need to change the tab query
```
bovine3dom: You found `activeTab` after all? I guess your problem was the await - you need `(await activeTab()).title`, otherwise the dot tries to operate on the promise returned by activeTab. It would be much clearer if `await` was just a function rather than a magic operator but, you know, it's JavaScript...

src/excmds.ts
```diff
@@ -3072,7 +3072,38 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
+        //function callback for promise for getting title of current tab
+        //requires [tabs] permission which tridactyl should have grabbed anyway
+        if (tabs[0]) {
+            var currentTab = tabs[0];
+            currTitle = currentTab.title
+        }
+    }
+    var titleSourceTab = browser.tabs.query({active: true, currentWindow: true});
+    //TODO: define and implement sane behaviour for if the title is blank, but the URL
+    //      that is being bookmarked is not actually the current tab. just using
```
bovine3dom: if `url` is defined, don't bother trying to find a title, just use the URL. Similarly, if the tab title is undefined, use the URL.

We could do more complicated things with requests but I'm not too bothered. I suspect most people only ever bookmark the current page they are on, or add a bookmark in their RC and specify a title.

src/excmds.ts
```diff
@@ -3072,7 +3072,38 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
+    await titleSourceTab;
+
+    // if title is blank, get it from the page.
+    if (title == "") {
+        title = currTitle
+    }
+
+
+
     if (path != "") {
```
bovine3dom: We probably need to skip this step if we're getting the title from the page - we don't want "BBC News - '/' becomes part of the alphabet" to end up in a "BBC News - " folder.

src/excmds.ts
reversebreak: I only found that helper function in the last moments of development! 
I considered using activeTab after I found it, but wasn't sure about simplifying things in case I needed to edit the lookup again.

reversebreak:
Alright, sorry about the delay in replying.  Thanks for taking such a thorough look at the PR - I'll get on to cleaning these up.

(Can you tell I'm not a native JS/extension coder?   :grimacing:  I'm usually in the comfy land of VisStudio and SQL.  Thanks for your patience.   :simple_smile:  )

src/excmds.ts
reversebreak: `path` is actually determined before the function ever hits the default title code, so there's no issue here.  Cases are:
* Bookmark function is passed a 'title' including a path, in which case the first few lines of code pick it up and the default blank title code never triggers, then the path processing code triggers
* Bookmark function is passed a title with no path.  Default title code never triggers, and neither does the path processing code.  
* Bookmark function is passed no title at all.  The default title code triggers and grabs the current tab's title.  Whatever that title is doesn't matter, as the path var is already just nothing and thus the path processing block doesn't trigger.

bovine3dom:
I think using the tab title is a better idea. : )

src/excmds.ts
```diff
@@ -3079,6 +3079,7 @@ export async function bmark(url?: string, ...titlearr: string[]) {
     // bookmarking, but too unlikely to bother with for now
     if (title == "") {
         //retrieve title from current tab
+        var currTitle
```
bovine3dom: You don't need var statements at all :)

You can just revert this commit.

src/excmds.ts
```diff
@@ -3084,6 +3097,15 @@ export async function bmark(url?: string, ...titlearr: string[]) {
         }
         const validpaths = flatten(treeClimber(tree, "")).filter(x => "path" in x)
         title = title.substring(title.lastIndexOf("/") + 1)
+        //technically an initial title string like `Firefox/` can give us a blank title
+        //once we remove the path, so let's fix that
+        if (title == "") {
+            //retrieve title from current tab
+            var currTitle
+            let currTitle = (await activeTab()).title
```
bovine3dom: These three lines can just be replaced with `title = (await...`. currTitle doesn't do anything.

src/excmds.ts
```diff
@@ -3072,7 +3072,20 @@ export async function bmark(url?: string, ...titlearr: string[]) {
     dupbmarks.map(bookmark => browser.bookmarks.remove(bookmark.id))
     if (dupbmarks.length != 0) return
     const path = title.substring(0, title.lastIndexOf("/") + 1)
-    // TODO: if title is blank, get it from the page.
+
+
+    // if title is blank, get it from the current page.
+    // technically could race condition if someone switched tabs REALLY quick after
+    // bookmarking, but too unlikely to bother with for now
+    if (title == "") {
```
bovine3dom: You need to check if `url` is defined as an argument, so move all of this above the ternary `url = url ===` line, otherwise you'll get your current tab's title for unrelated things.

reversebreak: I'm not sure I follow you - what kind of call to bmark are you thinking of that would be an 'unrelated thing'?

bovine3dom: e.g. `bmark google.com` on this page would end up with the title from this page

bovine3dom: `bmark https://www.google.com` while you were on a random page ends up with the title from that page, not "Google" or "google.com"

src/excmds.ts
```diff
@@ -3084,6 +3095,14 @@ export async function bmark(url?: string, ...titlearr: string[]) {
...
         const validpaths = flatten(treeClimber(tree, "")).filter(x => "path" in x)
         title = title.substring(title.lastIndexOf("/") + 1)
+        //technically an initial title string like `Firefox/` can give us a blank title
+        //once we remove the path, so let's fix that
+        if (title == "") {
+            //retrieve title from current tab
+            let currTitle = (await activeTab()).title
+            title = currTitle
+        }
+
```
bovine3dom: I'm not sure this is your fault, but 

`bmark google.com` gives me the error `# Error: Type error for parameter query (Value must either: be a string value, or .url must match the format "url") for bookmarks.search. ` which is slightly confusing - you could try using the `forceURI` function.

bovine3dom:
It's been a while since the author of this PR commented, so I'd be happy for anyone else to take it over. It should be a nice route into getting used to developing Tridactyl.

Thanks to reversebreak for all their work on this so far - I understand that people are busy : )

reversebreak:
Ah, sorry for going dark.

I was very busy, and this kinda dropped off the back of my mind.

I do feel that perhaps I bit off a bit more than I can chew - if anyone wants to take what I've done that'd be perfectly fine.
If not, I'll probably be back eventually, though I'll need to set up a Linux VM or something so that the commit scripts don't scream every time I try interacting with Git.

PR #873, opened 2018-07-29
Fix another tuple unpacking on the wrong side of brackets

MoSal:
Missed in 66af0b8be4abd9e7f9683b86eb240b105336e928


bovine3dom:
Ta. These really ought to have been caught by the type checker. Would be worth finding out what we're doing wrong so that they can be caught in future.

PR #881, opened 2018-07-30
native_messenger: fix invalid syntax

poinck:
fix invalid syntax for `os.path.join(...`

this fixes "Native messenager not found" bug after installation on a Gentoo linux system.


bovine3dom:
Thanks!

PR #884, opened 2018-07-31
Add auconenable config check.

### LINKED ISSUE #883: Incompatible with the Multi-Account Containers extension
I can open an [alternative account container tab](https://github.com/mozilla/multi-account-containers/) fine, but when I enter a URL it switches back to the default container.

Maybe you could add an optional toolbar button to quickly toggle off Tridactyl when I want to use an alt account briefly?

antonva:
Ignores all autocontain directives unless config variable is set to
"true". This can tide us over until I actually sit down and write it 
to be compatible with the Mozilla container extensions.

This resolves #883 and #882

### DIFF
--- a/src/lib/autocontainers.ts
+++ b/src/lib/autocontainers.ts
@@ -71,9 +71,9 @@ export class AutoContain implements IAutoContain {
     autoContain = async (
         details: IDetails,
     ): Promise<browser.webRequest.BlockingResponse> => {
-        // Lets not break everyone's user experience ok?
-        let enabled = Config.get("auconenable")
-        if (enabled === "false") return { cancel: false }
+        // No autocontain directives, no nothing.
+        let aucons = Config.get("autocontain")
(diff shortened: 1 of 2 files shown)


src/lib/autocontainers.ts
```diff
@@ -71,18 +71,22 @@ export class AutoContain implements IAutoContain {
     autoContain = async (
         details: IDetails,
     ): Promise<browser.webRequest.BlockingResponse> => {
+        // Lets not break everyone's user experience ok?
+        let enabled = Config.get("auconenable")
```
bovine3dom: Please just have `config.get("wherever the autocontains are stored") !== []` or the sane equivalent

antonva:
As per request.

the case for an actual config to enable this is to make living with the feature easier. With the proposed changes people have to unset their autocontains in order to have basic functionality like `:tabopen  -c` work again.

bovine3dom:
Oh, I see. Feel free to add your setting back set to true by default &&'d with the empty check.

PR #892, opened 2018-08-04
tabmove: Treat pinned and unpinned tabs as separate groups

### LINKED ISSUE #877: Circular tabmove broken by pinned tabs
Steps to reproduce:

- Pin tab with A-p
- Make two new tabs
- Press `<<` repeatedly

Expected behaviour: tab moves to beginning and then to end, forever and ever

Actual behaviour: tab gets stuck at beginning next to pinned tab.

Related: #161.

MoSal:
Fixes #877


bovine3dom:
Works well, thanks!

PR #900, opened 2018-08-07
Prevent cutting off a few pixels at the top

henrebotha:
The minimal userChrome file causes 8 pixels to be cut off from the top of the "content" part of the page, but only when the navbar is hidden. This change, as written, will reverse the situation: it will add 8 pixels of extraneous padding when the navbar is not hidden. Given that we spend most of our browsing time with the navbar hidden, this optimises for the common case instead of the uncommon case.

### DIFF
--- a/src/static/css/userChrome-minimal.css
+++ b/src/static/css/userChrome-minimal.css
@@ -77,6 +77,10 @@
     padding-top: 7px !important;
 }
 
+#main-window[sizemode="maximized"] #content-deck {
+    padding-top: 8px;
+}
+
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks for this. Couple of short points:

- I'm at a conference right now so won't have time to review (i.e. test) this until next week
- userChrome is kind of deprecated, you might want to look at `guiset` which starts in excmds.ts.

Thanks again for this - the reduced margin at the top annoys me a lot...

bovine3dom:
@henrebotha have you had a chance to look at fixing this for `guiset`?

PR #928, opened 2018-08-19
Update typescript to 3.0.1

antonva:
TypeDoc is updated to 0.12.0 as that's the latest release to support TS 3.0.


src/excmds.ts
```diff
@@ -1763,17 +1763,24 @@ export async function tabclosealltoleft() {
...
-    const sessions = await browser.sessions.getRecentlyClosed()
+    const sessions = await browser.sessions.getRecentlyClosed({ maxResults: 10 })
 
     // The first session object that's a window or a tab from this window. Or undefined if sessions is empty.
-    let closed = sessions.find(s => {
-        return "window" in s || (s.tab && s.tab.windowId == current_win_id)
+    const lastSession = sessions.find(s => {
+        if (s.window) {
+            return true
+        } else if (s.tab && s.tab.windowId) {
```
bovine3dom: Why did you move check down? If I close tabs in another window I don't want undo not to work...

antonva: Oh right, oversight on my part. Got interrupted while rewriting that. The check was moved down in another implementation where it would have made sense.

bovine3dom:
LGTM :)

bovine3dom:
NPM complains about

`npm WARN ts-jest@21.1.3 requires a peer of typescript@2.x but none is installed. You must install peer dependencies yourself.`

should I be worried?

PR #946, opened 2018-08-24
Honor config.profiledir in getProfileDir

willghatch:
I have my profiledir in a non-standard place.  The `getProfileDir` function can't find it automatically, but it also doesn't look at the `config.profiledir` variable which I set.  This fixes it.

Since the default value of `config.profiledir` is `"auto"`, I put the check before any automatic searching.  However, I think it would be nice to use the `--profile` argument but fall back to the configuration setting if that doesn't work.  In other words, I would like it to be automatic in the case where the automatic handling works, but defer to my configuration when the reliable search fails.  Since that behavior doesn't match what I expect when I change a variable *away* from `auto`, I have not implemented that.

Let me know your thoughts on the options here.  Maybe it would make more sense to do what I explicitly didn't.

### DIFF
--- a/src/native_background.ts
+++ b/src/native_background.ts
@@ -208,7 +208,7 @@ export async function firstinpath(cmdarray) {
     let ind = 0
     let cmd = cmdarray[ind]
     // Try to find a text editor
-    while (!await inpath(cmd.split(" ")[0])) {
+    while (!(await inpath(cmd.split(" ")[0]))) {
         ind++
         cmd = cmdarray[ind]
         if (cmd === undefined) break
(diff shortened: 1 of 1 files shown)


bovine3dom:
Having it try automatically when your config fails would be surprising, I think.

cmcaine:
Always automatic then falling back to manual config is definitely counter-intuitive. You were right to implement it this way.

Do you want that so that you can use the same tridactyl config on different machines? I think we expose the hostname somewhere so you could just set that configuration variable from a `:js` command with an if statement.

If your use case is very different and you don't think that kind of thing suits you we could look at other solutions.

willghatch:
Using a `:js` command should be fine.  Thanks for making Tridactyl!

PR #947, opened 2018-08-24
use /usr/bin/env to find bash

willghatch:
`/usr/bin/env` is a more reliable way to find `bash`, because it doesn't always live at `/bin/bash`.  In particular I am using NixOS, so `/bin/bash` doesn't work for me.  But in general I think `/usr/bin/env` will be available on any system where `/bin/bash` is available.

### DIFF
--- a/scripts/wine-pyinstaller.sh
+++ b/scripts/wine-pyinstaller.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash -e
 
 # This script must be run from the root Tridactyl directory
 TRIDIR="$(pwd)"
--- a/hooks/pre-commit
+++ b/hooks/pre-commit
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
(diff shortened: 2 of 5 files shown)


bovine3dom:
Huh. I could have sworn we'd already done this. I guess we got lax somewhere.

scripts/wine-pyinstaller.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash -e
```
bovine3dom: FWIW, this doesn't work on Linux: https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e

willghatch:
Sorry about that.  I think I already knew that at some level but forgot.  Is starting bash with `-e` the same as using `set -e` after the `#!`?  The `-e` flag is not in (or at least at the top with the other flags) of my man page for some reason.

cmcaine:
Yep, same thing.

Issue #952, opened 2018-08-26
tabopen and hinting navigations should consult autocontain directives


saulrh:
We have to keep the "hook webRequest.onBeforeRequest, close and reopen in right tab" logic to handle mouse clicks and other navigations, but when tridactyl kicks off a navigation itself it should be able to avoid the open-close-and-reopen dance using the autocontain configuration to set -c automatically.

antonva:
@saulrh  can we close this now #953 is merged?

saulrh:
Oops, yes, this feature was added in #953, and it doesn't depend on the as-yet-unobtained management permission so it actually works.

Fixed by #953.

PR #953, opened 2018-08-27
Autocontain coexistence

### LINKED ISSUE #951: autocontain un-contains all requests not caught by an autocontain directive
https://github.com/cmcaine/tridactyl/blob/master/src/lib/autocontainers.ts#L171
https://github.com/cmcaine/tridactyl/blob/master/src/lib/autocontainers.ts#L86

parseAucons handles the "not matched" case by autocontaining the tab in the default container. This is fine if tridactyl is the only container-related extension operating and tabs are in the default container by default, but if you have MAC or the [auto temporary containers](https://addons.mozilla.org/en-US/firefox/addon/temporary-containers/) extensions installed, tridactyl will into fights with them - they open in a temp container/"always open in" container, tridactyl reopens in the default container, infinite tab-reopening loop.

I think that this default is a bit unexpected. It took me a while to figure out why experimentally turning on a single autocontain directive made *every* request go into an infinite loop, which is what happens when you have auto temp containers configured.

Note if you close your browser while this kind of configuration is set and have "continue where I left off" set it'll continue when firefox reopens and tries to open the first tab, which will hang the UI hard enough that you won't be able to recover your session. If tridactyl is fighting with an auto temp container the temp container count can get into the hundreds in a hurry, which can't be good for the browser. I was lucky enough to have tridactyl configured to read a tridactylrc and could use that to clear out my autocontain directives before restarting FF again.

saulrh:
Does most of #951 and most of #952.

This has a few parts.

First, we need a bit of code for keeping track of what other extensions are installed and enabled. This isn't completely trivial because we need to listen for "on {en,dis}abled" and "on {,un}installed" events. This requires a new permission, `management`, which is _annoying_, but not having this permission would require extension detection to be done using kludgy messaging hacks that would be on the critical path for essentially every navigation operation.

Second, we need to write code to talk to the other addons and ask them if they're handling things. Thankfully they do provide public APIs and and we can use their sample code to do exactly what we need to do.

Third, it turns out some important chunks of the firefox webext API aren't handled by the web-ext-browser ts declaration we're using. They *are* handled by a PR on @types/firefox-webext-browser, though, so we can copy and paste that to make TS happy.

Fourth, and finally, we need to add some code to the autocontainer logic to use the compatibility code. This is pretty easy, but autocontainer logic starts taking a noticeable amount of time becuase of all the sequences awaits we're doing, so I also have to tweak things to do all of the async stuff in parallel.

#951: Should be no more infinite tab-creation loops. It's not entirely done yet, because I didn't *actually* give -c priority over autocontainers, so if you don't have an autocontain directive set for a domain and try to `tabopen -c` that domain it'll just go back to the ff default. However, I *did* give the Temporary Containers extension priority over autocontainers, and Temporary Containers will not un-contain tabs that are in containers, so you can work around the last bits of 951 by installing the Temporary Containers addon. Which is probably a good idea anyway, since that means you don't have this "default" container laying around that still leaks your information just as much as before.

#952: Tabopen will now consult autocontain directives when doing its thing. Anything that goes through tabopen should benefit from this logic, which I think also means that hinting and suchlike will autocontain without the create-close-recreate flickering that costs cycles and messes up tree-style tabs. I'll wait a bit before closing that in case there are some more hinting modes that need special handling.

I'm sure that there are *all* sorts of style and best-practices fails here - my limited exposure to web technology means that my approach to typescript depends primarily on chicken blood and artfully arranged magazine clippings. And I'm sure that I've forgotten to update some test or docstring to boot. I'm promising myself a banana split if CI succeeds on the first go. :V


bovine3dom:
I really don't want to add any more permissions until we sort out #708. About a third of our users get stuck on old versions every time we ask for more permissions...

src/manifest.json
```diff
@@ -49,6 +49,7 @@
         "clipboardRead",
         "downloads",
         "history",
+	"management",
```
bovine3dom: Please sort out tab spacing (I'm surprised prettier didn't catch this)

src/excmds.ts
```diff
@@ -198,7 +199,7 @@ export async function editor() {
     let tab = await activeTab()
     let selector = await Messaging.messageTab(tab.id, "excmd_content", "getInputSelector", [])
     let url = new URL(tab.url)
-    if (!await Native.nativegate()) return
```
bovine3dom: We should probably add a pre commit test for this. `!await` is almost always a bug...

bovine3dom:
Please also update docs/AMO.md to explain why we need the management permission.

saulrh:
> About a third of our users get stuck on old versions every time we ask for more permissions...

Blast. I'd forgotten about that part.

I'll split this up into a dirty version and a patch that fixes it. You can turn this entire thing off by not having any autocontainer directives, so if it has stability or performance problems people can just turn it off until we have an update pesterment system.

bovine3dom:
That's OK - I should really get around to #708 sometime anyway. Should probably just have a stable_manifest.json and beta_manifest.json on the build bot, bake the manifest into the storage, check for any differences, prompt user to update (i.e: take them to the AMO or the beta page).

saulrh:
> Please also update docs/AMO.md to explain why we need the management permission.

Done. Any edits needed on the copy?

The `Access browser activity during navigation` line is also out of date, I think? I can't be _sure_ that the human-readable text corresponds to the `webNavigation` permission, but if it does, I think we use it for autocommands in this very PR. Update while I'm here?

doc/amo.md
```diff
@@ -50,6 +50,8 @@ Since Tridactyl aims to provide all the features Vimperator and Pentadactyl had,
     *   The URLs of websites you've visited previously can be suggested as arguments for `:tabopen` and similar commands.
 *   Access recently closed tabs:
     *   If you've accidentally closed a tab or window, Tridactyl will let you open it again with the `:undo` command which is bound to `u` by default.
+*   Monitor extension usage and manage themes:
```
saulrh: I *think* that this is the line produced by the `management` permission but I'm not certain; is there a resource that lays out the mapping, or do we guess by randomly sampling extensions and comparing their requested permissions against their manifests?

bovine3dom:
> Update while I'm here?

Yes please : )

bovine3dom:
Unblocked (see #708). Is this ready to be merged/reviewed, @saulrh?

saulrh:
Right, finally remembered to come back around and bring this back up to date. This is ready to be reviewed and merged.

bovine3dom:
@antonva you wrote most of this autocontain code, would you mind reviewing this?

src/lib/autocontainers.ts
```diff
@@ -158,12 +173,86 @@ export class AutoContain implements IAutoContain {
...
+        let macAssignment = await browser.runtime
+            .sendMessage(
+                ExtensionInfo.KNOWN_EXTENSIONS.multi_account_containers,
+                {
+                    method: "getAssignment",
+                    url: details.url,
+                },
+            )
+            .catch(error => {
+                // TODO: figure out a good way to recover?
```
antonva: I don't know if there's any state to recover if MAC fails. Shouldn't we just assume that if it does, aucons take over?

We should also add logging here I feel.

saulrh: Added some logging.

I was actually debating whether continuing was the right thing to do here. If we've gotten to this point, then theoretically MAC is installed; if this error is a transient communications failure or some bug in the communication, then continuing could lead to MAC and aucons fighting over their containers and looping infinitely.

saulrh:
> Something I noticed during testing is that redirected sites do not stay in their container, which is by design but probably surprising to the user. This however stopped happening after testing MAC/temp-containers, which is a bit baffling.

> I think documenting the design of autocontainers is probably the only way to address the surprise but the fact that it stopped happening at some point while testing MAC/temp-containers is worrying.

I suspect that, like containers themselves, some of the MAC/temp-containers settings related to redirects are just thin wrappers around out-of-the-box browser features that just aren't exposed by default. MAC/temp-containers could plausibly change these settings and then fail to clean up after themselves when you're done testing.

I'll add some documentation and try to dig around in FF's containers settings to see if I can find anything.

antonva:
Looks good to me. @bovine3dom wanna give it a glance and/or merge?

bovine3dom:
Can you give me a quick description of what it fixes? Short on time ATM and the walls of text scared me off.

saulrh:
There are a number of other extensions that work with containers. For example, there's a temporary containers extension that will intercept anything that tries to open in the default container and put it in an ephemeral temporary container just for that website, and Mozilla's own Multi-Account Containers code allows you to manually configure durable containers which particular domains will be put into.

The issue is that these extensions work by intercepting a request, stopping it, closing the tab the request is in (or preventing it from opening a new tab in the first place), and then re-opening the URL in a new tab with a new container. To make it worse, sometimes this is glitchy and the original tab sticks around alongside the new one. If you have several of these extensions at once and they don't collaborate, you can end up with sorceror's apprentice mode: MAC is configured to open url in container A, Tridactyl is configured to open url in container B; url opens, MAC reopens in A, Tri reopens in B but glitches and you have two copies of url, MAC reopens both in A, Tri reopens both in B but glitches and now you have three tabs...

This PR's changes: Whenever Tridactyl is about to reopen a URL in a container (`:aucon`), it uses the "manage extensions" permission to first go around asking other common container-management extensions if they exist and, if so, if they want to handle the URL. If one of them does, tridactyl cedes control to the other extension; if none of them want it, tridactyl proceeds with the autocontainer code. This doesn't work perfectly, but it generally does a good job preventing the runaway loop of infinite open tabs.

...

Sorry for adding to the wall of text. >_>

bovine3dom:
Ah, in which case, we should fix #1360 first, as this adds a new permission. We still have quite a few people on old versions from the last time we needed more permissions.

saulrh:
I think I'll do that, yeah. Might put this behind a config option or something too. It is quite annoying to keep it synced.

saulrh:
I think that that should do the trick. Note that I haven't tested to make sure that the way I've blocked it actually _works_, since positive verification would require blowing up my browser.

Issue #727, opened 2018-06-22
autocmds can run in a different tab to the one they were triggered in


bovine3dom:
Steps to reproduce:

1. `autocmd DocStart www.reddit.com urlmodify -t www old`
2. `open news.bbc.co.uk`
3. `tabopen www.reddit.com`
4. Hit Ctrl-Shift-Tab as quickly as possible
5. Watch as old.bbc.co.uk is loaded.

bovine3dom:
Alternative, more annoying steps to reproduce:

1. `autocmd DocStart www.reddit.com urlmodify -t www old`
2. `open search reddit.com`
3. Open a reddit link in a background tab with `F`
4. Watch as old.google.co.uk is loaded

sjug:
I'm seeing this quite a bit, how can I disable these autocmds?

antonva:
@sjug Have you set any autocmds yourself? You can double check by entering `:viewconfig autocmds`. This should contain all autocommands you might have issued at some point or another.

If you're experiencing the exact issue @bovine3dom is describing it could be that you are using the example `tridactylrc` which does indeed have the `autocmd DocStart www.reddit.com urlmodify -t www old` directive.

That being said, to remove an issued aucmd, on the latest betas you can use `:autocmddelete` from #750. If you're using a version prior to that your best bet is to issue `:unset autocmds` from the command line.

bovine3dom:
(and remove it from your Tridactyl RC)

antonva:
Even after making sure that the DocStart event was being fired from the expected tab I noticed something. The aucmd doesn't use the content script provided `urlmodify` but rather the one provided via `background.js`.

`function urlmodify(mode, ...args) {
    return Object(__WEBPACK_IMPORTED_MODULE_0__messaging__["messageActiveTab"])("excmd_content", "urlmodify", [mode, ...args]);
}`

Which then proceeds and urlmodifies the active tab as expected.

bovine3dom:
Why refactor when you hack around bad decisions?

`autocmd DocStart www.reddit.com js tri.excmds.urlmodify("-t www old")` works fine.

PR #958, opened 2018-08-29
Fix the incorrect invocation of urlmodify in the tridactylrc example

quinoa42:
The old one (first mentioned as a workaround in #727) doesn't work for me so I checked the function signature in [excmds.ts](https://github.com/cmcaine/tridactyl/blob/15140b1bea64cdfff887f837a87b2abe2e6a9a76/src/excmds.ts#L1238).

This PR gives the workable example (tested on macos, firefox 61.0.2, tridactyl 1.13.3).


bovine3dom:
I apparently have been using a placebo for the last month without noticing...

Thanks!

PR #962, opened 2018-09-01
Parsers and state in content

saulrh:
Based off the autocontainers code because I was being lazy, which I should probably undo.

I've been programming on pure instinct for like five hours and I'm not sure what exactly I did any more. I *think* that what I did was something like:

* Split `state.ts` into `state.ts` and `content_state.ts`, then removed mode from state and moved it to content_state.
* Made the types in content_state.ts a hair more powerful
* Fixed all errors resulting from those two changes, in the process swapping state out for content_state wherever appropriate.
* Moved a couple excmds from background into content since they depended heavily on the mode and should live with the mode
* Split the controller in half, moving the parser logic to the content script and leaving acceptExCmds in the background version. content_controller forwards excmds to background script using messaging.
* Nuked the keydown_* code, since we no longer need to forward keys to the background.
* Went around fixing everything resulting from *those* changes, mostly the various perversions of keydown listeners
* Various tweaks here and there that probably didn't do anything but I was just changing things at random a third of the time and I really need to go through this and clean it up.

Things that work:
* excmds are nice now. each tab has its own commandline; they stay open when you navigate between tabs and you can come back to them after doing stuff in another tab and finish your command line input.
* keybinds that don't don't involve mode switching: `[[`, `]]`, `j`, `k`, `r`, etc, all still functional.
* You can turn on hinting mode and, again, navigate to another tab which will *not* be in hinting mode, do whatever you need to do there including excmds or more hinting, and then come back to finish the hint you started in the first tab
* You can exit hint mode by pressing escape. :P

Things that I broke:
* ...Honestly no idea, since I've barely tested it.
* I don't seem to be able to *finish* a hint any more. Sometimes I can't even get one key in.

This is, right now, a super janky proof of concept. DO NOT MERGE YET. I'm just putting it here so it doesn't get lost if I collapse in a sleep-deprived mess or go AWOL or something, so I can get feedback on the approach, and to see if anyone can figure out all the stupid things I did that will result in really subtle errors that I won't be able to find in my own testing.


bovine3dom:
Regressions I've found thus far:

- `# Error: Could not establish connection. Receiving end does not exist. 4 [object Object] ` on closing a tab (regardless of how it is closed). The number changes.
- `# DataCloneError: The object could not be cloned. 5 [object Object] ` on pressing any key in any hint mode

(will update this as I find more)

src/keydown_content.ts
```diff
@@ -1,119 +0,0 @@
...
-let insertparser = keys => generic.parser("imaps", keys)
-let ignoreparser = keys => generic.parser("ignoremaps", keys)
-let keys = []
-
-function genericResponse(ke: KeyboardEvent, parser: any): void {
-    keys.push(ke)
-    const response = parser(keys)
-
-    // Suppress if there's a match.
-    if (response.isMatch) {
```
bovine3dom: these lines were important - it's what made Tridactyl binds override Firefox binds, if they existed

src/content_state.ts
```diff
@@ -0,0 +1,61 @@
+import Logger from "./logging"
```
bovine3dom: This file should probably be called `state_content.ts`.

src/keydown_content.ts
```diff
@@ -1,119 +0,0 @@
...
-    modeSpecificSuppression(ke)
-
-    Messaging.message("keydown_background", "recvEvent", [
-        msgsafe.KeyboardEvent(ke),
-    ])
-}
-
-// {{{ Bad key suppression system
-
-// This is all awful and will go away when we move the parsers and stuff to content properly.
```
bovine3dom: I think I'm just beginning to realise how awful this was - did we _really_ have two parsers running at the same time just so we could suppress the keys?

src/keydown_content.ts
saulrh: Key suppression regression looks to have been resolved. Moving on to fixing hint-mode so that I can dogfood it while hunting down the rest of the issues.

src/content_state.ts
saulrh: Done, whoops.

saulrh:
Force-pushed a new version that's based on master instead of #953.

saulrh:
> # Error: Could not establish connection. Receiving end does not exist. 9 [object Object] when switching tabs (with Ctrl-Tab) through an unresponsive / loading tab - e.g. https://goodblock.gladly.io

This looks to have not been a regression but actually something I surfaced with some of the debug logging I scattered around. It's fixed in #963 and I've rebased to include that PR.

cmcaine:
Re: didn't realise how bad it was

Yeah, I've been sad about this code for a while ;)

saulrh:
Hinting is now working again, and with far cleaner code to boot - No more round-trips through the background!

Now to get the same for excmds. ^__^

saulrh:
I think I have a weird regression where the content script stops responding on one page and then refuses to load on any further tabs or pages - it doesn't even appear to get to "content script loaded, boss!".

saulrh:
...Nevermind, pretty sure that that "regression" is actually a temp-installed addon not being totally decoupled from its files, so when I accidentally invoke "build an extension" after a failed invocation of the build script I get a half-loaded addon with broken everything.

cmcaine:
Thank you very much for this refactor :) We've known it needed to be done for a while and this will make Tridactyl a lot nicer!

saulrh:
I think that I managed to nuke msgsafe. In the process the type system detected and forced me to fix a bug that we probably had with our detection of elements with the ARIA role of `application`, which was being concealed by `pick` in `dom.ts` completely bypassing type safety.

saulrh:
I think I've resolved all identified regressions this PR is good for review. I wouldn't be surprised if it was covered in style guide violations; if you want to get it merged tonight you can go ahead and fix them yourself. The remaining checkboxes are all feature requests that I was hoping to get in before someone merged it, but it looks like the remainder are big enough that they'll want to have their own issue numbers and PRs and I'll move them to the issue tracker when appropriate.

src/keydown_content.ts
```diff
@@ -1,119 +0,0 @@
-/** Shim for the keyboard API because it won't hit in FF57. */
-
-import * as Messaging from "./messaging"
-import * as msgsafe from "./msgsafe"
-import { isTextEditable, getAllDocumentFrames } from "./dom"
-import { isSimpleKey } from "./keyseq"
-
-function keyeventHandler(ke: KeyboardEvent) {
-    // Ignore JS-generated events for security reasons.
-    if (!ke.isTrusted) return
```
bovine3dom: The removal of this line is what led to https://github.com/tridactyl/tridactyl/security/advisories/GHSA-7qr7-93pf-hr8f. Mentioning https://github.com/tridactyl/tridactyl/issues/1833 here so we get a link back to this line.

PR #995, opened 2018-09-09
Basic attempt at excmds in content

saulrh:
The next step in my possibly-quixotic quest to kill #727: Permit excmds to be executed entirely in the context of the content script.

This is *not* a v2 excmds rewrite; this is just a hack in `excmd_macros.py` to add shims in the opposite direction, some shuffling in the controller, and a _hideous_ hack in `acceptExCmd` to let us call it transparently from an excmd without knowing whether we're in background or content.

I'm not sure this change is mergeable; it's mostly up for bikeshedding. `repeat` is seriously busted no matter how you come down on it being a content-mode or background-mode excmd. If it's content-mode, then the repeat queue is per-tab and trying to repeat things like `tabprev` is super frustrating. If it's global, then content-mode excmds invoked from the content script (which is a ton of stuff) can't be repeated at all because they never make their way to the background script to be recorded. I've left it set to `background`, since it seemed like that behavior was much less WTF, but since I never actually use `repeat` I don't know what kind of behavior people expect from it and can't actually make a statement either way.

I'm not sure what fixing `repeat` would actually look like. `vim` has a convenient discriminator - repeat the last command that resulted in a change to the document. Since we don't have something like a document that we can watch for edits, I'm not sure we can get that kind of DWIM without, like, manually annotating all of our commands. Even that might not be sufficient, since many of our excmds are compositions of others.


bovine3dom:
I haven't quite had time to appreciate the full horror of `acceptExCmd` yet, but I don't see why we can't bake the magic excmds variable in.

I think we sometimes told people to use acceptExCmd in their own `js` scripts, so it would be nice to keep it sane for end users if at all possible.

saulrh:
acceptExCmd can be called with one argument again so users can invoke it from `:js`. All it cost was the remaining wisps of my soul.

I tried to get this working using `inContentScript` from `lib/webext`, but it refused to work for some reason; I can't figure out exactly why, but it looked like `inContentScript` was incorrectly reporting that tabs were in the background script at times. This is roughly what I tried:
```typescript
let stored_excmds
if (inContentScript()) {
    console.log("Controller set up in content mode")
    stored_excmds = excmds_content
} else {
    console.log("Controller set up in background mode")
    stored_excmds = excmds_background
}
```

saulrh:
While this has my attention - did you make a decision about how you wanted the `repeat` functionality to work? If so, I think that I might reimplement this; with the major issues already identified and addressed it shouldn't be hard to reimplement it, probably faster than trying to get this merged.

bovine3dom:
No, I didn't think about it at all.

saulrh:
Obsoleted by #1489

PR #1012, opened 2018-09-17
Base16 Themes

bezmi:
I decided to modify the Shydactyl theme with base16 colours. I thought I'd incorporate all of the current base16 themes and start a pull request in case anybody else would like to use them too. 

LICENSE is in there because the author of base16 has licensed it under MIT, so I just included it as that is part of the terms, I'm new to contributing so I have no idea if this is necessary.

I might also modify the classic themes to base16 colours if people are interested.

### DIFF
--- a/src/static/themes/base16/LICENSE
+++ b/src/static/themes/base16/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Shahzeb Imran
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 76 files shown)


bovine3dom:
There seems to be a lot of redundancy between these. I'm reluctant to merge this as it is - I don't want to have to change 76 files every time we change something theme related.

Perhaps some concept of a palette that could be applied on top of a theme would make sense? Then we would get dark versions of all the themes for free.

bezmi:
you just need to run `set theme 3024` I wasn't sure when that pull request was going to be merged, but it will work much better that way and also allow me to publish these to the main base16 repo. 

The only things that change between files are the definitions of the 16 colours themselves. Is there some way for me to have the theme pallettes defined in separate files and then include them as variables to the main file that has the actual layout in it? Not sure how I could go about doing that, this is my first time playing around with css and web-extensions in general.

bovine3dom:
I'm not sure either, @bezmi. My suggestion which I thought about for all of 20 seconds was to have a `set palette` ex command in Tridactyl that set the variables after the theme.

It might be better to just allow people to apply as many themes as they like on top of each other.

cmcaine:
CSS files can include one another with @import so there's definitely no
need to duplicate the files in their entireties. Introducing a notion of
palettes might be better, but it seems like just introducing tridactyl
specific styling syntax. I think we should just expose the CSS if users
want to customise it.

bezmi:
I believe that is the right way to go here, I will close this pull request for now.

glacambre:
@bezmi #865 has been merged, are your themes available somewhere on github in order for us to mention them on our readme/soon-to-be wiki?

pjvds:
> @bezmi #865 has been merged, are your themes available somewhere on github in order for us to mention them on our readme/soon-to-be wiki?

Would love to have a single overview of the available themes ❤️

bezmi:
I deleted the original pull request repo, but I will set up a base16-tridactyl repo when I have time - not for several weeks.

PR #1020, opened 2018-09-20
Update config.ts: stop loading

FrankEular:
I added the command ":stop" and an nmap <C-c> to this command (as in vimperator and pentadactyl).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -159,7 +159,7 @@ class default_config {
         u: "undo",
         r: "reload",
         R: "reloadhard",
-        x: "js window.stop()",
+        "<C-c>": "stop",
         gi: "focusinput -l",
         "g;": "changelistjump -1",
         gt: "tabnext_gt",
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks.

Why did you remove the `x` bind? Is that not what it was bound to?

FrankEular:
just because it seems to me that vimperator and pentadactyl don't bind x to :stop by default (if their help page is genuine).

bovine3dom:
I'd rather it was bound to `x` by default. I'm sure _someone_ uses that as a bind for stop.

My reasoning is that Ctrl-C is widely used to copy text and this would break that (probably - or at least introduce weird side effects to copying). I'd be happier to bind it to Ctrl-C once we have a visual mode.

You're obviously welcome to bind it to whatever you want in your own configuration : )

bovine3dom:
I've changed the bind to `x` and merged. Thanks!

PR #1033, opened 2018-09-29
Clarify which find is used

lorenzleutgeb:


bovine3dom:
I think we can word this better still. We don't really use the Firefox find mode - we just don't bind anything to / so people can use it.

bovine3dom:
We should probably mention `unbind <C-f>` too.

PR #1038, opened 2018-09-29
Add build-webext command, for those of us that don't use web-ext run

saulrh:


bovine3dom:
Closing as per discussion on Riot - one can load manifest.json from build/ directly under about:debugging.

PR #1042, opened 2018-09-30
Fix build scripts on paths with spaces

joaomsa:
Fix build scripts broken if you clone the repo to a directory that has spaces anywhere along it's path like:
`~/workspace/some folder name with spaces/firefox/tridactyl`


bovine3dom:
Thanks!

PR #1065, opened 2018-10-07
WIP: Refactor hinting

lorenzleutgeb:
Not done yet. This is just to make you aware that I am working on this, and to collect early feedback if anyone feels like providing that.

My initial motivation was to implement cycling through active hints using `Tab`. But then, since there are many other issues that talk about hinting, maybe a more general clean-up would be beneficial.

I tried to extract what is common for both currently implemented hinting strategies (and probably future ones). That `BooleanFilter` thing is immutable so one might use it to navigate the history of hints.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -19,26 +19,98 @@ import {
     map,
     unique,
 } from "@src/lib/itertools"
-import { hasModifiers } from "@src/lib/keyseq"
 import { contentState } from "@src/content/state_content"
-import { messageActiveTab, message } from "@src/lib/messaging"
 import * as config from "@src/lib/config"
-import * as TTS from "@src/lib/text_to_speech"
 import Logger from "@src/lib/logging"
-import * as Messaging from "@src/lib/messaging"
 const logger = new Logger("hinting")
 
-/** Simple container for the state of a single frame's hints. */
+// A BooleanFilter is meant to represent a selection of active
+// elements of an array, while maintaining which elements have
+// been deactivated already.
+//
+// Invariant: pos.length + neg.length = xs.length
+// Invariant: set(pos) union set(neg) = set(xs)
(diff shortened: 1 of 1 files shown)


bovine3dom:
Ah, cool! Thanks for doing this. The hinting code hasn't had a critical eye on it for a long time.

I would really like hint mode to be a mode more like the others with a parser that lives with the rest of them and accepts user configurable binds.

Hinting is probably the weirdest mode we have as it has a lot of state. The other sensible modes we have bind key sequences to ex commands. I think we can still do that but we might want some namespacing so e.g. hint.select and hint.filter etc.

I haven't thought about this much.

lorenzleutgeb:
> I would really like hint mode to be a mode more like the others with a parser that lives with the rest of them and accepts user configurable binds.

Sorry to be that ignorant, but I have not checked how the other modes work. My plan would be to clean up hinting in itself first, which is easier, might still improve the hinting code and allow for new features. In a second step it could be normalized wrt. other modes, but I am not sure whether I'll have time for that. So maybe I'll just clean up what is there now and leave integration (of a nicer-to-integrate implementation, hopefully) to someone else.

bovine3dom:
That's fine! Tidying it up will definitely make it easier to migrate eventually.

saulrh:
I like the direction that this is going. :+1:

bovine3dom:
Have had a quick look at the code and it looks good.

Let us know if you need anything more specific : )

bovine3dom:
Hey @lorenzleutgeb, do you think you'll ever have the time to pick this up again?

lorenzleutgeb:
I have to admit that I since migrated back to Vimium-FF, but yeah I could definitely give this another go. It's just that I am a little busy, the idea of this PR also being that it would be easier for others to find and finish up what I started.

PR #1077, opened 2018-10-10
Extend zoom excmd keybindings and improve documentation.

antonva:
This patch improves upon the existing zoom excmd and adds some features present
in Vimperator and Pentadactyl. `zI` and `zO` are now bound to the maximum and
minimum zoom levels respectively. `zM` (read zoom more) zooms in on increments
of 50pp and `zR` (read zoom reduce) zooms out on increments of 50pp.

In addition, overshooting the zoom level while using relative adjustments is now
handled by setting the level to it's max/min zoom level.

An error is thrown if an absolute value is given that is out of bounds of the
allowed zoom range.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1797,10 +1797,28 @@ export function geturlsforlinks(reltype = "rel", rel: string) {
     return []
 }
 
+/** Sets the current page's zoom level anywhere between 30% and 300%.
+ *
+ * If you overshoot the level while using relative adjustments i.e. level > 300% or level < 30%
+ * the zoom level will be set to it's maximum or minimum position.
+ *
+ * @param level - The zoom level to set.
+ * Expects percentages when changing the absolute zoom value and percentage points when making relative adjustments..
+ * @param rel - Set the zoom adjustment to be relative to current zoom level.
+ */
 //#background
 export async function zoom(level = 0, rel = "false") {
     level = level > 3 ? level / 100 : level
-    if (rel == "true") level += await browser.tabs.getZoom()
+    if (rel === "false") {
+        if (level > 3 || level < 0.3) throw new Error(`[zoom] level out of range: ${level}`)
+    }
+    if (rel === "true") {
+        level += await browser.tabs.getZoom()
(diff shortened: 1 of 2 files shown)


src/lib/config.ts
```diff
@@ -250,7 +250,11 @@ class default_config {
         A: "bmark",
         zi: "zoom 0.1 true",
         zo: "zoom -0.1 true",
```
bovine3dom: dupe?

src/lib/config.ts
```diff
@@ -405,6 +409,7 @@ class default_config {
             "fillcmdline_tmp 3000 !jsb is deprecated. Please use jsb instead",
         current_url: "composite get_current_url | fillcmdline_notrail ",
         stop: "js window.stop()",
+        zo: "zoom",
```
bovine3dom: dupe?

src/excmds.ts
```diff
@@ -1797,10 +1797,28 @@ export function geturlsforlinks(reltype = "rel", rel: string) {
     return []
 }
 
+/** Sets the current page's zoom level anywhere between 30% and 300%.
+ *
+ * If you overshoot the level while using relative adjustments i.e. level > 300% or level < 30%
+ * the zoom level will be set to it's maximum or minimum position.
+ *
+ * @param level - The zoom level to set.
+ * Expects percentages when changing the absolute zoom value and percentage points when making relative adjustments..
```
bovine3dom: too many dots

src/lib/config.ts
antonva: One is shorthand on the commandline `:zo[om]` the other is a keybind.

src/excmds.ts
antonva: Gah, blaming my keyboard for that.

src/lib/config.ts
antonva: As in, one is an exalias the other a keybind.

bovine3dom: oh wow. ta.

PR #1086, opened 2018-10-15
Allow executing code in bookmarklets

shouya:
This PR add support to correctly execute codes in [bookmarklets](https://en.wikipedia.org/wiki/Bookmarklet) using `:bmarks` command. Originally users are able to click on a bookmarklet to execute a short piece of JavaScript code on the current page, this PR allows them to do the same thing without using a mouse.

Admittedly this feature has some security concerns, for which I'm not sure if this should be merged. But so far I found bookmarklet very convenient and irreplacable for certain tasks.


bovine3dom:
This is a great little feature. Thanks so much!

PR #1150, opened 2018-11-06
Only cycle through visible tabs. Fixes #1084.

### LINKED ISSUE #1084: tabnext, tabprev, tabnext_gt, etc should only cycle to non-hidden tabs

antonva:
This leverages @glacambre 's fix for the index problem . I'd argue that we should keep this no matter what Mozilla decide to do on the index bug since it's unreliable if the user is using tab hiding. Hidden tabs retain their index, creating the problem with `tabnext | tabprev` et al.  described in #1084 

The metadata stuff is just prettier changing stuff up as usual.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1703,12 +1703,11 @@ export async function tabprev(increment = 1) {
     // Proper way:
     // return tabIndexSetActive((await activeTab()).index - increment + 1)
     // Kludge until https://bugzilla.mozilla.org/show_bug.cgi?id=1504775 is fixed:
-    return browser.tabs.query({currentWindow: true})
-        .then(tabs => {
-            tabs = tabs.sort((t1, t2) => t1.index - t2.index)
-            let prevTab = (tabs.findIndex(t => t.active) - increment + tabs.length) % tabs.length
-            return browser.tabs.update(tabs[prevTab].id, {active: true})
-        })
+    return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
+        tabs = tabs.sort((t1, t2) => t1.index - t2.index)
+        let prevTab = (tabs.findIndex(t => t.active) - increment + tabs.length) % tabs.length
+        return browser.tabs.update(tabs[prevTab].id, { active: true })
+    })
 }
 
 /** Switch to the first tab. */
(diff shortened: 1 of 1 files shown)


bovine3dom:
I get build errors:

```
✖ ｢atl｣: Checking finished with 2 errors

ERROR in [at-loader] ./src/.excmds_background.generated.ts:1407:54
    TS2345: Argument of type '{ currentWindow: true; hidden: boolean; }' is not assignable to parameter of type '{ active?: boolean; audible?: boolean; cookieStoreId?: string; currentWindow?: boolean; discarded?: boolean; highlighted?: boolean; index?: number; muted?: boolean; lastFocusedWindow?: boolean; pinned?: boolean; ... 4 more ...; windowType?: WindowType; }'.
  Object literal may only specify known properties, and 'hidden' does not exist in type '{ active?: boolean; audible?:
boolean; cookieStoreId?: string; currentWindow?: boolean; discarded?: boolean; highlighted?: boolean; index?: number; muted?: boolean; lastFocusedWindow?: boolean; pinned?: boolean; ... 4 more ...; windowType?: WindowType; }'.

ERROR in [at-loader] ./src/excmds.ts:1706:54
    TS2345: Argument of type '{ currentWindow: true; hidden: boolean; }' is not assignable to parameter of type '{ active?: boolean; audible?: boolean; cookieStoreId?: string; currentWindow?: boolean; discarded?: boolean; highlighted?: boolean; index?: number; muted?: boolean; lastFocusedWindow?: boolean; pinned?: boolean; ... 4 more ...; windowType?: WindowType; }'.
  Object literal may only specify known properties, and 'hidden' does not exist in type '{ active?: boolean; audible?:
boolean; cookieStoreId?: string; currentWindow?: boolean; discarded?: boolean; highlighted?: boolean; index?: number; muted?: boolean; lastFocusedWindow?: boolean; pinned?: boolean; ... 4 more ...; windowType?: WindowType; }'.
```

antonva:
It's because your web-ext-types is outdated. rm -r node modules and npm install again.

antonva:
Mea culpa. I'll stop nuking package.lock changes.

Issue #1190, opened 2018-11-23
Overwrite error pages


bovine3dom:
VVimpulation hijacks Firefox error pages so it can still steal input on it. We should probably do the same: https://github.com/amedama41/vvimpulation/commit/c0efa1910a9e27d10134100b0e0f86fbf2a55995

and 

https://github.com/amedama41/vvimpulation/blob/master/background/wimpulation.js#L892

bovine3dom:
Closing as this proved pretty pointless in #1191. Hopes lie with keyboard-api instead.

PR #1191, opened 2018-11-23
[WIP] Error page overriding

bovine3dom:
See #1190.

Unfortunately, this doesn't seem to catch "you mangled the web address you just typed" errors. It does catch "you reloaded the page and then cancelled it really quickly" type errors, but I don't get those very often.

It would be worth checking whether any common errors are covered by this before proceeding. I really dislike the `details` object which seems to be undocumented.


bovine3dom:
Closing this as I don't care enough about the handful of error pages it would fix to warrant the effort.

PR #1196, opened 2018-11-26
excmds: Allow selector for all non-element hints

keegancsmith:
I wanted to support something like `hint -qbc selector`. I noticed we can
instead deprecate the `-c` flag and just let any hint command take a selector.

I am unsure if this is the correct approach, so this is just a minor change to
get what I want working. After feedback I'll make a more "correct" change
(including documentation updates).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3357,7 +3357,7 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
         case "-b":
             // Open in background
             selectHints = hinting.pipe(
-                DOM.HINTTAGS_selectors,
+                selectors || DOM.HINTTAGS_selectors,
                 async link => {
                     link.focus()
                     if (link.href) {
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -3443,7 +3443,7 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
 
         case "-W":
             selectHints = hinting.pipe(
-                DOM.HINTTAGS_selectors,
+                selectors || DOM.HINTTAGS_selectors,
                 elem => {
                     // /!\ RACY RACY RACY!
                     run_exstr(selectors + " " + rest.join(" ") + " " + elem)
```
bovine3dom: As I mentioned in Riot - we'll need to think of some way to tell if selectors are actually selectors or really part of rest[].

keegancsmith:
Sorry I don't have much bandwidth at the moment to work on this, but will likely have some time later in December.

keegancsmith:
Don't have time to work on this. Abandon ship.

Issue #1188, opened 2018-11-21
make command prefixes visible


MilanV:
There are several multicharacter commands where the first key typed does not do any immediate action. For example ";" (for extra-hint mode commands), "g" (for several commands) etc. However, the first typed key is not shown in any way. It would be comfortable (and less confusing) if there was a visual reaction.
For example a change of "normal" to "normal g" in the mode flag in the right-bottom corner as we do not have the real status bar in tridactyl.

PR #1201, opened 2018-11-28
Show command prefix in the "mode status" field

MilanV:
Solving Issue #1188

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -184,12 +184,19 @@ config.getAsync("modeindicator").then(mode => {
         })
     }
 
-    addContentStateChangedListener((property, oldValue, newValue) => {
+    addContentStateChangedListener((property, oldMode, oldValue, newValue) => {
+	let mode = newValue
+	let suffix = ""
+	let result = ""
         if (property != "mode") {
-            return
-        }
+	    if (property === "suffix") {
+		mode = oldMode
+		suffix = newValue
+	    } else {
+		return
+	    }
+	}
 
-        let mode = newValue
         const privateMode = browser.extension.inIncognitoContext
             ? "TridactylPrivate"
             : ""
@@ -210,8 +217,17 @@ config.getAsync("modeindicator").then(mode => {
             statusIndicator.textContent = "normal"
             // statusIndicator.style.borderColor = "lightgray !important"
         } else {
-            statusIndicator.textContent = mode
+            result = mode
         }
+	if ( suffix != "" ) {
+	    result = mode + " " + suffix
+	}
+	logger.debug (
+	    "statusindicator: ",
+	    result,
+	    ";"
+	)
+	statusIndicator.textContent = result
         statusIndicator.className +=  " TridactylMode" + statusIndicator.textContent
 
         if (config.get("modeindicator") !== "true") statusIndicator.remove()
--- a/src/content.ts
+++ b/src/content.ts
@@ -7,6 +7,7 @@ import "@src/lib/html-tagged-template"
 /* import "@src/content/commandline_content" */
 /* import "@src/excmds_content" */
 /* import "@src/content/hinting" */
+import * as Config from "@src/lib/config"
 import * as Logging from "@src/lib/logging"
 const logger = new Logging.Logger("content")
 logger.debug("Tridactyl content script loaded, boss!")
@@ -219,13 +220,16 @@ config.getAsync("modeindicator").then(mode => {
         } else {
             result = mode
         }
-	if ( suffix != "" ) {
+	let showcmd = Config.get("showcmd")
+	if ( showcmd === "true" && suffix != "" ) {
 	    result = mode + " " + suffix
 	}
 	logger.debug (
 	    "statusindicator: ",
 	    result,
-	    ";"
+	    ";",
+	    "config",
+	    showcmd
 	)
 	statusIndicator.textContent = result
         statusIndicator.className +=  " TridactylMode" + statusIndicator.textContent
--- a/src/content.ts
+++ b/src/content.ts
@@ -131,7 +131,9 @@ config.getAsync("modeindicator").then(mode => {
         ? "TridactylPrivate"
         : ""
     statusIndicator.className =
-        "cleanslate TridactylStatusIndicator " + privateMode + " TridactylModenormal "
+        "cleanslate TridactylStatusIndicator " +
+        privateMode +
+        " TridactylModenormal "
 
     // Dynamically sets the border container color.
     if (containerIndicator === "true") {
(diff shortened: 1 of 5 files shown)


src/content/controller_content.ts
```diff
@@ -12,6 +12,32 @@ import * as generic from "@src/parsers/genericmode"
...
+	result === "Meta" ||
+	result === "Alt" ||
+	result === "Shift" ||
+	result === "OS"
+	) {
+	    return ""
+	}
+
+    if (k.altKey) {
+	result = "Alt-" + result
```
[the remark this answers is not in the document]
MilanV: Good point. I changed that, see the additional commit.

src/content.ts
```diff
@@ -210,8 +217,17 @@ config.getAsync("modeindicator").then(mode => {
             statusIndicator.textContent = "normal"
             // statusIndicator.style.borderColor = "lightgray !important"
         } else {
-            statusIndicator.textContent = mode
+            result = mode
         }
+	if ( suffix != "" ) {
+	    result = mode + " " + suffix
```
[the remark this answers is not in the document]
MilanV: Thanks, that's a good idea. I hope this is the standard way of making things configurable as I did in my next commit - look at it, please.

MilanV:
I addressed all your requests. However, while testing my code I found one problem I do not know any solution of: keystrokes of plain FF commands (like <ArrowUp>) are not recognized as "eaten" by FF and the mode indicator includes such suffix until the user presses some tridactyl command keystroke (whatever, i.e. "j"). Do you have any idea? Or shall we live with that for now?

bovine3dom:
I'll look into it within a few days. If it wasn't obvious how to avoid it, it will probably require some moderate faffing around with the parser.

An aside: if you want GitHub to attribute your commits to your GitHub account, you need to tell GitHub about the email address you have set in git.

src/content.ts
```diff
@@ -219,13 +220,16 @@ config.getAsync("modeindicator").then(mode => {
         } else {
             result = mode
         }
-	if ( suffix != "" ) {
+	let showcmd = Config.get("showcmd")
```
bovine3dom: I don't like this name. I haven't thought of a better one yet. Perhaps `indicatorkeys`, and migrate `modeindicator` to `indicatormode`?

Posting this here so that I don't forget about it. I'm not expecting you to fix it : )

src/lib/config.ts
```diff
@@ -757,6 +757,12 @@ class default_config {
      *
      */
     perfsamples: string = "10000"
+
+    /**
+     * Show (partial) command in the mode indicator.
+     * Corresponds to 'showcmd' option of vi.
```
bovine3dom: Heh, I guess this explains why you chose it : )

bovine3dom:
@MilanV out of curiosity, did you have trouble with `prettier`? Somehow your branch has a mixture of tabs and spaces, which `npm run pretty` should have nuked - and you shouldn't have been able to commit (unless you use `git commit -n`) without it.

src/content/controller_content.ts
```diff
@@ -98,13 +99,6 @@ function* ParserController() {
                     response,
                 )
 
-		// show current keyEvents as a suffix of the contentState
```
bovine3dom: It was a pretty easy fix in the end - just had to move this block down until the parser returns the keys which could be part of a valid bind.

bovine3dom:
This is almost ready to merge. Just need to think about what we want to call `showcmd`. My objections to `showcmd`:

- Unlike Vim, key sequences aren't commands - we map them to ex-strings which are the commands - so we should relate the setting to keys or key sequences
- Our pseudo-namespacing is important for making things easy to find, so we ought to prefix it with something like `indicator`
    - We should (eventually) move `modeindicator` to the same naming scheme

My suggestion is `indicatorkeys` but I'm not married to it.

Also - should we default it to true or false?

Bike-shed away!

MilanV:
Thanks for fixing both the bug and the indentation. (I must check my vim settings why it made that mess.)

* the option (un)set by default: I believe that now, with only reasonable prefixes shown, it may be set by default, it's handy and not harming.
* the option name: renaming `modeindicator` to anything like `indicator_mode` does not make any sense - it would suggest what's not true (yet): that Tridactyl provides a generic indicator where one can switch on/off each part: mode name, keys typed... 
This is what I'd like for future but we are not there yet: making the indicator a real statusline which the user can format according his needs. `modeindicator` renamed to `statusline`, new option `statusline_format` with items like
  * `%m` for mode name
  * `%k` for hidden keys typed
  * `%l` showing the URL of focused link
  * `%h` showing the history status
  * `%b` showing the bookmark status of the page
  * etc.

Until then, let's call this option by any meaningful name easy to find. What about `showtypedkeys`?

bovine3dom:
Yeah, I agree that indicatormode was a bad idea.

We have a vague namespacing standard where the noun comes first and then the verb comes second, so you can find most tab related commands by typing `:tab`. I'd like us to do the same here. We could rebrand the mode indicator as just indicator but I'm not sure that's easier to find.

I hadn't meant to hold this up so long to decide on a name - I honestly just forgot about it : )

MilanV:
In fact, it's my "fault" (meaning a design fault): I kind of misused a modeindicator for something else, this is why we have problems with a name for it now. It's tight to modeindicator (it does not display without it) but naming it like `modeindicator_keys` does not make a good sense, it's not understandable.
We need a real `statusline` option in future, as I described it above :)
For now, I vote for `keystypeddisplay` - it starts with subject (as you described the 'policy'), it says what it does and the fact it depends on the `modeindicator` may be explained in the help message. Agree?

bovine3dom:
Feel free to make a new issue about the status line or add your thoughts to an existing one (it can't be the first time someone has brought it up).

---

I've been toying with the idea of calling the current thing a `statusindicator`. We can then have sub-settings like `statusindicator.showmode` and `statusindicator.showkeys` (and, eventually, `statusindiactor.style = line | nugget`). The downside is that completions don't work for nested keys, yet.

It seems really silly to hold this up over this, though, so I'm just going to call it `modeindicatorshowkeys`, and we can fix it later (along with the other 250-odd issues ;) ).

bovine3dom:
Oh. I hadn't noticed that. That sounds like another bike shed. I'll think about it. I'm concerned that it might be slightly distracting as eyes are sensitive to changes at the periphery, and I don't think many people want the feature. Please file an issue and we can think about it.

MilanV:
I do not think it's anything for new issue - if you are not sure about it, let it be false by default. I thought you just forgot that after all that fun with option renaming ;)

PR #1224, opened 2018-12-09
Cycle the hint focus through the available hints

furgerf:
Hi guys, I had some time to catch up with all the awesome changes to tridactyl - great job and thanks a lot for your effort! :+1: 

I was really looking forward to the vimperator-style hint filtering, great that we can do that now! However, I couldn't find out what the difference between `vimperator` and `vimperator-reflow` are. I saw that this has been asked before, so I added that to the documentation (at least I _hope_ I did).

Additionally, back when using vimperator, I would often filter links until just a few remain, then cycle through them with `<Tab>` and `<S-Tab>`. So, I had a look at `src/content/hinting.ts` and took a shot at implementing that. What's missing is the opposite direction - the `key` in the parser function doesn't seem to be a [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent), so I can't check `key.shiftKey`.

I hope this is something you'll find useful and maybe you can help with the shift-thing. Thanks in advance :)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -147,7 +147,7 @@ You can bind your own shortcuts in normal mode with the `:bind` command. For exa
 
 -   How can I change the colors or theme used by Tridactyl?
 
-    Use `:colors dark` (authored by @fugerf), `:colors shydactyl` (authored by @atrnh) or `:colors greenmat` (authored by @caputchinefrobles). Tridactyl can also load themes from disk, which would let you use one of the themes authored by @bezmi ([#1012](https://github.com/tridactyl/tridactyl/pull/1012)), see `:help colors` for more information.
+    Use `:colors dark` (authored by @furgerf), `:colors shydactyl` (authored by @atrnh) or `:colors greenmat` (authored by @caputchinefrobles). Tridactyl can also load themes from disk, which would let you use one of the themes authored by @bezmi ([#1012](https://github.com/tridactyl/tridactyl/pull/1012)), see `:help colors` for more information.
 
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -147,7 +147,7 @@ You can bind your own shortcuts in normal mode with the `:bind` command. For exa
 
 -   How can I change the colors or theme used by Tridactyl?
 
-    Use `:colors dark` (authored by @fugerf), `:colors shydactyl` (authored by @atrnh) or `:colors greenmat` (authored by @caputchinefrobles). Tridactyl can also load themes from disk, which would let you use one of the themes authored by @bezmi ([#1012](https://github.com/tridactyl/tridactyl/pull/1012)), see `:help colors` for more information.
+    Use `:colors dark` (authored by @furgerf), `:colors shydactyl` (authored by @atrnh) or `:colors greenmat` (authored by @caputchinefrobles). Tridactyl can also load themes from disk, which would let you use one of the themes authored by @bezmi ([#1012](https://github.com/tridactyl/tridactyl/pull/1012)), see `:help colors` for more information.
```
bovine3dom: Eek. Sorry!

bovine3dom:
Thanks for this. I'll look into your key stuff tomorrow. Hinting is a bit strange as the binds are hardcoded (I think it's the only remaining mode where that is true). It's possible that we just don't support modifiers yet and need to pass that information along.

bovine3dom:
LGTM. Thanks!

PR #1260, opened 2018-12-24
Add TST integration

rootkit-ty:
So I've added basic TST integration (tab control)

I'd love some feedback to make this work more cleanly.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1709,55 +1709,47 @@ export async function tabprev(increment = 1) {
 
     if (config.get("treestyletabintegration")) {
         // Ok so this entire piece here is really inefficient, it looks up all tabs, gets the active tab id, gets all tabs again (this time as a tree), iterates through all those tabs flattening them into an array, and then iterates over them once more to find the index of the active tab in that flattened array. This has a lot of room for improvement in the future.
-        return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
-            tabs = tabs.sort((t1, t2) => t1.index - t2.index)
-            // Find the current TAB id
-            let activeTabId = tabs[tabs.findIndex(t => t.active)].id
-            // Get the whole tab tree
-            browser.runtime
-                .sendMessage("treestyletab@piro.sakura.ne.jp", {
-                    type: "get-tree",
-                    tabs: "*",
-                })
-                .then(tabTree => {
-                    // Convert the Tree to a flat array
-                    function flattenVisibleTree(node) {
-                        // Damn Javascript type errors
-                        if (typeof node === "undefined") return []
-
-                        // There is no root parent node so the first iteration should be delt with differently
-                        var ids = []
-                        if (node instanceof Array) {
-                            node.forEach(child => {
-                                ids.push(...flattenVisibleTree(child))
-                            })
-                        } else {
-                            // Tree is collapsed, therefore it nor it's children are visible
-                            if (typeof node.states !== "undefined" && node.states.indexOf("collapsed") !== -1) return []
-
-                            ids.push(node.id)
-                            // Recurse over every child and add their ID array to ours
-                            if (typeof node.children !== "undefined")
-                                node.children.forEach(child => {
-                                    ids.push(...flattenVisibleTree(child))
-                                })
-                        }
-
-                        // Return our ids followed by our childrens as an array
-                        return ids
-                    }
-
-                    // TODO: Have this check if the tabTree is an array
-                    if (typeof tabTree === undefined || tabTree.constructor !== Array) return null
-                    let tabList = flattenVisibleTree(tabTree)
-
-                    let prevTab = tabList[tabList.indexOf(activeTabId) - increment]
-
-                    return browser.tabs.update(prevTab, { active: true })
-                })
-
-            //return browser.tabs.update(tabs[prevTab].id, { active: true })
+        // Find the current TAB id
+        let activeTabId = (await activeTab()).id
+        // Get the whole tab tree
+        let tabTree = await browser.runtime.sendMessage("treestyletab@piro.sakura.ne.jp", {
+            type: "get-tree",
+            tabs: "*",
         })
+        // Convert the Tree to a flat array
+        function flattenVisibleTree(node) {
+            if (typeof node === "undefined") return []
+
+            // There is no root parent node so the first iteration should be delt with differently
+            var ids = []
+            if (node instanceof Array) {
+                node.forEach(child => {
+                    ids.push(...flattenVisibleTree(child))
+                })
+            } else {
+                // Tree is collapsed, therefore it nor it's children are visible
+                if (typeof node.states !== "undefined" && node.states.indexOf("collapsed") !== -1) return []
+
+                ids.push(node.id)
+                // Recurse over every child and add their ID array to ours
+                if (typeof node.children !== "undefined")
+                    node.children.forEach(child => {
+                        ids.push(...flattenVisibleTree(child))
+                    })
+            }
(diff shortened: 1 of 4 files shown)


src/excmds.ts
```diff
@@ -1706,11 +1706,65 @@ export async function tabprev(increment = 1) {
...
-        let prevTab = (tabs.findIndex(t => t.active) - increment + tabs.length) % tabs.length
-        return browser.tabs.update(tabs[prevTab].id, { active: true })
-    })
+
+    if (config.get("treestyletabintegration")) {
+        // Ok so this entire piece here is really inefficient, it looks up all tabs, gets the active tab id, gets all tabs again (this time as a tree), iterates through all those tabs flattening them into an array, and then iterates over them once more to find the index of the active tab in that flattened array. This has a lot of room for improvement in the future.
+        return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
+            tabs = tabs.sort((t1, t2) => t1.index - t2.index)
+            // Find the current TAB id
+            let activeTabId = tabs[tabs.findIndex(t => t.active)].id
```
antonva: You could use  `activeTabId()` here instead of doing this. The definition lives in `src/lib/webext.ts` It has some comon shortcuts for things you're likely to do a lot when messing with excmds.

src/excmds.ts
```diff
@@ -1706,11 +1706,65 @@ export async function tabprev(increment = 1) {
     // Proper way:
     // return tabIndexSetActive((await activeTab()).index - increment + 1)
     // Kludge until https://bugzilla.mozilla.org/show_bug.cgi?id=1504775 is fixed:
-    return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
-        tabs = tabs.sort((t1, t2) => t1.index - t2.index)
-        let prevTab = (tabs.findIndex(t => t.active) - increment + tabs.length) % tabs.length
-        return browser.tabs.update(tabs[prevTab].id, { active: true })
-    })
+
+    if (config.get("treestyletabintegration")) {
```
bovine3dom: Can this go in a separate file? I'd rather have `if config.get... interop.tst.tab(...)` etc so if TST changes its API we only have to look at one file

src/lib/config.ts
```diff
@@ -773,6 +773,11 @@ class default_config {
      *
      */
     perfsamples: string = "10000"
+
+    /**
+     * Whether or not to integrate with the TST tab
+     */
+    treestyletabintegration: boolean = false
```
bovine3dom: I think this should default to true and we should use the `management` permission to check if TST is installed.

src/excmds.ts
```diff
@@ -1706,11 +1706,65 @@ export async function tabprev(increment = 1) {
...
+        // Ok so this entire piece here is really inefficient, it looks up all tabs, gets the active tab id, gets all tabs again (this time as a tree), iterates through all those tabs flattening them into an array, and then iterates over them once more to find the index of the active tab in that flattened array. This has a lot of room for improvement in the future.
+        return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
+            tabs = tabs.sort((t1, t2) => t1.index - t2.index)
+            // Find the current TAB id
+            let activeTabId = tabs[tabs.findIndex(t => t.active)].id
+            // Get the whole tab tree
+            browser.runtime
+                .sendMessage("treestyletab@piro.sakura.ne.jp", {
```
bovine3dom: You can swap the `then` for `let tabTree = await browser.runtime...` to avoid nesting but I'm not too bothered.

src/excmds.ts
rootkit-ty: Yeah, probably best to shove it in it's own file, wasn't quite sure of the structure of the whole project so I initially just shoehorned it into one existing function.

src/lib/config.ts
rootkit-ty: Maybe, I was more thinking to let the user enable it incase they have TST installed but not in use.

src/excmds.ts
rootkit-ty: Yeah, probably a good idea, I avoided it originally because I didn't want to change the function to async to avoid accidentally introducing issues.

src/excmds.ts
rootkit-ty: Where should I throw the file? Is there an existing structure that I should take advantage of?

src/excmds.ts
rootkit-ty: Fixed it so it uses activeTabId, and will be throwing it in it's own file.

src/excmds.ts
```diff
@@ -1706,11 +1706,66 @@ export async function tabprev(increment = 1) {
...
+    if (config.get("treestyletabintegration") && hasTST) {
+        // Ok so this entire piece here is really inefficient, it looks up all tabs, gets the active tab id, gets all tabs again (this time as a tree), iterates through all those tabs flattening them into an array, and then iterates over them once more to find the index of the active tab in that flattened array. This has a lot of room for improvement in the future.
+        // Find the current TAB id
+        let activeTabId = (await activeTab()).id
+        // Get the whole tab tree
+        let tabTree = await browser.runtime.sendMessage("treestyletab@piro.sakura.ne.jp", {
+            type: "get-tree",
+            tabs: "*",
+        })
+        // Convert the Tree to a flat array
```
bovine3dom: You might check config.ts for similar functions - I recall writing a tree walker for that.

src/excmds.ts
```diff
@@ -1706,11 +1706,66 @@ export async function tabprev(increment = 1) {
...
+    } catch (e) {
+        hasTST = false
+    }
+
+    if (config.get("treestyletabintegration") && hasTST) {
+        // Ok so this entire piece here is really inefficient, it looks up all tabs, gets the active tab id, gets all tabs again (this time as a tree), iterates through all those tabs flattening them into an array, and then iterates over them once more to find the index of the active tab in that flattened array. This has a lot of room for improvement in the future.
+        // Find the current TAB id
+        let activeTabId = (await activeTab()).id
+        // Get the whole tab tree
+        let tabTree = await browser.runtime.sendMessage("treestyletab@piro.sakura.ne.jp", {
```
bovine3dom: Does this play nicely with multiple windows?

src/lib/config.ts
```diff
@@ -773,6 +773,11 @@ class default_config {
      *
      */
     perfsamples: string = "10000"
+
+    /**
+     * Whether or not to integrate with the TST tab
```
bovine3dom: You should explain what sort of integration there is (i.e, tab changing) and also that nothing bad will happen if TST is not installed.

bovine3dom:
I think you told us this somewhere, but `gt/gT` don't loop around the tabs once you have reached the end.

src/excmds.ts
```diff
@@ -1706,11 +1706,66 @@ export async function tabprev(increment = 1) {
...
     // Kludge until https://bugzilla.mozilla.org/show_bug.cgi?id=1504775 is fixed:
-    return browser.tabs.query({ currentWindow: true, hidden: false }).then(tabs => {
-        tabs = tabs.sort((t1, t2) => t1.index - t2.index)
-        let prevTab = (tabs.findIndex(t => t.active) - increment + tabs.length) % tabs.length
-        return browser.tabs.update(tabs[prevTab].id, { active: true })
-    })
+
+    var hasTST = false
+    try {
+        // Not sure why this is an error
```
bovine3dom: I just pushed a commit that "fixes" this error. It was because the type checker didn't know about `management.get`.

bovine3dom:
I'm going to adopt this PR :)

PR #1301, opened 2019-01-24
Option to disable javascript hinting

Rahlir:
Allowing use of `-j` option that disables hints for javascript elements. This option can be combined with other options. This is particularly useful when one is using the `-c` option to generate hints only for specified elements. 

Should solve #1299

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ tags
 compiler/*.js
 compiler/**/*.js
 src/.metadata.generated.ts
+*.swp
--- a/.gitignore
+++ b/.gitignore
@@ -16,4 +16,3 @@ tags
 compiler/*.js
 compiler/**/*.js
 src/.metadata.generated.ts
-*.swp
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3495,6 +3495,11 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
         option = "-" + option.slice(2)
         rapid = true
     }
+    let jshints = true
+    if (option.startsWith("-j")) {
+        option = "-" + option.slice(2)
+        jshints = false
+    }
 
     let selectHints = new Promise(r => r())
     let hintTabOpen = async (href, active = !rapid) => {
@@ -3528,6 +3533,7 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
                     return link
                 },
                 rapid,
+                jshints,
             )
             break
 
@@ -3541,6 +3547,7 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
                     return elem
                 },
                 rapid,
+                jshints,
             )
             break
 
@@ -3598,6 +3605,7 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
                     return elem
                 },
                 rapid,
+                jshints,
             )
             break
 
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -16,3 +16,4 @@ tags
 compiler/*.js
 compiler/**/*.js
 src/.metadata.generated.ts
+*.swp
```
bovine3dom: I'm somewhat reluctant to add this as it is editor specific. Out of curiosity, why isn't it in your global gitignore?

bovine3dom:
Sorry for the drive-by pedantry. I'll look at this soon. I don't remember why I told glacambre to do it this way, so I probably ought to figure that out first before we change it...

.gitignore
Rahlir: You are probably right that this belongs to a global gitignore. That's just because I recently made switches back and forth from different editors and didn't bother changing it, simply added it where I needed.

src/excmds.ts
```diff
@@ -3446,6 +3446,7 @@ import * as hinting from "@src/content/hinting"
         - `-pipe selector key` e.g, `-pipe a href` returns the key. Only makes sense with `composite`, e.g, `composite hint -pipe * textContent | yank`. If you don't select a hint (i.e. press <Esc>), will return an empty string.
         - `-W excmd...` append hint href to excmd and execute, e.g, `hint -W exclaim mpv` to open YouTube videos.
         - -q* quick (or rapid) hints mode. Stay in hint mode until you press <Esc>, e.g. `:hint -qb` to open multiple hints in the background or `:hint -qW excmd` to execute excmd once for each hint. This will return an array containing all elements or the result of executed functions (e.g. `hint -qpipe a href` will return an array of links).
+        - -j* disable javascript hints. Don't generate hints related to javascript events. This is particularly useful when used with the `-c` option when you want to generate only hints for the specified css selectors. Also useful on sites with plenty of useless javascript elements such as google.com
         - -br deprecated, use `-qb` instead
```
bovine3dom: Please add -jc.

Also, I think it makes more sense that j would enable rather than disable JS hints.

I think the usual convention is that uppercase letters disable an option, so -J might make more sense?

bovine3dom: Let's have it include JS hints by default (this is the current behaviour, right?).

Your code seems to preclude, e.g. `-qjc`. Is that on purpose?

Rahlir: Ok, I will do that.

What do you mean preclude? -qjc works just as -qc minus the javascript hints. It doesn't prevent the use of other options or anything like that. Did you mean precede? Not sure what you are referring to there.

src/excmds.ts
```diff
@@ -3484,6 +3485,11 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
         option = "-" + option.slice(2)
         rapid = true
     }
+    let jshints = true
+    if (option.startsWith("-j")) {
```
bovine3dom: My reading of this code is that `-qjc` would trigger the rapid option but the jshints option; `-jqc` would trigger jshints but not rapid.

"preclude" is just a fancy way of saying "stop".

Rahlir: You are right, I didn't realize that we are not looping though the options as in bash, my bad, I will fix this.

Rahlir:
Ok, I think this should be all that was needed. Doc is updated with an example and we are using `-J` as an option to disable js hints. Default is still hints **with** javascript elements.

Rahlir:
Can this be merged now?

bovine3dom:
Thanks!

PR #1372, opened 2019-02-18
feat/minor: add hint that scrolls element to the top of a page

MannySchneck:
Hello all,

Thank you for all of your work on tridactyl!

I've added a hint that scrolls the selected element to the top of the viewport (operating on text elements).

This is something I've wanted for a while, since I have a hard time maintaining reading flow when scrolling the page ends up cutting off part of a paragraph. 

If this is an undesired feature or I've missed something, let me know!

Cheers,

Manny

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3497,6 +3497,7 @@ import * as hinting from "@src/content/hinting"
           - `bind ;c hint -c [class*="expand"],[class="togg"]` works particularly well on reddit and HN
         - -w open in new window
         - -wp open in new private window
+        - -z scroll an element to the top of the viewport
         - `-pipe selector key` e.g, `-pipe a href` returns the key. Only makes sense with `composite`, e.g, `composite hint -pipe * textContent | yank`. If you don't select a hint (i.e. press <Esc>), will return an empty string.
         - `-W excmd...` append hint href to excmd and execute, e.g, `hint -W exclaim mpv` to open YouTube videos.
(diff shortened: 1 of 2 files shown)


antonva:
It's missing documentation. If you add your feature to  the comment above the function declaration, TypeDoc should take care of the rest.

Other than that it looks good.

MannySchneck:
Thanks for the feedback!

I've updated the typedoc and I think I've made the correct change. Let me know if I missed something.

One question I had when I was adding this: I'm not sure if there's an ordering scheme for the hinting modes. Would you accept a (future) PR that re-ordered the doc-comments and the code so that the options are alphabetical?

bovine3dom:
Hmm, I'm not sure I see the benefit of alphabetical ordering.

I'd prefer them to be ordered in terms of level of usefulness - e.g, "open in background tab" near the top, "save image as" near the bottom. On which note - could you move `-z` down the list? Either at the bottom, or somewhere in the middle.

MannySchneck:
Done. I wonder if we could define some ordering, so that this conversation can be avoided in the future? =) It would also make it easier to keep the ordering in the docs/config consistent with the code.

Here's a first pass, where things higher up in the list sort first:
* Operations related to yanking
* Operations related to navigation (new tab, new window, follow link in this tab, etc...)
* Operations related to navigation within the page, (scrolling, focus element, etc...)
* Operations related to saving 
* 'Complex' operations, (run excmd on element, etc...)

Feel free to ignore the suggestion if it seems overly-specific. Happy to make a PR that does this.

bovine3dom:
I prefer:

* Operations related to navigation (new tab, new window, follow link in this tab, etc...)

* Operations related to navigation within the page, (scrolling, focus element, etc...)

* Operations related to yanking

* 'Complex' operations, (run excmd on element, etc...)

* Operations related to saving

I'd be happy to merge a PR that did it.

MannySchneck:
👍 Sounds good. That makes more sense to me too.

PR #1381, opened 2019-02-22
Added `hint -t`, `hint -tc` and `hint -Jtc`, to open links in a new tab.

nfvs:
Using `-w newtab` wouldn't allow us to use selectors with `-c`, and using composite would always open a new empty tab on cancel (esc).

This simplifies the google/HN hinting results example:
`bindurl www.google.com F hint -Jtc .rc>.r>a`

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3473,6 +3473,7 @@ import * as hinting from "@src/content/hinting"
 /** Hint a page.
 
     @param option
+        - -t open in a new foreground tab
         - -b open in background
         - -y copy (yank) link's target to clipboard
         - -p copy an element's text to the clipboard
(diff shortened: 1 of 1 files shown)


nfvs:
Also has the advantage of observing `tabopenpos`/`relatedopenpos` correctly, since the expected behavior is for new tabs from hinting to observe `relatedopenpos` (which `-b` already does), but without the `-t` flag we need to use `-W tabopen` which uses `tabopenpos`.

src/excmds.ts
```diff
@@ -3487,7 +3488,7 @@ import * as hinting from "@src/content/hinting"
         - -A save-as the linked image
         - -; focus an element
         - -# yank an element's anchor URL to clipboard
-        - -c [selector] hint links that match the css selector
+        - -c* [selector] hint links that match the css selector
```
bovine3dom: What does the star mean?

src/excmds.ts
```diff
@@ -3473,6 +3473,7 @@ import * as hinting from "@src/content/hinting"
 /** Hint a page.
 
     @param option
+        - -t open in a new tab
```
bovine3dom: We should probably say `open in a new foreground tab`

bovine3dom:
LGTM. Will merge once we fix doc issues.

nfvs:
Can't comment on the review apparently.

> What does the star mean?
I thought it meant it could be combined with other arguments, like -J and -c

src/excmds.ts
bovine3dom: I saw your reply by email but apparently can't see it here. I think it's obvious that the flag can be combined. I would interpret the * as being a literal part of the flag.

nfvs:
Done and squashed into a single commit.

bovine3dom:
Thanks!

PR #1489, opened 2019-04-21
Excmds in content redux

### LINKED ISSUE #727: autocmds can run in a different tab to the one they were triggered in

saulrh:
Fixes #727, mitigates #1236, should help with #1208 and probably many others. Obsoletes #995 and #1296. Thanks to #1488 for making some of this quite a bit cleaner.

The core principal:
* Teach `excmd_macros.py` to create shims for the background script that forward content-mode excmds via `messageActiveTab`.
* Generalize `controller_background.ts` into `lib/controller.ts` so that we can instantiate one for the content script and one for the background script. We choose to do this by adding a module-level `excmds` variable to it that we configure at the entry-point to the script.
* Generalize a few excmds so they can run in whichever context they're invoked in. For example, `composite` can now execute from both the content and background script.
* We now have all four of the `{controller,excmd}_{content,background}` message endpoints. `controller_*` has `acceptExCmd` for when you need to parse an exstr, while `excmd_*` has all of the excmds so you can invoke them with your own arguments.
* When `commandline_frame` dispatches an excmd, it now does uses `messageOwnTab`. The excmd can be executed directly in the content script if it's a content-mode excmd, guaranteeing that it will execute in the right tab. Background-mode aucmds (for example, `:tabprev`) will be sent to the background by the new shims.
* Similarly, `loadaucmds` dispatches directly to content-mode excmds, which are guaranteed to execute in the right tab if content-mode or forward to the background through shims if background-mode.

Downsides: `:repeat` is now... slightly funky. See the documentation for details.

Note that I have not in the general case actually verified that content-mode excmds are actually being executed without an RPC to the background. I've just written code that should make it work like that. I _have_ verified that aucmds are executed in the correct tab now, however, which is probably the most important part.

Additionally, there are a ton of excmds that are currently `//#background` that could (and therefore probably should) be `//#both`. For example, almost everything related to configs could be executed in the content script because `config.ts` does its own RPCs. I have skipped these for now since I don't want to clutter this PR up too much.

DO NOT MERGE until #1488 is committed, since I based this on top of that one. Once it's submitted I'll rebase on master to make this mergeable.


src/excmds.ts
```diff
@@ -2511,19 +2531,26 @@ async function getnexttabs(tabid: number, n?: number) {
...
+    If `cmd` doesn't exist, re-executes the last command.
     Executes the command once if `n` isn't defined either.
+
+    Note that "the last command" is a strange definition that depends
+    on some internal details of tridactyl. In general, `:repeat`
+    executed in the context of a tab will re-execute the last command
+    that was executed in that tab. However, many commands that touch
+    global state (for example, `:tabprev`) are also partially executed
+    in a global context, and if you manage to execute `:repeat` in
+    that global context it will repeat the last such global command.
```
bovine3dom: Could you add examples of when you might execute it in a global context? Is that the new tab page etc?

saulrh: ...Huh. Apparently that's harder to do than I expected. Updated doc.

Issue #810, opened 2018-07-15
Autocommand arguments


noctuid:
Pentadactyl has a bunch of arguments that it passes to autocommands. Most of them are for autocommands that don't exist in tridactyl and probably aren't worth implementing unless someone specifically requested them. The only ones I'd really like to have is `<url>` (potentially useful in general) and `<file>` (if #807 is implemented). You can reference these arguments in pentadactyl autocommands like this:
```
au DownloadPost * silent !dl_move "<file>"
```

Here is the full list of arguments for reference:
```
For Ex [cmd]s, the following keywords are replaced with the appropriate value
before the commands are executed. For JavaScript commands, they may be accessed
as ordinary variables, sans angle brackets.

<bookmark>
    The JavaScript bookmark object. Only for Bookmark*.
<changed>
    The name of the property that has changed. Only for BookmarkChange.
<doc>
    The document for which the event occurred. Only for DOMLoad, PageLoad and PageLoadPre.
<file>
    The target destination of a download. Only for DownloadPost.
<icon>
    The icon associated with <url>. Only for Bookmark*.
<keyword>
    The keywords applied to the bookmark. Only for BookmarkChange, BookmarkRemove.
<name>
    The name of the item. Only for ColorScheme and Sanitize.
<size>
    The size of a downloaded file. Only for DownloadPost.
<state>
    The new state. Only for Fullscreen.
<tab>
    The tab in which the event occurred. Only for DOMLoad, PageLoad and PageLoadPre.
<tags>
    The tags applied to <url>. Only for Bookmark*.
<title>
    The page, bookmark or download title.
<url>
    The URL against which the event was selected.
<win>
    The window for which the event occurred. Only for DOMLoad, PageLoad and PageLoadPre.
```

bovine3dom:
<url> can be had pretty easily with `js  document.location`.

PR #1495, opened 2019-04-23
[RFC] Aucmds  refactor and Downloadpost

antonva:
This addresses issues #810 and #807. An optional argument of the type
`AutocmdArgs` can be passed to `loadaucmds`. Only a small subset of the variables
mentioned in #810 are implemented. I think it's better to implement these as the
need/request arrives but for the `DownloadPost` aucmd I felt like this was the
most straightforward way to implement.

The `parseaucmdargs` function takes an aucmd and looks for the pattern "<param">
where param is any of the keys provided by `AutocmdArgs` and replaces the pattern
with the matching information from the args.

Example usage:
`:au DownloadPost .* fillcmdline "<file>"`
Will print the full path of the downloaded file on the cmdline.

I'm not attached to the naming convention but it's the same as pentadactyl used.

I have yet to write the documentation for the feature.

Any comments welcome.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1680,7 +1680,7 @@ if (fullscreenApiIsPrefixed) {
 
 /** @hidden */
 //#content
-export async function loadaucmds(cmdType: "DocStart" | "DocLoad" | "DocEnd" | "TabEnter" | "TabLeft" | "FullscreenEnter" | "FullscreenLeft" | "FullscreenChange") {
+export async function loadaucmds(cmdType: AUCMDS) {
     const aucmds = await config.getAsync("autocmds", cmdType)
     const ausites = Object.keys(aucmds)
     const aukeyarr = ausites.filter(e => window.document.location.href.search(e) >= 0)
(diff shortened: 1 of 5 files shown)


src/excmds.ts
```diff
@@ -1678,13 +1678,32 @@ if (fullscreenApiIsPrefixed) {
 }
 // }
 
+/** Expands all occurrences of "<.*>" to matching args.
+   @hidden
+ */
+
+//#content_helper
+export function parseaucmdargs(aucmd: string, args: AutocmdArgs): string {
+    const re = /^\"<(\w+)>\"$/
```
saulrh: How hard would it be to factor this out to into a well-contained library so you can add some simple test cases? It _looks_ right, and I expect that Pentadactyl collected enough real-world observations to establish that strings like `<file>` and such didn't occur that often in URLs, but I'd prefer to not start accidentally replacing random other `<....>` pairs.

antonva: I'm all for factoring it out into a library, the problem there would be the implementation. If we remove loadaucmds from excmds along with the parser I'm guessing it'd have to live in the content script somehow similar to what autocontainers does in the background.

The current implementation is pretty heavy handed so I'm nigh sure you won't get it mixed with urls or other things. I split the aucmd string on spaces and check for exact matches on `^"<WORD>"$`. The problem with this is that you can't combine params, `"<file> <url>"` will not be matched since the split on space would create `"<file>` and `<url>"` which won't match the regex.

bovine3dom:
@antonva any thoughts on #1885? They look fairly tangential, except WebRequest has its own filter syntax, so we might need to write some conversion between the two.

antonva:
@bovine3dom it doesn't behave all that different either if I remember correctly. 

I must admit I had to read over what I had  originally proposed here as it's been a while, but something very similar could be set up for webrequests.

PR #1520, opened 2019-05-05
Add `tri` omnibox keyword as another escape hatch

saulrh:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/omnibox

This allows the user to enter a string like `tri tabopen -c container www.google.com` to execute an exstr from the omnibox. It is _not_ a replacement for the tridactyl commandline; display and interactions are both extremely limited. However, it's probably a decent escape hatch for when you land on a page with the URL bar focused.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,7 +111,7 @@ You can bind your own shortcuts in normal mode with the `:bind` command. For exa
 ## WebExtension-related issues
 
 -   Navigation to any about:\* pages using `:open` requires the native messenger.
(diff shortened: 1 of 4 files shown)


bovine3dom:
Does this count as a permission, do you know?

saulrh:
I don't believe that this is a permission. Nothing relevant appears in the help page about `permissions`, and it can only be activated if the user enters the key so it doesn't feel like it'd have meaningful security or privacy implications.

PR #1527, opened 2019-05-06
Draft: Remove global state in content script

saulrh:
The goal: No code executed in module scope. As few variables as possible in module scope. This will have several major benefits.

First, initialization will much more tightly controlled. Init currently happens as modules are loaded, which means we have uncontrollable race conditions (due to `import`s being hoisted to the opt of each module where we can't synchronize them) and all sorts of fun debugging issues.

Second, it allows modules to be included without initialization code being executed. This will mean you can include your module as many times as you want and from wherever you want without, for example, adding infinitely-recursing commandline iframes because your operation in the commandline frame script wanted some module that eventually pulls in the content script which creates another commandline iframe.

Third, it means we regain encapsulation. Right now we have a ton of global variables being exported from various modules where other modules can read and write them freely. This is *bad*. This means that our modules are tightly coupled and global state can have far-reaching consequences that are hard to debug and even harder to refactor. This *needs* to change. One way to fix it is to tighten variable scopes and carefully design hierarchies and accessors, which is what I'm doing here. A robust IPC solution would be another method.

This is nowhere near complete, even for this one bit of code. I cleaned up a lot of the simple stuff in `content.ts` but the hard part remains: plumbing the content state (with the mode) down into the excmds. I'm not entirely certain about the best way to accomplish this, honestly. We can't even go with the common solution of enclosing `excmds.ts` in an object which encapsulates the global state, because getting that to play nice with `excmd_macros.py` would be a nightmare.

Anyway, the net result is that we can pull the actual logic in `content.ts` out into a module that does exactly nothing in its module scope. Other modules can pull this in as necessary without fear of messing things up.


saulrh:
Note that this structure for data substantially parallels the model arrived at by essentially all of the modern reactive UI frameworks, so integration should be pretty clean at the point of contact between script code and UI components.

PR #1534, opened 2019-05-08
Add arrows navigation in the hint mode

kurnevsky:
Arrows navigation is a convenient addition for the next/previous links choosing with `Tab` key.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -24,6 +24,15 @@ import * as config from "@src/lib/config"
 import Logger from "@src/lib/logging"
 const logger = new Logger("hinting")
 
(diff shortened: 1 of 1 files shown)


src/content/hinting.ts
```diff
@@ -87,6 +96,142 @@ class HintState {
         this.focusedHint = activeHints[nextFocusedIndex]
         this.focusedHint.focused = true
     }
+
+    changeFocusedHintTop() {
```
bovine3dom: There's a reasonable amount of redundancy in these changeFocusedHint* functions. If you fancied refactoring it so that the filter and reduction could be reused, I'd merge that PR :)

PR #1535, opened 2019-05-08
Fix tridactylrc location in documentation

hrubi:


hrubi:
I hit this when I was setting up tridactylrc. Hopefully it'll save somebody else a bit of time.

bovine3dom:
Ah, thanks. Could you document all of them as listed here?

https://github.com/tridactyl/tridactyl/blob/master/native/native_main.py#L95

Issue #534, opened 2018-05-14
Provide build without new tab page


oxysoft:
This is a big problem. I use momentum for my new tab page and Tridactyl overrides it. I know there's a setting in tridactyl to forward the page to something else, but it's not very fast and it means I can't start typing as soon as I have typed **CTRL-T** because the URL bar changes a second or two after.

bovine3dom:
Have you tried uninstalling Momentum and installing it after Tridactyl? Or just disabling/enabling it?

bovine3dom:
Duplicate of #384.

oxysoft:
Yes I have.

bovine3dom:
Magic, I guess. I think the only way of solving this is to offer an API to other add-ons to let Tridactyl run on their page, if that's even possible.

oxysoft:
Personally I don't care for Tridactyl to run on my new tab page, I just CTRL-T CTRL-L and start typing.

bovine3dom:
> A build of the extension without a new tab page override is offered.

You can do this yourself very easily by just removing this line: https://github.com/cmcaine/tridactyl/blob/master/src/manifest.json#L16

If you want a signed build, getting Mozilla API keys is pretty easy - just go to https://addons.mozilla.org/en-US/developers/addon/api/key/ while signed in with a Firefox account. Put these details in a file called `AMOKEYS` which looks like:
```
AMOKEY=youruser:id
AMOSECRET=abighashcodething
```

Then, you'll need to make a new addon listing on the Firefox site (it can be hidden) and change the ID everywhere to that.

Once this is all done, just run `scripts/sign stable` and you'll get an XPI in web-ext-artifacts/ that you can add to Firefox permanently.

I hope this helps :)

bovine3dom:
Thanks to @mphe and @cmcaine this now exists at https://tridactyl.cmcaine.co.uk/betas/nonewtab/.

It's a bit broken - see https://github.com/tridactyl/tridactyl/labels/nonewtab - but I'm closing this issue because it is "fixed" in spirit.

Anecdotally - I've been using the build for about 10 minutes, and it's _extraordinarily_ annoying to get stuck on a new tab page where Tridactyl isn't running, so I feel vindicated in my ambivalence towards dedicating any time to getting this working ;)

oxysoft:
[...]

PR #1539, opened 2019-05-09
Provide build without new tab page

mphe:
See #534 and #384.
If I forgot something, please let me know.

### DIFF
--- a/scripts/sign
+++ b/scripts/sign
@@ -4,7 +4,7 @@ set -e
 
 sign_and_submit() {
     # Don't trust the return value of web-ext sign.
-    (source AMOKEYS && (web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET || true))
+    (source AMOKEYS && (web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET "$@" || true))
 }
 
 publish_beta_nonewtab() {
@@ -13,7 +13,7 @@ publish_beta_nonewtab() {
     scripts/version.js beta
     sed 's/tridactyl.vim@cmcaine/tridactyl.vim.betas.nonewtab@cmcaine/' -i build/manifest.json
     sed '/\s*"newtab":.*/d' -i build/manifest.json
-    sign_and_submit
+    sign_and_submit -a web-ext-artifacts/nonewtab
 }
 
 publish_beta() {
--- a/scripts/sign
+++ b/scripts/sign
@@ -12,7 +12,7 @@ publish_beta_nonewtab() {
     npm run build
     scripts/version.js beta
     sed 's/tridactyl.vim@cmcaine/tridactyl.vim.betas.nonewtab@cmcaine/' -i build/manifest.json
-    sed '/"newtab": "static\/newtab.html"/d' -i build/manifest.json
+    sed '/\s*"newtab":.*/d' -i build/manifest.json
     sign_and_submit
 }
 
--- a/scripts/sign
+++ b/scripts/sign
@@ -7,6 +7,15 @@ sign_and_submit() {
     (source AMOKEYS && (web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET || true))
 }
 
+publish_beta_nonewtab() {
+    npm run clean
+    npm run build
+    scripts/version.js beta
+    sed 's/tridactyl.vim@cmcaine/tridactyl.vim.betas.nonewtab@cmcaine/' -i build/manifest.json
+    sed '/"newtab": "static\/newtab.html"/d' -i build/manifest.json
+    sign_and_submit
+}
+
 publish_beta() {
     npm run clean
     npm run build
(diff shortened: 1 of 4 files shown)


scripts/sign
```diff
@@ -7,6 +7,15 @@ sign_and_submit() {
     (source AMOKEYS && (web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET || true))
 }
 
+publish_beta_nonewtab() {
+    npm run clean
+    npm run build
+    scripts/version.js beta
+    sed 's/tridactyl.vim@cmcaine/tridactyl.vim.betas.nonewtab@cmcaine/' -i build/manifest.json
+    sed '/"newtab": "static\/newtab.html"/d' -i build/manifest.json
```
bovine3dom: I'd make this a little more robust - [any whitespace]*newtab:*/d for example

bovine3dom:
I need to register this add-on with Mozilla, and add it to our build process (probably in a nonewtab/ folder on the webserver).

Could you look at how the beta is updated and tell the nonewtab to look in the same path but in a nonewtab/ folder? I think it's called update.json.

mphe:
Do you mean adapting the URL in this line: https://github.com/tridactyl/tridactyl/blob/master/scripts/version.js#L74 to "https://tridactyl.cmcaine.co.uk/betas/nonewtab/updates.json"?
What about the URL in line 78 below, does this need to be adapted, too?

scripts/sign
mphe: Fixed it

bovine3dom:
Yes, those lines exactly. I think :)

mphe:
I assume https://github.com/tridactyl/tridactyl/blob/master/scripts/version.js#L23 needs to be adapted, too. Should I just add another entry for "tridactyl.vim.betas.nonewtab@cmcaine.co.uk" or provide a function argument to specifiy another id/path?

bovine3dom:
Oh, that suggests we can use the one updates json for both add-ons, in which case the /nonewtab/updates.json is not needed, so just add another key like you suggest, with the /nonewtab/ path

mphe:
Done

scripts/sign
```diff
@@ -7,6 +7,15 @@ sign_and_submit() {
     (source AMOKEYS && (web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET || true))
```
bovine3dom: `web-ext sign -s build --api-key $AMOKEY --api-secret $AMOSECRET` needs to have `-a web-ext-artifacts/nonewtab` added to it for the newtab build otherwise the newtab vs nonewtab builds will overwrite each other :)

mphe: Fixed.
Wouldn't the same problem appear with beta and stable? Or does stable have a different output name?

bovine3dom: I guess it does? I've never noticed any problems. I'm not sure it outputs an xpi locally because the stable build is distributed on the AMO and is automatically uploaded in that step. 🤷‍♂️

bovine3dom:
Thanks, this is just waiting for cmcaine and me to register the ID.

If you want, you could add documentation to the new tab page where it says how to disable it to say that builds without a new tab exist. Extra brownie points if you write a small guide on migrating configuration between builds :)

mphe:
Added a note about the nonewtab build in readme.md and newtab.md.

> Extra brownie points if you write a small guide on migrating configuration between builds :)

No idea about that :)
Aren't all configs located in `.tridactylrc` anyway? What exactly needs to be migrated?

bovine3dom:
> Aren't all configs located in `.tridactylrc` anyway? What exactly needs to be migrated?

All `set` commands you run in Tridactyl are stored in the Firefox Sync storage by default (which exists even if you don't have a Sync account, it just stays local to the profile). An RC file is just a file which is read and executed line-by-line every time you start Firefox. If you don't have "sanitise ..." at the top as the exemplar suggests your RC file and Firefox Sync storage may diverge (and even then they could potentially diverge depending on how many machines you have fighting).

I don't have numbers but I suspect most Tridactyl users don't install the native messenger and so don't use the RC files.

So migration would involve extracting that stuff. See https://github.com/tridactyl/tridactyl/issues/79#issuecomment-351132451 :)

mphe:
Ah, thanks for the explanation.
I always thought it was more common to have a `.tridactylrc` to have all settings in one easily readable and modifiable place.

Don't think I will write a guide for that, but I could add a link to that comment to the readme or new tab page.

cmcaine:
I had to make a few minor changes to get this all to work. See commits 64160b8d9672c4f18cb2f0239a7db801183565f9  and f1f10567f72a436b3c2997509a67eb58927bf0b3

It's up now and auto-updates work.

cmcaine:
Thanks for your contribution @mphe  :)

mphe:
Thanks for merging this :)

I noticed the URL change in https://github.com/tridactyl/tridactyl/commit/f1f10567f72a436b3c2997509a67eb58927bf0b3 seems to break the download URLs in readme.md and newtab.md.

bovine3dom:
Fixed. Thanks!

mphe:
In `native/tridactyl.json` the new ID must be added, too, otherwise the native messenger won't work.

PR #1580, opened 2019-05-23
Text filtered hints

vlmarek:
Hi,

So this is where I stand now. I'm happy with the code, but I gave it just a little testing so far. Since everything is new for me I don't expect to have very polished code.

I still miss regular expressions and searching on a whole page, not just in the visible part. I wold like to implement find patters similar to what pentadactyl had

http://bug.5digits.org/help/pentadactyl/pattern.xhtm

\c    Perform case insensitive find
\C    Perform case sensitive find
\L    Search the entire page.
\r    Process the entire pattern as a regular expression.
\R    Process the entire pattern as an ordinary string.

Thank you for everyone's support

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -705,7 +705,10 @@ export function hintableImages() {
     return DOM.getElemsBySelector(DOM.HINTTAGS_img_selectors, [DOM.isVisible])
 }
 
-export function hintByText(match) {
+/** Get array of selectable elements that display a text matching either plain
+ * text or RegExp rule
+ */
+export function hintByText(match: string|RegExp) {
     return DOM.getElemsBySelector(DOM.HINTTAGS_filter_by_text_selectors, [
         DOM.isVisible,
         hint => {
@@ -716,8 +719,12 @@ export function hintByText(match) {
             } else {
                 text = hint.textContent
             }
-            return text.indexOf(match) >= 0
-        },
+            if (match instanceof RegExp) {
+                return text.match(match) !== null
+            } else {
+                return text.includes(match)
+            }
+        }
     ])
 }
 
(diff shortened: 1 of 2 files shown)


vlmarek:
Hi,

I have updated the changes to be lint clean. I had to rewrite the hintPage function even more, but I believe that it is better now

src/content/hinting.ts
```diff
@@ -716,7 +716,11 @@ export function hintByText(match) {
             } else {
                 text = hint.textContent
             }
-            return text.indexOf(match) >= 0
+            if (match.constructor.name == "RegExp") {
+                return text.match(match) !== null
+            } else
+                return text.indexOf(match) >= 0
+            }
```
cmcaine: use `text.search(match) >= 0` instead. It works for both regexp and string patterns.

cmcaine:
You probably noticed that I already merged your older hint -f stuff. It didn't appear here on github because I had to rebase. I'll merge once the issue above is fixed (or I'll fix it later and merge anyway ;)

src/content/hinting.ts
vlmarek: I have tried that and it seems that it always interprets the text as a regex. That is
`:hint -f $`
matches all links, not only those containing '$'. This seems to confirm it: https://www.w3schools.com/jsref/jsref_match.asp

cmcaine: Sorry, my bad there.

Please use `match instanceof RegExp` for the check and `text.includes(match)` for the string match.

Happy to merge then.

vlmarek: Thanks for the tips, it's the first time I'm touching typescript. I have also added some comment above the function.

bovine3dom:
I made `hint -f` case insensitive because it surprised me when it wasn't. Merged. Thanks!

PR #1585, opened 2019-05-25
native.ts: fix getProfileUncached when loading fallback default profile

thechunk:
On my setup, I was getting an error on `:guiset` where the profile would never get discovered. I found that the `iniObject[profileName]` object had a `Default` property as Number type so it was always failing. This seems to fix that. Let me know if anything needs to be changed, thanks!


thechunk:
Thanks & appreciate your work so I could feel at home on my FF again 😄

PR #1593, opened 2019-05-27
browser.extension.getURL -> browser.runtime.getURL

saulrh:
Function is deprecated:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getURL

Replacement has essentially identical functionality:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getURL

I'm sure that there's _some_ difference between the "path in install directory" and "path relative to manifest.json", but I don't think that we're going to be the ones running into it.

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -15,7 +15,7 @@ import { enumerate } from "@src/lib/itertools"
 import { toNumber } from "@src/lib/convert"
 import * as aliases from "@src/lib/aliases"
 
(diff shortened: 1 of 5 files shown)


saulrh:
Alternatively, we could completely replace use of this function with https://github.com/webpack-contrib/file-loader

PR #1602, opened 2019-05-27
Fix #1156 - if Firefox ever implements these APIs

### LINKED ISSUE #1156: installing beta does not over-write earlier stable install, and both are enabled by default
After I installed the latest beta, I had two Tridactyl plugins enabled.  Here's a screenshot showing the state after I disabled the old stable one.  There doesn't seem to be a broad problem with having two Tridactyls enabled, but I did notice a specific problem: until I manually disabled the old add-on, the recently fixed bug from #1031 was still showing up for me.

screen shot 2018-11-07 at 11 06 13

bovine3dom:
Don't merge this unless/until Firefox implements the API used.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -215,3 +215,19 @@ window.tri = Object.assign(window.tri || Object.create(null), {
 omnibox.init()
 
 // }}}
+//
+//
+
+// Disable all other versions of Tridactyl
+browser.management.getSelf().then(extinfo => {
+    browser.management.getAll().then(theminfos => {
+        // Should consider adding Vim Vixen, Surfingkeys, etc to this list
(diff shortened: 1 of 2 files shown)


src/background.ts
```diff
@@ -215,3 +215,19 @@ window.tri = Object.assign(window.tri || Object.create(null), {
 omnibox.init()
 
 // }}}
+//
```
bovine3dom:
```suggestion
```

src/background.ts
```diff
@@ -215,3 +215,19 @@ window.tri = Object.assign(window.tri || Object.create(null), {
...
+//
```
bovine3dom:
```suggestion
```

PR #1603, opened 2019-05-27
Add type checking to first arg of config.get

bovine3dom:
Also document noiframeon, which I discovered was undocumented
because of the type checking.

---

It makes splatted config.gets rather uglier with `...args` ->
`args[0] as keyof config.default_config, args.slice(1)...`.

It has apparently also broken config completions.


bovine3dom:
I'll make a helper function for dynamic usage that hides the `as default config, ...slice(1)`  ugliness.

PR #1644, opened 2019-06-01
Mktridactylrc

### LINKED ISSUE #695: mktridactylrc
We want to convert the user's current settings into a tridactylrc file; i.e, a series of ex-commands.

- should check to see whether a tridactylrc already exists and only allow overwriting it if a ! is appended to the command.
- `mktridactylrc /path/to/tridactylrc` should also be supported
- pipe support for e.g, the clipboard could be nice

If we restrict ourselves to just replicating the non-default settings, then this should not be too difficult.

I think it's fine to have `guiset` and `fixamo` be out of scope at least initially.

Bonus: in the grand tradition of Pentadactyl, we should ensure that `mkv` and `mkp` are bound to some unexpected commands; `mkv` for installing tridactylrc ftdetect etc is fine, and perhaps `mkp` could be "make print preview"?

antonva:
- Subconfigs are now parsed into `js tri.config.set(...)` blobs and will stay that way.
- Force flag is adhered to.

Closes #695

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -501,6 +501,9 @@ export class default_config {
         zo: "zoom",
         installnative: "nativeinstall",
         nativeupdate: "updatenative",
+        mkt: "mktridactylrc",
+        "mkt!": "mktridactylrc -f",
+        "mktridactylrc!": "mktridactylrc -f"
     }
 
     /**
@@ -1367,7 +1370,7 @@ export function removeChangeListener<P extends keyof default_config>(
 }
 
 /** Parse the config into a string representation of a .tridactylrc config file.
-    Tries to parse the config into sectionable chunks based on keywords. 
+    Tries to parse the config into sectionable chunks based on keywords.
     Binds, aliases, autocmds and logging settings each have their own section while the rest are dumped into "General Settings".
 
     @returns string The parsed config file.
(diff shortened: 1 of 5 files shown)


src/lib/config.ts
```diff
@@ -989,7 +992,7 @@ export function getURL(url: string, target: string[]) {
                     k =>
                         url.match(k) &&
                         getDeepProperty(conf.subconfigs[k], target) !==
-                            undefined,
+                        undefined,
```
bovine3dom: Should probably just run prettier on this file to fix all these indentation issues.

src/lib/config.ts
```diff
@@ -1366,6 +1369,129 @@ export function removeChangeListener<P extends keyof default_config>(
...
+                    // Map the int values in e to a log level
+                    let level
+                    if (pconf[i][e] === 0) level = "never"
+                    if (pconf[i][e] === 1) level = "error"
+                    if (pconf[i][e] === 2) level = "warning"
+                    if (pconf[i][e] === 3) level = "info"
+                    if (pconf[i][e] === 4) level = "debug"
+                    parseobj.logging.push(`set logging.${e} ${level}`)
+                } else {
+                    parseobj.conf.push(`set ${i}.${e} ${pconf[i][e]}`)
```
bovine3dom: Surely this line should be identical to the subconfigs one for future proofing?

src/lib/config.ts
antonva: Will do.

src/lib/config.ts
antonva: Yeah you're right. Totally forgot to update it.

antonva:
I forgot to write documentation for the excmd. I'll do that in the morning.

PR #1657, opened 2019-06-04
Persist history completion selection if it is still valid on source update

cmcaine:


src/completions/History.ts
```diff
@@ -76,22 +76,29 @@ export class HistoryCompletionSource extends Completions.CompletionSourceFuse {
...
+        this.deselect()
 
-    updateChain() {
-        // Options are pre-trimmed to the right length.
-        this.options.forEach(option => (option.state = "normal"))
+        // Set initial state to normal, unless the option was selected a moment
+        // ago, then reselect it so that users don't lose their selections.
+        this.options.forEach(option => {
+            if (lastFocused !== undefined && lastFocused.value === option.value) {
+                this.select(option)
```
bovine3dom: Can we stick a `break` here to ensure we can only ever have one selected?

src/completions/History.ts
```diff
@@ -76,22 +76,29 @@ export class HistoryCompletionSource extends Completions.CompletionSourceFuse {
...
+            } else
+                option.state = "normal"
```
bovine3dom: Obviously we'd need to do a first pass with this on everything

src/completions/History.ts
```diff
@@ -87,12 +87,10 @@ export class HistoryCompletionSource extends Completions.CompletionSourceFuse {
...
-        this.options.forEach(option => {
-            if (lastFocused !== undefined && lastFocused.value === option.value) {
-                this.select(option)
-            } else
-                option.state = "normal"
-        })
+        this.options.forEach(option => option.state = "normal")
+        for (const option of this.options) {
+            if (lastFocused !== undefined && lastFocused.value === option.value) this.select(option)
+        }
```
cmcaine: You probably meant to put a break in there ;)

As it is, I don't think you actually changed the functionality.

bovine3dom: Wow

PR #1698, opened 2019-06-23
Fix editor and rcfile encoding errors

joaomsa:
Fixes encoding mangling happening within the rc file and editorcmd


bovine3dom:
Sorry for the delay in reviewing this. I'll try to get round to it tomorrow, or the end of the week at the latest.

Issue #1633, opened 2019-05-30
Config file tridactylrc only partially loaded


madduck:
Tridactyl version: 1.14.10-186-gae6deb0
Firefox version: Mozilla Firefox 66.0.4
Operating system: linux

The configuration file included below is only partially loaded when starting Firefox. It gets fully loaded when I run `source` manually. This seems related to #1409 / #1197, but it's a separate issue I think.

For a start, I use `storageloc==sync`, i.e. the default. Here is the [full config](https://scratch.madduck.net/tridactyl_config.json), and this is `tradactylrc`:

```
sanitize tridactylsync tridactyllocal

set editorcmd x-terminal-emulator -name tridactyl-edit -e vim -c "autocmd BufWritePost <buffer> exe 'write ' . fnameescape((exists('$TMPDIR') ? expand('$TMPDIR') : '/tmp') . '/tridactyl.backup.' . strftime('%Y-%m-%d-%H%M%S'))" +start

set searchengine duckduckgo
set searchurls.pad https://pad.riseup.net/p/%s

set theme dark

set modeindicator false

" autocmd TabEnter .* unfocus

bind <C-1> tabnext_gt 1
bind <C-2> tabnext_gt 2
bind <C-3> tabnext_gt 3
bind <C-4> tabnext_gt 4
bind <C-5> tabnext_gt 5
bind <C-6> tabnext_gt 6
bind <C-7> tabnext_gt 7
bind <C-8> tabnext_gt 8
bind <C-9> tabnext_gt 9

" #639 will fix this
seturl in.xero.com noiframe true
seturl trademe.co.nz noiframe true
```

This only gets read to the first comment, i.e. the `bind` and `seturl` calls won't have been made. Removing the comment has no effect.

madduck:
@glacambre Yes, that seems to fix it. At least in 3 starts, I didn't see the problem again. But boy does FF now take forever to load. ;)

madduck:
@glacambre I've had the sleep in there for a while now, but all of a sudden, the same problem has come back. So no, not fixed.

bovine3dom:
Closing as a duplicate of #1409.

PR #1764, opened 2019-07-30
:source race conditions

arcnmx:
There have been [a](https://github.com/tridactyl/tridactyl/pull/724) [number](https://github.com/tridactyl/tridactyl/pull/686) [of](https://github.com/tridactyl/tridactyl/pull/1197) [issues](https://github.com/tridactyl/tridactyl/pull/1409) [related](https://github.com/tridactyl/tridactyl/pull/1633) to racy configuration settings and `:source`, `tridactylrc`, etc. This is an attempt to address them. Note this isn't related to the `storageloc` setting though I imagine the `sync` and `local` storage backends may differ in how they're affected due to when and how they call `onChanged`.

First, a description of the problem by outlining the steps which occur when running `:set` multiple times:

1. `config.set("key", value)` updates values in a global `USERCONFIG` object
2. `config.save()` implicitly gets called to save this object into browser storage (local or sync depending on what `storageloc` is set to)
3. The promise returned by save/set resolves with the call to `storage.sync.set()`, and after an await tridactyl continues on to your next command
4. `config.set("anotherkey", value)` repeats the above 3 steps to update `USERCONFIG` and browser storage. The promise resolves again, no problems so far?
5. The browser storage uses the `onChanged` callback to indicate that "key" has been updated, and then this callback incorrectly assigns the `newValue` into the `USERCONFIG` global.
   - At this point, the `anotherkey` setting has been reverted.
6. `config.set("yetanotherkey", value)` repeats the steps again, combining it with your `key` setting, but having lost the `anotherkey` change.
7. Maybe `onChanged` gets called again at this point and brings `anotherkey` back but erases `yetanotherkey` with it, it just all results in an [inconsistent leapfrog](https://github.com/tridactyl/tridactyl/pull/1409) I guess.

This basically boils down to:
- #1290 mitigated this using promises but that doesn't really fix the problem because `onChanged` is [not guaranteed to be run before the `set` promise completes](https://bugzilla.mozilla.org/show_bug.cgi?id=1554088). The current implementation makes incorrect assumptions, but it certainly was a lot worse before when everything got set at once though!
- Trying to maintain global state/cache that matches async storage is messy and inherently racy.
   - Since set() updates the global immediately, a way to tell (and ignore) which `onChanged` are coming from us and which are coming from remote contexts/browsers could help further mitigate the problem, but isn't a full solution.
- onChanged probably should be doing a recursive/deep intersection across the changed settings
- async everywhere on all get/set operations would probably be the full/proper solution in the end, but that's a huge change?

So the commits go into detail on the mitigations I'm adding here, but basically:
- only update settings that have actually changed in `onChanged` (partial fix, not recursive/deep like it ideally should be)
  - this mitigates clobbering when doing multiple sets in a row on distinct settings
  - unfortunately nested objects are still susceptible (like say, any of the `nmap`/etc keybindings)
    - so doesn't fix things `bind`
- check `oldValue` that the browser storage passes to `onChanged` rather than the global `USERCONFIG` state to determine what has changed
  - this enhances the above change to make things less racy
- `get()` the configuration from browser storage before setting any new values
  - this basically adds synchronization between sets to prevent the clobbering race

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1106,6 +1106,18 @@ export async function getAsync(
     ...target: string[]
 ) {
     if (INITIALISED) {
+        // TODO: consider storing keys directly
+        let browserconfig
+        switch (get("storageloc")) {
+            case "local":
+                browserconfig = await browser.storage.local.get(CONFIGNAME)
+                break
+            case "sync":
+                browserconfig = await browser.storage.sync.get(CONFIGNAME)
+                break
+        }
+        USERCONFIG = browserconfig[CONFIGNAME]
+
         return get(target_typed, ...target)
     } else {
         return new Promise(resolve =>
@@ -1129,7 +1141,7 @@ export function setURL(pattern, ...args) {
 
     @hidden
  */
-export function set(...args) {
+export async function set(...args) {
     if (args.length < 2) {
         throw "You must provide at least two arguments!"
     }
@@ -1137,8 +1149,16 @@ export function set(...args) {
     const target = args.slice(0, args.length - 1)
     const value = args[args.length - 1]
 
-    setDeepProperty(USERCONFIG, value, target)
-    return save()
+    if (INITIALISED) {
+        // wait for storage to settle, otherwise we could clobber a previous incomplete set()
+        await getAsyncDynamic(...target)
+
+        setDeepProperty(USERCONFIG, value, target)
+
+        return save()
+    } else {
+        setDeepProperty(USERCONFIG, value, target)
+    }
 }
 
 /** @hidden
@@ -1354,6 +1374,7 @@ async function init() {
     schlepp(localConfig[CONFIGNAME])
 
     await update()
+    await save()
     INITIALISED = true
     for (const waiter of WAITERS) {
         waiter()
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1595,7 +1595,8 @@ browser.storage.onChanged.addListener((changes, areaname) => {
             )
 
             // TODO: this should be a deep comparison but this is better than nothing
-            changedKeys.concat(unsetKeys).forEach(key => USERCONFIG[key] = newValue[key])
+            changedKeys.forEach(key => USERCONFIG[key] = newValue[key])
+            unsetKeys.forEach(key => delete USERCONFIG[key])
 
             // Trigger listeners
             unsetKeys.forEach(key => triggerChangeListeners(key, DEFAULTS[key]))
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1541,10 +1541,10 @@ const parseConfigHelper = (pconf, parseobj) => {
 // TODO: BUG! Sync and local storage are merged at startup, but not by this thing.
 browser.storage.onChanged.addListener((changes, areaname) => {
     if (CONFIGNAME in changes) {
-        const { newValue } = changes[CONFIGNAME]
-        const old = USERCONFIG
+        const { newValue, oldValue } = changes[CONFIGNAME]
+        const old = oldValue || {}
 
-        function triggerChangeListeners(key, value = USERCONFIG[key]) {
+        function triggerChangeListeners(key, value = newValue[key]) {
             const arr = changeListeners.get(key)
             if (arr) {
                 const v = old[key] === undefined ? DEFAULTS[key] : old[key]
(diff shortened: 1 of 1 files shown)


bovine3dom:
Hey, thought I should just let you know I haven't forgotten about this. I realistically won't get round to reviewing these two PRs till mid-October.

bovine3dom:
image

The code for this looks OK but `configversion` claims to be undefined in the completions. `get configversion` returns `0.0` until I open two tabs?

I don't get that behaviour on master. Has this PR slowed `set` down considerably?

arcnmx:
> I don't get that behaviour on master. Has this PR slowed set down considerably?

Specifically in the case of `configversion`, it's set as part of `config.init()` inside `config.update()`. The main changes there are:
1. all use of `set()` during init happens immediately to the local USERCONFIG context, so if anything it should be sped up?
2. The introduction of [`save()`](https://github.com/tridactyl/tridactyl/pull/1764/commits/5464e377328cdcff18da1b6683babb8e16615239#diff-3df87e205f06d43c78aaeb7c68d01490R1367) then makes the `WAITERS` wait for storage to sync before `getAsync()` results are given back.

So not really sure, I'd expect a `get()` or `getAsync()` to produce valid results, unless it's caching things before init runs? Or maybe the fact that every context tries to `init()` things simultaneously makes things vulnerable to weird race conditions, hm...

bovine3dom:
> Or maybe the fact that every context tries to init() things simultaneously makes things vulnerable to weird race conditions, hm...

I think this is a perennial problem. We should probably introduce some kind of global named lock mechanism.

I really want to make sure we understand this PR totally before it is merged as it's such a key part of Tridactyl. I know we don't really know how the current config works, but it has the rather large advantage that we know that it does work lots of the time.

When I manually `source` the exemplar .tridactylrc, lots of the settings briefly show up as undefined in the `get` completion, so I'm not sure it's something unique to the config init.

I'm back to being busy again so I probably won't look at this again seriously for a couple of weeks.

arcnmx:
> I think this is a perennial problem. We should probably introduce some kind of global named lock mechanism.

While it isn't great, this PR actually aims to improve the situation there by not triggering storage syncs until every setting has been updated, then merging only the ones that have changed. It's a bit awkward that init is run each time, but also it should be harmless as long as it's idempotent (and it not being async helps with that).

> I really want to make sure we understand this PR totally before it is merged as it's such a key part of Tridactyl.
> When I manually source the exemplar .tridactylrc, lots of the settings briefly show up as undefined in the get completion, so I'm not sure it's something unique to the config init.

Yeah, this behaviour certainly should be looked into - it sounds like a bug in how I'm handling clearing settings or defaults or something in onChanged - afair `undefined` shouldn't ever be observable in anything you get from `get()`. I just haven't touched this since I wrote it so am not as familiar anymore with how all the pieces interact, so it's hard to give any immediate explanation for it or where the issue might lie.

> I know we don't really know how the current config works

IIRC the main points to why it's not currently consistent or predictable are fairly straightforward:

1. Mixing async with racy calls, and calling async functions from non-async contexts. Maybe I should look into at least changing as many config calls as I can find to async ones - at least can fix the ones that are currently already in async functions.
    - If there are performance concerns/issues with using async everywhere, then we'll maybe have to batch calls around a timer a bit and try to ensure things follow a sort of [eventual consistency](https://en.wikipedia.org/wiki/Consistency_model#Eventual_consistency) model?
2. Storing the whole object in browser storage, then not merging changes to it. Storing the individual keys/settings instead would at least let the browser do this merging for you, since the storage api is already supposed to provide a distributed database!

> but it has the rather large advantage that we know that it does work lots of the time.

I somewhat disagree considering that it not working is half the reason I avoid using firefox :p - the other half of course being webext limitations :(

> I'm back to being busy again so I probably won't look at this again seriously for a couple of weeks.

Pretty busy now myself but hopefully I'll get around to looking at it again in more detail soon.

arcnmx:
> The code for this looks OK but configversion claims to be undefined in the completions. get configversion returns 0.0 until I open two tabs?

Okay so this was the result of a wrong assumption made by me about how `get()` works:

```javascript
USERCONFIG = await browser.storage.local.get(CONFIGNAME)
```
is incorrect, it returns a wrapper object so needs to be more like:
```javascript
USERCONFIG = (await browser.storage.local.get(CONFIGNAME))[CONFIGNAME]
```

bovine3dom:
There's still some something a little odd going on here:

```
:set editorcmd blah
# navigate to another page
:unset editorcmd
:get editorcmd # completion displays it as undefined but when I press Enter it returns "auto"
```

I'd like us to understand why that happens in case it is a symptom of a more serious bug. I can't replicate it on master so it is unique to this branch.

arcnmx:
Sorry, I didn't look close enough to fully fix the original issue you found because there were actually 2 distinct problems there:
1. the above explains why you were getting `0.0` instead of `1.7`, and fixes that
2. however, I overlooked the fact that it did not explain why the completion was showing `undefined` instead of `0.0`. This is because completions do the following:
    1. `const settings = config.get()` to retrieve the entire object rather than individual keys
    2. Then iterates over the keys and uses custom fallback logic to retrieve the defaults for any missing keys

The problem here is that `config.ts` treats an `undefined` value (even if present) as unset, while `completions/Settings.ts` requires the key to be missing from the object to be considered unset. I'll change it to always delete the key from the object, I just wasn't aware there was other code making that assumption.

PR #1831, opened 2019-09-04
Fix instruction to install the native client on Arch

nesmyslny:
Hey!

There is no official Arch package of the native client. At least I'm only aware of the AUR package: https://aur.archlinux.org/packages/firefox-tridactyl-native/
So people can install this with their preferred installing method (there are different AUR helpers, like `yay`). I kept it general as I think people using Arch know what to do.

To make it easier to review (Github doesn't show the exact diff inlines), that's the changed sentence:

```diff
- Arch users can run `pacman -S firefox-tridactyl-native` instead.
+ Arch users can install the [AUR package](https://aur.archlinux.org/packages/firefox-tridactyl-native/) `firefox-tridactyl-native` instead.
```


bovine3dom:
Ah, good catch. Thanks!

Issue #377, opened 2018-04-11
Tree Style Tab interop


bovine3dom:
TST has a nice sane API: https://github.com/piroor/treestyletab/wiki/API-for-other-addons

Does anyone use TST and Tridactyl? Is there anything you'd like to control via Tridactyl?

bovine3dom:
I see two obvious things we could do:

- show/hide TST via a Tridactyl command/bind (this would need https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/sidebarAction/close + open adding to TST API)
- using https://github.com/piroor/treestyletab/wiki/API-for-other-addons#apis-to-get-tree-information to display `:buffer ` as a tree, which we should probably bear in mind during the completions rewrite.

timss:
Neat. Looks one can do a lot of useful stuff with the new TST API.

I've been using TST for a long time, as well as with Vimperator. There's a couple of things I'm missing currently (for reference here's my old [vimperatorrc](https://github.com/timss/dotfiles/blob/master/.vimperatorrc)):

**Toggle subtree collapse:**  
https://github.com/piroor/treestyletab/wiki/API-for-other-addons#collapse-expanded-tree

**Promote/demote tabs:**  
https://github.com/piroor/treestyletab/wiki/API-for-other-addons#indent-demote-tab

vebyast:
I use TST and tridactyl; these two being functional and performant again are the reason I switched back to FF from chrome.

Several features that I'd find useful:
- `:buffer` showing some tree information, as suggested.
- A syntax for `:tabmove` that selects a tab under which to re-parent the current tab.
- A way to create the empty parent tabs that TST uses for organization; you can see these if you create several children from a pinned tab.

saulrh:
Other useful features:

- Commands for `:buffer` that permit navigation inside a tree - might try treating the tree like directories and looking at how helm does things?
- A command for closing child tabs along with the targeted tab.

rootkit-ty:
It would also be useful to have the ability to rely on TST's order for next and previous tab commands.
Since that the order of tabs used by firefox will differ from that displayed by TST.

ELLIOTTCABLE:
@rootkit-ty I'd think that'd … just work? Doesn't TST inject JS to ‘fix’ tab-navigation so it follows the tree structure and skips collapsed nodes?

[...]

PR #1838, opened 2019-09-09
Add recontain excmd

robertgzr:
Allows putting the current tab into (or out of) a container.

This has been one of the last things I always need to reach to the mouse for... 

The other being [simple tab groups](https://github.com/Drive4ik/simple-tab-groups) which I would give a shot at integrating whenever #377 makes it in (tridactyl is already whitelisted to use it's api: https://github.com/Drive4ik/simple-tab-groups/issues/406)

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2567,6 +2567,38 @@ export async function viewcontainers() {
             .replace(/ /g, "%20")
 }
 
+/** Opens the current tab in another container.
+
+    This is probably not a good idea if you care about tracking protection!
+    Transfering URLs from one container to another allows websites to track
+    you across those containers.
+
+    Read more here:
+    https://github.com/mozilla/multi-account-containers/wiki/Moving-between-containers
+
+    @param containerName The container name, fuzzy matched like `-c` on [[tabopen]]. Leave empty to uncontain.
+ */
+//#background
+export async function recontain(containerName = "") {
+    const thisTab = await activeTab()
+
+    let container
+    await Container.fuzzyMatch(containerName[0])
+        .then(match => {
+            container = match
+        })
+        .catch(() => {
+            container = Container.DefaultContainer.cookieStoreId
+        })
+
+    await openInNewTab(thisTab.url, {
+        active: true,
+        related: true,
+        cookieStoreId: container,
+    })
+    return browser.tabs.remove(thisTab.id)
+}
+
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -2567,6 +2567,32 @@ export async function viewcontainers() {
             .replace(/ /g, "%20")
 }
 
+/** Opens the current tab outside the container
+
```
bovine3dom: Please add a link to https://github.com/mozilla/multi-account-containers/wiki/Moving-between-containers and, because no-one will read it, a precis that is short enough to appear in the interactive help warning people that it might be a bad idea.

src/excmds.ts
```diff
@@ -2567,6 +2567,38 @@ export async function viewcontainers() {
...
+    https://github.com/mozilla/multi-account-containers/wiki/Moving-between-containers
+
+    @param containerName The container name, fuzzy matched like `-c` on [[tabopen]]. Leave empty to uncontain.
+ */
+//#background
+export async function recontain(containerName = "") {
+    const thisTab = await activeTab()
+
+    let container
+    await Container.fuzzyMatch(containerName[0])
```
bovine3dom: There's something strange going on with this fuzzy match.

```
:containercreate blah
:recontain blah # switches to default Banking container
:tabopen -c blah news.bbc.co.uk # opens blah container
```

src/excmds.ts
```diff
@@ -2567,6 +2567,38 @@ export async function viewcontainers() {
...
+    Transfering URLs from one container to another allows websites to track
+    you across those containers.
+
+    Read more here:
+    https://github.com/mozilla/multi-account-containers/wiki/Moving-between-containers
+
+    @param containerName The container name, fuzzy matched like `-c` on [[tabopen]]. Leave empty to uncontain.
+ */
+//#background
+export async function recontain(containerName = "") {
```
bovine3dom: I think we should probably call this function `containermove` to match up with the other container management commands. You can put an alias in the default config for `recontain` if you like.

src/excmds.ts
bovine3dom: @robertgzr do you think you'll have a chance to look at this any time soon?

bovine3dom:
I suspect it wouldn't be hard to fix this PR if anyone is interested.

Issue #1660, opened 2019-06-05
add `extoptions`


bovine3dom:
From #90.

Now we have #1479, this should be straightforward to add: `extoptions` as an ex-command that has completions for all installed extensions and then opens the `optionsUrl` for the one selected. Should probably open in a new tab as Tridactyl won't be able to operate on the page opened.

PR #1851, opened 2019-09-13
add extoptions excmd

rektrex:
Adds an excmd: `extoptions` as specified in [#1660](https://github.com/tridactyl/tridactyl/issues/1660#issue-452658372).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4409,4 +4409,13 @@ browser.runtime.onInstalled.addListener(details => {
     // could add elif "update" and show a changelog. Hide it behind a setting to make it less annoying?
 })
 
+/** Opens optionsUrl for the selected extension in a new tab.
+ *
+ * NB: Tridactyl cannot run on this page!
+ */
+//#content
+export async function extoptions(optionsUrl: string) {
+    return tabopen(optionsUrl)
+}
+
 // vim: tabstop=4 shiftwidth=4 expandtab
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -158,6 +158,7 @@ import { firefoxVersionAtLeast } from "@src/lib/webext"
 import * as rc from "@src/background/config_rc"
 import * as css_util from "@src/lib/css_util"
 import * as Updates from "@src/lib/updates"
+import * as Extensions from "@src/lib/extension_info"
 
 ALL_EXCMDS = {
     "": BGSELF,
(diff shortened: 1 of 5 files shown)


bovine3dom:
This is neat, thanks. Could you add aliases for Pentadactyl compatibility? https://github.com/5digits/dactyl/commit/d7e8bf1543be916835d639d345743b33501f5f2a#diff-a174ed1766e22517f9b129d5f6ad9318 listed at the bottom. They go in src/lib/config.ts under exaliases.

src/excmds.ts
```diff
@@ -4409,4 +4409,13 @@ browser.runtime.onInstalled.addListener(details => {
     // could add elif "update" and show a changelog. Hide it behind a setting to make it less annoying?
 })
 
+/** Opens optionsUrl for the selected extension in a new tab.
+ *
+ * NB: Tridactyl cannot run on this page!
+ */
+//#content
+export async function extoptions(optionsUrl: string) {
+    return tabopen(optionsUrl)
```
bovine3dom: I think I would prefer it if the clever bit of matching the extension name happened here so it could be useful as a non-interactive command. What do you think?

Also it might make more sense to open a `winopen -popup` if Tridactyl can't run on these pages, but I'm not 100% sure.

rektrex: > I think I would prefer it if the clever bit of matching the extension name happened here so it could be useful as a non-interactive command.

Do you mean to say that `extoptions` should receive an extension object and then get the values of its members? 

> Also it might make more sense to open a winopen -popup if Tridactyl can't run on these pages, but I'm not 100% sure.

Sure, we can do that if you think it makes more sense that way.

bovine3dom: E.g. extoptions("ublock") should open the ublock extensions page. If extension names are often not unique then we should allow the URL approach too, but I think that would clutter up command history a little much.

If the extension names are almost always unique then the completion should insert the name rather than the URL.

src/lib/config.ts
```diff
@@ -509,6 +509,7 @@ export class default_config {
         "mkt!": "mktridactylrc -f",
         "mktridactylrc!": "mktridactylrc -f",
         mpvsafe: "js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('mpv ' + url))",
+        exto: "extoptions"
```
```suggestion
        exto: "extoptions",
        extpreferences: "extoptions",
        extp: "extpreferences",
```

rektrex: Thank  you for pointing this out. Fixed.

src/excmds.ts
rektrex: Got this working. Should I make the options page open in a popup window?

bovine3dom: Could you give it a go and see what feels more natural to you?

I won't have time to actually test this code for quite a while by the way.

rektrex: I think a popup window feels more natural. I'll make the options open in a popup. Thank you.

PR #1853, opened 2019-09-14
Fix #1365

### LINKED ISSUE #1365: Not working with web components (e.g. input triggers Tridactyl binds in shadow dom)
Does not work on input boxes in web components
I can post a jsFiddle if that would help but basically do this
Create a template object in a shadow dom containing an input box (type=text in my test case)

Because the parent element is not editable then I think the element is seen by tridactyl as not allowing input so directs all keyboard input to itself. Evidence for this is that you can quite happily delete text content and even add spaces. But all alpha chars fail. Fires tridactyl instead.

if I set contentEditable=true on the web component it's happy but of course this is undesirable in the context of the component itself :)

Could it be that possibly a bit of recursion is needed on each test component if the element contains a dash (i.e. a custom component?) Could be fun! :)

pvs:
Closes #1365

### DIFF
--- a/src/content/controller_content.ts
+++ b/src/content/controller_content.ts
@@ -2,7 +2,7 @@ import { isTextEditable } from "@src/lib/dom"
 import { contentState, ModeName } from "@src/content/state_content"
 import Logger from "@src/lib/logging"
 import * as controller from "@src/lib/controller"
-import {KeyEventLike} from "@src/lib/keyseq"
+import { KeyEventLike } from "@src/lib/keyseq"
 
 import * as hinting from "@src/content/hinting"
 import * as gobblemode from "@src/parsers/gobblemode"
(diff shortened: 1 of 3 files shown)


bovine3dom:
Thanks for this. I won't be able to test it for a while. Perhaps as soon as next week but potentially later.

src/content/controller_content.ts
```diff
@@ -110,10 +110,20 @@ function* ParserController() {
...
+
                 // _just to be safe_, cache this to make the following
                 // code more thread-safe.
                 const currentMode = contentState.mode
-                const textEditable = keyevent instanceof KeyboardEvent ? isTextEditable(keyevent.target as Element) : false
+                const textEditable =
+                    keyevent instanceof KeyboardEvent
+                        ? shadowRoot === null
+                            ? isTextEditable(keyevent.target as Element)
+                            : isTextEditable(shadowRoot.activeElement)
```
bovine3dom: Does any of this code need to be specific to keyboard events?

I ask because "isTextEditable" appears in a few places and `gi` doesn't work on inputs in the shadow DOM. It would be nice if it just worked everywhere.

```
src/content.ts
293:            dom.isTextEditable(document.activeElement) &&
302:            !dom.isTextEditable(document.activeElement)

src/lib/dom.ts
23:export function isTextEditable(element: Element) {
513:    if (isTextEditable(elem)) {
544:        if (isTextEditable(e.target as HTMLElement)) {

src/content/controller_content.ts
1:import { isTextEditable } from "@src/lib/dom"
116:                const textEditable = keyevent instanceof KeyboardEvent ? isTextEditable(keyevent.target as Element) : false
```

pvs:
Initial commit was just to not binds when an input field inside shadow dom is focused. Didn't care about focusing input field with keybindings. Reason why `gi` or any other hinting inside shadow DOM doesn't work is Tridactyl doesn't query elements buried within the shadow DOM. It should now be fixed.

bovine3dom:
That's great, thanks! arix on Matrix has confirmed it works so merging.

cmcaine:
I don't think we should have merged this, it has a bunch of unrelated whitespace and formatting changes in it which make the commit log hard to read.

I'm leaving this as a comment to remind me to maybe revert it and re-apply a clean commit.

bovine3dom:
I think it is the fault of #1841.

Issue #1378, opened 2019-02-21
Custom/Native Theme Weirdness on Windows


cougarten:
Hi,
thanks for this add-on! I found a bunch of issues with themes. Sadly I couldn't get it to work, so it's a bit hard to separate the issues.

## my system:

FF nightly 67.0a1 (2019-02-16)
windows 10
native 0.1.10 
trydactyl 1.14.8pre2398

## 1. Theme dir does not match config dir

- if findUserConfigFile() in native_main.py does not find a config file, than trying to use a custom theme will result in an error (path not found)
- if findUserConfigFile() does find a config path, tridactyl will look for my theme in ~/themes/

I think the behavior should be:

- if findUserConfigFile() finds a config file: search for a themes/ folder in the same folder the config file was found
- if findUserConfigFile() does not find a config file, it should look for themes/ in a non-generic path (e.g. ~/.tridactyl/themes or ~/.config/tridactyl/themes

## 2. Theme dir on windows uses wrong slashes

When trying to load a non-existent theme I get an error (that's fine). In the message I see the path gets constructed with a mix of forward and backward slashes: C:\Users\myusername/themes/mytheme.css

correct for windows would be `\` but I don't know what your code expects or needs.

I don't know if this is the cause for the next problem.

## 3. Loading Theme fails without error, falls back to default

trying to load my theme fails without an error. If a e.g. used "dark" before, it falls back to the default theme.

chaorace:
Issue 2 is caused by this line:
https://github.com/tridactyl/tridactyl/blob/ddfb93acdb787c151156c6c06292b15520e6878e/src/excmds.ts#L442

(await browserBg.runtime.getPlatformInfo().os) evaluates to null

bovine3dom:
For 1. we need to rewrite https://github.com/tridactyl/tridactyl/blob/e3b44a69a9729ff8903a72b004c3bf6a44d44432/src/excmds.ts#L481 slightly

I'm not sure why 3. is happening, for me e.g. `:colours thisthemeaintreal` gives me an error as I would expect

PR #1862, opened 2019-09-17
Correctly detects Windows when applying colors

chaorace:
Partial solution to #1378
Previously, this expression would evaluate to null, producing a false negative on Windows systems, due to a syntax issue


bovine3dom:
I'm surprised this wasn't caught by our tooling. Thanks!

Issue #1818, opened 2019-08-25
Sync rc from web link


bovine3dom:
cVim has a neat little thing where you can give it a link to a GitHub gist containing an RC file. Could be nice for people who can't install the native messenger at work but still want to use their RC file.

bovine3dom:
`source --url` added a while ago; `:autocmd TriStart .* source_quiet --url [url]` should do the trick.

PR #1866, opened 2019-09-19
add exmcmds to load RC file from url

rektrex:
From #1818, adds two excmds: source_from_url and source_from_url_quiet to load RC file from a url. The provided url must point to a file.
Example: like this:- 

https://gist.githubusercontent.com/rektrex/97fc44e0ea440238b6d60fae9fe4b562/raw/b6d9a81193d8fa4c6cfd99135e2cb454e411cb73/tridactylrc

and not this:-

https://gist.github.com/rektrex/97fc44e0ea440238b6d60fae9fe4b562

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -726,7 +726,9 @@ export async function mktridactylrc(...args: string[]) {
  *
  * If no argument given, it will try to open ~/.tridactylrc, ~/.config/tridactyl/tridactylrc or $XDG_CONFIG_HOME/tridactyl/tridactylrc in reverse order. You may use a `_` in place of a leading `.` if you wish, e.g, if you use Windows.
  *
- * If no url is specified with the `--url` flag, the current page's URL is used to locate the RC file.
+ * If no url is specified with the `--url` flag, the current page's URL is used to locate the RC file. Ensure the URL you pass (or page you are on) is a "raw" RC file, e.g. https://raw.githubusercontent.com/tridactyl/tridactyl/master/.tridactylrc and not https://github.com/tridactyl/tridactyl/blob/master/.tridactylrc.
+ *
+ * Tridactyl won't run on many raw pages due to a Firefox bug with Content Security Policy, so you may need to use the `source --url [URL]` form.
  *
  * On Windows, the `~` expands to `%USERPROFILE%`.
  *
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -751,6 +751,26 @@ export async function source_quiet(...fileArr: string[]) {
     }
 }
 
+/** Use tridactylrc located at the given url
+ * @param url the url where the raw version of tridactylrc can be found
+ */
+//#background
+export async function source_from_url(url: string) {
+    if (!url) return
+    await rc.sourceFromUrl(url)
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -751,6 +751,26 @@ export async function source_quiet(...fileArr: string[]) {
     }
 }
 
+/** Use tridactylrc located at the given url
+ * @param url the url where the raw version of tridactylrc can be found
```
bovine3dom: Thanks for this PR! I didn't realise it would be so easy. Could you check that `autocmd TriStart .* source-from_url ...` works, and if it does, add it to this help?

src/excmds.ts
```diff
@@ -751,6 +751,26 @@ export async function source_quiet(...fileArr: string[]) {
...
+ */
+//#background
+export async function source_from_url(url: string) {
```
bovine3dom: I think I'd be happier if this was combined with the current `source` command, with this branch executed if there was a `--url` flag, e.g. `source --url news.bbc.co.uk`

src/excmds.ts
```diff
@@ -751,6 +751,26 @@ export async function source_quiet(...fileArr: string[]) {
...
+    if (!url) return
+    await rc.sourceFromUrl(url)
```
bovine3dom: Could you make sure we do some helpful stuff to turn URL-alikes to valid URLs? So we can leave off the schema, e.g. gist.github.com...

src/excmds.ts
rektrex: How do I test this? `src/lib/config.ts` has a default `autocmd` on `TriStart` but replacing it `".*": "source_from_url_quiet(fixed_url)"` doesn't do anything.

src/excmds.ts
rektrex: Sure. Will fix this.

src/excmds.ts
rektrex: Sorry I overlooked this. Will fix this.

src/excmds.ts
bovine3dom: That should work. I guess it doesn't, so don't add it to the help :)

src/excmds.ts
bovine3dom: Probably worth having `source --url` with no args or `source --url %` source the current page, too :)

src/excmds.ts
bovine3dom: You might want to look at what we do with `open` but obviously don't fall back to running a search query.

src/excmds.ts
rektrex: Tridactyl doesn't work on "raw" files(at least for me). So, a `source --url` without an url is not required.

bovine3dom: It works if you `set csp clobber`, which you shouldn't do unless you understand what it does. It's a Firefox bug anyway which Mozilla is fixing, albeit very slowly, so I would be happier if we supported it. It should fetch the URL rather than page content as otherwise you'd have to work around the mode indicator etc.

rektrex: Thank you. Adding this now.

PR #1895, opened 2019-10-13
Add attempt at context menu hint mode

### LINKED ISSUE #1454: [Feature Request]: hint option to emulate right click on links
Is it possible to add an option to :hint so that it can open the context menu as it is done by a mouse right click ?
Use case: some extensions add a context menu entry to enable a specific action that is unreachable without a mouse

thanks

bovine3dom:
Would fix #1454.

Doesn't seem to do anything, though. No errors appear either. :(

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3680,6 +3680,7 @@ export function unset(...keys: string[]) {
         - -# yank an element's anchor URL to clipboard
         - -c [selector] hint links that match the css selector
           - `bind ;c hint -c [class*="expand"],[class="togg"]` works particularly well on reddit and HN
+        - -C open context menu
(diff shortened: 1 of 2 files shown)


lydell:
As far as I know, neither web pages nor extensions can trigger the native context menu.

Legacy Firefox extensions could, though. See https://github.com/akhodakivskiy/VimFx/blob/75cce0ec4079d2e979f7fe1cd87aca0f2ec47aa8/extension/lib/utils.coffee#L356-L365

bovine3dom:
Yeah, I think VVimpulation just added some broken code (see #1454). I saw it and got all excited without fact checking it.

I'll close this PR. Thanks!

PR #1896, opened 2019-10-13
Hide hoverlink on command line open

bovine3dom:
Inspired by VVimpulation - https://github.com/amedama41/vvimpulation/commit/53065d015d1e9a892496619b51be83771f57b3d5

Currently scrolls page to the bottom; should fix that before merging.


bovine3dom:
Probably need to find an element in the viewport and attach to that to avoid scrolling.

PR #1902, opened 2019-10-16
autocontainmode: "strict" | "relaxed"

arcnmx:
The current (and still default) autocontain behaviour is to intercept all page loads that match an aucon pattern, and may even replace a tab to do so. "relaxed" mode is introduced here as a way to only apply the rules when opening a new tab, and makes containers more sticky. Alternative mode names to consider could be `"always" | "newtab"`?

### DIFF
--- a/src/lib/autocontainers.ts
+++ b/src/lib/autocontainers.ts
@@ -77,6 +77,9 @@ export class AutoContain implements IAutoContain {
     ): Promise<browser.webRequest.BlockingResponse> => {
         if (!this.autocontainConfigured()) return { cancel: false }
 
+        // Only handle in strict mode.
+        if (Config.get("autocontainmode") === "relaxed") return { cancel: false }
+
         // Only handle http requests.
         if (details.url.search("^https?://") < 0) return { cancel: false }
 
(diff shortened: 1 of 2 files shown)


src/lib/config.ts
```diff
@@ -439,6 +439,13 @@ export class default_config {
         // "youtube.com": "google",
     })
 
+    /**
+     * Strict mode will always ensure a domain is open in the correct container, replacing the current tab if necessary.
+     *
+     * Relaxed mode is less aggressive and instead treats container domains as a default when opening a new tab.
+     */
+    autocontainMode: "strict" | "relaxed" = "strict"
```
```suggestion
    autocontainmode: "strict" | "relaxed" = "strict"
```

All other settings are lowercase :)

src/lib/autocontainers.ts
```diff
@@ -77,6 +77,9 @@ export class AutoContain implements IAutoContain {
     ): Promise<browser.webRequest.BlockingResponse> => {
         if (!this.autocontainConfigured()) return { cancel: false }
 
+        // Only handle in strict mode.
+        if (Config.get("autocontainMode") === "relaxed") return { cancel: false }
```
bovine3dom:
```suggestion
        if (Config.get("autocontainmode") === "relaxed") return { cancel: false }
```

PR #1914, opened 2019-10-18
Replace awesome-typescript-loader with ts-loader

farnoy:


webpack.config.js
```diff
@@ -1,7 +1,9 @@
-const { TsConfigPathsPlugin } = require("awesome-typescript-loader")
+const TsconfigPathsPlugin = require("tsconfig-paths-webpack-plugin");
```
bovine3dom: You have a rogue semi-colon again ;)

bovine3dom:
If you squash this and improve the commit message (explain why we're switching, i.e. that the other loader is unmaintained), I'll merge it. Thanks!

farnoy:
@bovine3dom Should be fine now

PR #1915, opened 2019-10-18
Enable some low-hanging TypeScript strictness options

farnoy:


bovine3dom:
@farnoy could you check the linting errors? They look genuine

farnoy:
@bovine3dom should be good now

PR #1923, opened 2019-10-21
Fix error in default config

notJerl:
Fixes an error in the default config that prevents users from removing or rebinding these bindings.


notJerl:
Can do.

PR #1948, opened 2019-10-26
WIP: Type safe messaging

farnoy:
_will fill this in later_


src/tridactyl.d.ts
```diff
@@ -213,6 +213,10 @@ declare function html(
     ...values: any[]
 ): HTMLElement
 
+declare namespace browser.search {
```
farnoy: I removed this too eagerly in my earlier PR

src/message_protocols.ts
```diff
@@ -0,0 +1,7 @@
+// This file re-exports types for message protocols for background, content and commandline contexts
+
```
farnoy: This is done to safely export just the types. This won't translate to any runtime import, which is very important as we do not want to import content code in the background page, or vice versa.

src/message_protocols.ts
```diff
@@ -0,0 +1,7 @@
...
+export { Messages as Background } from "@src/background"
+export { Messages as CmdlineFrame } from "@src/commandline_frame"
+export { Messages as Content } from "@src/content"
```
farnoy: I don't like the fact that content script registers 3 different handlers because of ordering requirements, but I will try to defragment this later

src/lib/webext.ts
```diff
@@ -26,14 +26,8 @@ export function getContext() {
...
 
-export let browserBg
-
 // Make this library work for both content and background.
-if (inContentScript()) {
-    browserBg = browserProxy
-} else {
-    browserBg = browser
-}
+export const browserBg = inContentScript() ? browserProxy : browser
```
farnoy: This is sort of unrelated, but improves coverage of types a lot as the type is inferred correctly now

src/lib/messaging.ts
```diff
@@ -1,135 +1,84 @@
...
-    }
-    return handler
+    });
 }
 
 /** Send a message to non-content scripts */
-export async function message(type: NonTabMessageType, command, args?) {
-    return browser.runtime.sendMessage({ type, command, args } as Message)
+export function message<Root>() {
+    return async <Type extends keyof Root, Command extends keyof Root[Type]>(
```
farnoy: returning this closure sucks, but I was unable to come up with a better solution.

When I tried default values for some parameters, to be able to invoke `message<Messages.Background>(...)`, it would never infer types properly. The other thing I tried was `message(Messages.Background, ...)`, but this quickly backfires because it's used as a value and the module needs to be imported at runtime, which breaks all kinds of things if we import code across background & content scripts.

src/excmds.ts
```diff
@@ -3129,9 +3126,8 @@ function parse_bind_args(...args: string[]): bind_args {
         - [[reset]]
 */
 //#background
-export function bind(...args: string[]) {
```
farnoy: These are just some minor changes to typecheck with new return types of some functions. This has to be async to have a uniform interface with the shim generated for content scripts

src/completions/Find.ts
```diff
@@ -1,116 +0,0 @@
-import { activeTabId } from "@src/lib/webext"
```
farnoy: This file was not used but broke after new types were propagated, so I just removed it instead of fixing it

src/background/download_background.ts
```diff
@@ -146,14 +146,3 @@ export async function downloadUrlAs(url: string, saveAs: string) {
         browser.downloads.onChanged.addListener(onDownloadComplete)
     })
 }
-
```
farnoy: To defragment the handlers, I moved this to the main listener for bg page

src/commandline_frame.ts
```diff
@@ -318,7 +321,7 @@ function applyWithTmpTextArea(fn) {
 
 /** @hidden **/
 export async function setClipboard(content: string) {
-    await Messaging.messageOwnTab("commandline_content", "focus")
+    await Messaging.messageOwnTab<Messages.CommandlineContent>()("commandline_content", "focus", [])
```
bovine3dom: I don't like the empty array at the end, what's it for?

The `CommandlineContent` and `commandline_content` repetition is a little worrying too :)

src/excmds.ts
bovine3dom: Can they be a separate PR? Doesn't look like it has much to do with the messaging.

src/tridactyl.d.ts
bovine3dom: Make it a separate one, there's a free t-shirt in it, remember! :)

bovine3dom:
It's been a while - can you remember where we got to with this @farnoy?

farnoy:
@bovine3dom Yep, the background messaging improvements have been merged. I need to prepare another, smaller PR with just the page script communication going through a similar system.

PR #1949, opened 2019-10-26
Fix type of browserBg and adjust other code for it

farnoy:
this ternary operator picks up the type properly and more browserBg APIs are checked

### DIFF
--- a/src/lib/messaging.ts
+++ b/src/lib/messaging.ts
@@ -79,7 +79,7 @@ export async function messageActiveTab(
     return messageTab(await activeTabId(), type, command, args)
 }
 
-export async function messageTab(tabId, type: TabMessageType, command, args?) {
+export async function messageTab(tabId, type: TabMessageType, command, args?): Promise<any> {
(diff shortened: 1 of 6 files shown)


src/lib/messaging.ts
```diff
@@ -79,7 +79,7 @@ export async function messageActiveTab(
     return messageTab(await activeTabId(), type, command, args)
 }
 
-export async function messageTab(tabId, type: TabMessageType, command, args?) {
+export async function messageTab(tabId, type: TabMessageType, command, args?): Promise<any> {
```
farnoy: This keeps the old type and stops us from having to fix up a bunch more cases

src/tridactyl.d.ts
```diff
@@ -213,6 +213,10 @@ declare function html(
     ...values: any[]
 ): HTMLElement
 
+declare namespace browser.search {
```
farnoy: I removed this too eagerly in my earlier PR, let's bring it back

PR #2052, opened 2019-12-19
Fix #2050: add WIP visual mode

### LINKED ISSUE #2050: bind `Y` to select highlighted text
Tridactyl version: 1.17.1pre3367-30eadf31
Firefox version: Mozilla Firefox 68.2.0
Operating system: linux

The shortcut is handy on reducing mouse clicks when trying to copy text phrases from a webpage. 

`bind Y composite js document.getSelection().toString() | yank`

bovine3dom:
We need to make merging of nmaps happen at runtime before we merge this, or we'll break the binds of people who are just selecting text to help them read. Thoughts on how to do this would be appreciated.

The mode as it stands already works quite nicely with caret mode. Adding hjkl binds to extend the selection would probably be pretty easy.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -355,6 +355,15 @@ config.getAsync("leavegithubalone").then(v => {
     }
 })
 
+document.addEventListener("selectionchange", () => {
+    const selection = document.getSelection()
+    if (selection.anchorOffset == selection.focusOffset) {
+        contentState.mode = "normal"
+    } else {
+        contentState.mode = "visual"
+    }
+})
+
 // Listen for statistics from each content script and send them to the
 // background for collection. Attach the observer to the window object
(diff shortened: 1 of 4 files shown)


Philipp0205:
Is this the implemenation of a visual mode for this extention? I would be really happy.

bovine3dom:
Yes and no - it doesn't do very much (e.g. changing the selection) but it might form the basis for that. It currently just has a bind for copying text selected with the mouse or caret mode and automatically enters/leaves the mode when text is selected or deselected.

bovine3dom:
I need to check what this does if you select text in insert / input mode

bovine3dom:
It doesn't do anything in insert mode, which, I think, is what we want. Merging.

PR #2063, opened 2019-12-26
Drop ftdetect hack from rc output, makes it difficult to use plugin

alerque:
Having a modeline with a bogus filetype in the RC files makes it really hard to use [`vim-tridactyl`](https://github.com/tridactyl/vim-tridactyl) which will detect and set the filetype just fine as long as there are no dirty hacks resetting it afterwards.

The syntax is far from complete, but I personally think it's already far more usable than forcing `vim` syntax onto something that isn't. I suggest leaving this modeline hack off entirely, starting now.

### DIFF
--- a/.tridactylrc
+++ b/.tridactylrc
@@ -148,5 +148,5 @@ autocmd DocStart ^http(s?)://www.reddit.com js tri.excmds.urlmodify("-t", "www",
 " Mosquito nets won't make themselves
 autocmd DocStart ^http(s?)://www.amazon.co.uk js tri.excmds.urlmodify("-t", "www", "smile")
 
-" This will have to do until someone writes us a nice syntax file :)
-" vim: set filetype=vim:
+" For syntax highlighting see https://github.com/tridactyl/vim-tridactyl
+" vim: set filetype=tridactyl
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1485,7 +1485,7 @@ export function parseConfig(): string {
     if (p.logging.length > 0)
         s.logging = `" Logging\n${p.logging.join("\n")}\n\n`
 
-    const ftdetect = `" vim: set filetype=vim:`
+    const ftdetect = `" For syntax highlighting see https://github.com/tridactyl/vim-tridactyl\n" vim: set filetype=tridactyl`
 
     return `${s.general}${s.binds}${s.subconfigs}${s.aliases}${s.aucmds}${
         s.aucons


alerque:
The e2e test that failed here doesn't look like even remotely related to what I changed. Also of the two iterations of this commit it ran on the error was wildly different each time. Is the CI testing currently wonky or am I doing something wrong?

bovine3dom:
The CI is quite unreliable. I only get worried if it fails lots and lots of tests.

bovine3dom:
Surely we should change this to our own filetype? Can vim magically detect filetype even if the filename is nonstandard, e.g. `mysecretrc`?

alerque:
VIM's file type detection can in fact snoop on the contents of a file and try to guess its format, but such monkey business is usually reserved for when two file name patterns clearly conflict. It's generally best to have file name patterns be the first line of attack.

No we won't catch `mysecretrc`, but why would we need to? What would be the use case for that? Vim won't detect _most_ file types if you disguise the names form their common or expected patterns.

These files I'm changing here certainly don't need any overrides. If people rename them to some disguised naming pattern then they can add their own overrides as well.

I will add file name patterns for the other formats looked for by the current native bridge script to the plugin.

bovine3dom:
`mkt` accepts a file path as an argument. I don't really know what it's used for - perhaps sourcing specific file based on hostname? Source also accepts a filename.

E.g. I could have a series of complex commands I want to execute when I visit a page - `autocmd DocLoad www.google.com source ~/weirdstuffrc`. RC files allow newlines in JS so are a much nicer way of executing it.

src/lib/config.ts
```diff
@@ -1485,11 +1485,9 @@ export function parseConfig(): string {
     if (p.logging.length > 0)
         s.logging = `" Logging\n${p.logging.join("\n")}\n\n`
 
-    const ftdetect = `" vim: set filetype=vim:`
```
bovine3dom: For clarity, I'm suggesting we change this to `filetype=tridactyl`

.tridactylrc
```diff
@@ -147,6 +147,3 @@ command translate js let googleTranslateCallback = document.createElement('scrip
 autocmd DocStart ^http(s?)://www.reddit.com js tri.excmds.urlmodify("-t", "www", "old")
 " Mosquito nets won't make themselves
 autocmd DocStart ^http(s?)://www.amazon.co.uk js tri.excmds.urlmodify("-t", "www", "smile")
-
-" This will have to do until someone writes us a nice syntax file :)
```
bovine3dom: Can we replace this comment with a link to vim-tridactyl?

src/lib/config.ts
bovine3dom: We could even append it to only nonstandard RC names if you would like :)

PR #2137, opened 2020-02-03
Remove onchanged listener from state

bovine3dom:
As detailed in #1764, the `onChanged` event leads to race conditions.

This PR sketches out how we might move away from it in state.ts. Once we've worked that out, we can do a similar thing to src/lib/config.ts and fix, with luck, #1409. 

Todo:

- ~~Get tabs to reply to messages~~
- ~~Look into named locks (has someone already done this for WebExtensions before? Probably, right?)~~ done in #2161
- ~~fix `g;`~~

I'm not sure I'll have time to do either so any help with this would be appreciated.

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -15,6 +15,7 @@
 */
 
 import Logger from "@src/lib/logging"
+import * as messaging from "@src/lib/messaging"
 const logger = new Logger("state")
 
 class State {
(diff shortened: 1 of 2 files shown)


src/state.ts
```diff
@@ -55,17 +56,39 @@ const state = (new Proxy(overlay, {
...
         return true
     },
 }))
 
-browser.storage.onChanged.addListener((changes, areaname) => {
-    if (areaname === "local" && "state" in changes) {
-        Object.assign(overlay, changes.state.newValue)
-    }
+// Keep instances of state.ts synchronised with each other
+messaging.addListener("state", message => {
```
bovine3dom: switch to 
```
... "state", (message, sender, sendResponse) => {
    do_stuff()
    sendResponse("I'm done")
}

**Comment by bovine3dom** [comment]
As it stands, this breaks `g;`. cmdHistory still works which is perplexing.

File: src/state.ts [hunk 2]
```diff
@@ -53,19 +55,38 @@ const state = (new Proxy(overlay, {
...
+            await locks.acquire("state")
+
+            logger.debug("State changed!", property, value)
+            target[property] = value
+            browser.storage.local.set({ state: target } as any)
+
+            // Wait for reply from each script to say that they have updated their own state
+            await Promise.all([
+                // dispatch message to all content state.ts's
+                messaging.messageAllTabs("state", "stateUpdate", [{state: target}]),
```
bovine3dom: If there's a tab that Tridactyl can't run on open, e.g. `about:addons`, this causes a `# Error: Could not establish connection. Receiving end does not exist.` to pop-up whenever you click in an input box.

We should probably catch those errors in `messageAllTabs`, as I don't think they'll ever be useful.

PR #2162, opened 2020-02-16
Hint for partially visible links

mozbugbox:
Links with only bottom half visible are hinted too.

Partially visible links at the top of the screen were currently ignored by the hinter.


bovine3dom:
Thanks for this! I'll try to test it out today.

PR #2169, opened 2020-02-17
Sort completion result based on Fuse.js score

mozbugbox:
The completion uses Fuse.js to match user filter. The matching is fuzzy, therefore sometime looks weird. The results are much easier to follow by putting better matched results at the top.

There might need some fine tuning on how to make use of the fuse matching for better result. Fuse seemed to have the ability to put weight on searching fields, like `title` has more weight than the `url` etc. I'm not familiar with Fuse to work on that.

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -155,6 +155,7 @@ export abstract class CompletionSourceFuse extends CompletionSource {
     fuse = undefined
 
     protected lastExstr: string
+    protected sortScoredOptions = false
 
     protected optionContainer = html`<table class="optionContainer"></table>`
 
@@ -274,8 +275,10 @@ export abstract class CompletionSourceFuse extends CompletionSource {
         }
 
         // sort this.options by score
-        const sorted_options = matches.map(index => this.options[index])
-        this.options = sorted_options.concat(hidden_options)
+        if (this.sortScoredOptions) {
+            const sorted_options = matches.map(index => this.options[index])
+            this.options = sorted_options.concat(hidden_options)
+        }
     }
 
     /** Call to replace the current display */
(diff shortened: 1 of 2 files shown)


src/completions.ts
```diff
@@ -268,6 +272,10 @@ export abstract class CompletionSourceFuse extends CompletionSource {
         } else {
             this.deselect()
         }
+
+        // sort this.options by score
+        const sorted_options = matches.map(index => this.options[index])
+        this.options = sorted_options.concat(hidden_options)
```
bovine3dom: Need to check whether this overrides e.g. history sorting based on frequency.

mozbugbox: On an orthogonal topic, the history did not go through the fuse match as it override the `updateChain()` which goes through the default score related methods.

bovine3dom:
This PR appears to break the "magic" treatment given to tab numbers. E.g. `b3` should select the second tab such that `b3<Enter>` should always take you to the third tab. With this PR, `b[n]` doesn't always select the nth tab in my testing.

Would you be able to look in to restoring this behaviour?

mozbugbox:
Shouldn't just type the number without the prefix `b`  does that? From the code, a bare number and `#` used to jump to nth tab or last visited tab.

https://github.com/tridactyl/tridactyl/blob/d66db550c66cb6354da3c917c03e8ef890f8c776/src/completions/Tab.ts#L95

bovine3dom:
Yeah, sorry. I was being a little imprecise. I mean if the user presses the keys: b, 2, `<Enter>`, they should be taken to the second tab. Equivalent to typing `:tab 2<Enter>` :)

PR #2181, opened 2020-02-20
Revert "Revert "Merge branch 'config_no_onchanged'""

bovine3dom:
This un-reverts #2171 which seems to break Tridactyl on Windows and Ubuntu (but not Arch) - see #2180. 

:(


src/lib/config.ts
```diff
@@ -1270,11 +1273,38 @@ export function unset(...target) {
...
+                throw e
+            }
+            p = true
+        }
+        return p
+    }
+
+    await Promise.all([
+        // dispatch message to all content state.ts's
+        messaging.messageAllTabs("config", "confUpdate", [settingsobj]),
```
bovine3dom, line 1299: On the one hand I am still traumatised by how long this PR took and how much it broke Tridactyl on various platforms.

On the other hand the `escapehatch` logic we have for finding live Tridactyl tabs could salvage it, maybe?

PR #2188, opened 2020-02-24
Feature: expand selection range to entire node

gramcracker:
The visual mode is great, but for webpages I have found that elements like span can obstruct text selection. I have been using this bind to make selection way easier, and thought others would find it beneficial as well.

### DIFF
--- a/.tridactylrc
+++ b/.tridactylrc
@@ -36,6 +36,9 @@
 " Binds
 "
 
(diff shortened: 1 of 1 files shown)


bovine3dom:
I'm happy for this to be added as a default bind in src/lib/config.ts :)

gramcracker:
cool, I'll add patch it in there and request another pull.

PR #2189, opened 2020-02-24
Feature: new vmap '=' to expand selection to encompass parent nodes c…

gramcracker:
This is in response to an earlier pull request [#2188](https://github.com/tridactyl/tridactyl/pull/2188)


bovine3dom:
This is now merged (but I used `gh pr` rather than `hub` to fiddle with it and that apparently has caused a kerfuffle)

PR #2191, opened 2020-02-25
Allow js/jsb to load local javascript file.

mozbugbox:
Add a flag `-s` to source js files from user's config directory:

    :js -s test.js

With the ability to load js files, `tridactylrc` can have its javascript
code into seperate files.

Since the js code was run by calling `eval()` in local scope of calling
function, the variables and functions decleared in the js file will lost
after eval().

To make the content of javascript file reusable, user could export
functions into some globals like `window.` or `tri.`.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4368,34 +4368,50 @@ export async function echo(...str: string[]) {
  * @hidden
  */
 async function js_helper(str: string[]) {
-    let doSource = false
-    let jsContent = null
     /* tslint:disable:no-unused-declaration */
     /* tslint:disable:no-dead-store */
     let JS_ARG = null
+    let jsContent = null
 
-    while (true) {
-        if (str[0].startsWith("-p")) {
-            JS_ARG = str[str.length - 1]
-            str = str.slice(1, -1)
-        } else if (str[0].startsWith("-s")) {
-            doSource = true
-            str = str.slice(1)
-        } else {
-            break
+    let doSource = false
+    let fromRC = false
+    let done = false
+
+    while (!done) {
+        switch (str[0]) {
+            case "-p":
+                // arg of -p comes from the end of str[]
+                JS_ARG = str.pop()
+                break
+            case "-s":
+                doSource = true
+                break
+            case "-r":
+                doSource = true
+                fromRC = true
+                break
+            default:
+                done = true
+                break
         }
+        if (!done)
+            str.shift()
     }
 
     if (doSource) {
-        const sourceFilename = str.join(" ")
-        if (sourceFilename.search(/(^|[\/\\])..[\/\\]/) >= 0) {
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -4360,6 +4360,48 @@ export async function echo(...str: string[]) {
     return str.join(" ")
 }
 
+/** helper function for js and jsb
+ *
+ * -p to take an extra argument located at the end of str[]
+ * -s to load js script of a source file from the config path
```
bovine3dom: Thanks for this PR. I don't think I like this "relative path-ness" by default. Could you perhaps add a `-r` option for paths relative to rc and have `-s` use the "normal" path behaviour (which is relative to the native messenger's installation directory). I think relative to RC by default could be a little surprising.

Why do you say that "/../" can't be in the path? It should "just work" with Native.read and resolve it to wherever that path is. If it doesn't, that's a bug :)

src/excmds.ts
```diff
@@ -4360,6 +4360,48 @@ export async function echo(...str: string[]) {
...
+        const jsPath = [...rcPath, sourceFilename].join(sep)
+        jsContent = (await Native.read(jsPath)).content
+    } else {
+        jsContent = str.join(" ")
+    }
+    return eval(jsContent)
+}
+
 /**
  * Lets you execute JavaScript in the page context. If you want to get the result back, use `composite js ... | fillcmdline`
```
bovine3dom: Please document all the arguments here (and mention attaching things to `tri.`or `window.`) :)

bovine3dom: Please document all the arguments here (and mention attaching things to `tri.`or `window.`) :)

src/excmds.ts
mozbugbox: The /../ was a safety precaution to prevent the script descent to arbitrary file in users system, like /etc/passwd (well that particular file was less important now a day.)

If you think that Native already does arbitrary things on user system, it doesn't matter. The check can be removed.

bovine3dom: I think I want people to be able to execute arbitrary files on their system, which Native.read already allows :)

Tridactyl's security model is this:

- the command line is isolated from web pages in an iframe and so Tridactyl's commands are not accessible to the web page
- key events are either real key events from users or fake key events triggered by Tridactyl
- users are trusted

So I don't see an issue with having it. Thanks for caring about security!

src/excmds.ts
```diff
@@ -1657,7 +1657,7 @@ export function urlparent(count = 1) {
  *   path). -1 will append to the existing path, -2 will remove the last path
  *   level, and so on.
  *
- *   ```text
+ *   ```plaintext
```
bovine3dom: Ah, good spot! That error has been annoying me for months :)

PR #2195, opened 2020-02-27
Editor

mozbugbox:
Add binding for back-word, delete-word for cmdline editing.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -98,8 +98,8 @@ export class default_config {
         "<ArrowUp>": "ex.prev_history",
         "<ArrowDown>": "ex.next_history",
 
-        "<S-ArrowLeft>": "text.backward_word",
-        "<S-ArrowRight>": "text.forward_word",
+        "<A-b>": "text.backward_word",
+        "<A-f>": "text.forward_word",
         "<C-a>": "text.beginning_of_line",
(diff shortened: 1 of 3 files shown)


src/lib/config.ts
```diff
@@ -97,11 +97,17 @@ export class default_config {
         "<Escape>": "ex.hide_and_clear",
         "<ArrowUp>": "ex.prev_history",
         "<ArrowDown>": "ex.next_history",
+
+        "<S-ArrowLeft>": "text.backward_word",
+        "<S-ArrowRight>": "text.forward_word",
```
bovine3dom: I'm somewhat reluctant to add these two as default binds as shift+arrow to select text is common and I think we might confuse users. I don't think the other two binds are used often so I'm happy for those to be added :)

mozbugbox: Those were vim cmdline bindings.

How about emacs bindnig,  `<A-f>, <A-B>`? Those keys bring up the `File` and `Bookmark` menus by default.

bovine3dom: I imagine people use those less so I'd be happier with that.

We don't follow vim bindings religiously (e.g. ctrl-a in insert mode is purposefully left as select all). People who want to be more pure can add their own.

PR #2201, opened 2020-02-29
Call user defined excmds function through controller

mozbugbox:
If user defined functions are added into tri.excmds, the functions can
be called throught `controller_background/content` message.

The arguments are not converted so the args will all be passed as
strings.


mozbugbox:
Another option will be creating a special namespace just for user defined function to which user can add. Then there will be no arguement type conversion in that namespace.

bovine3dom:
I don't think we need to create a special namespace - I think people can live with everything being strings. We can just document that on `js`.

PR #2229, opened 2020-03-14
Add 16 items to the default value of searchurls.

ethylene-glycol:
The ability to search websites directly from the Tridactyl command line is useful, but there are few websites in the default list for this. This pull request adds sixteen, including eBay.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -621,6 +621,22 @@ export class default_config {
         gentoo_wiki:
             "https://wiki.gentoo.org/index.php?title=Special%3ASearch&profile=default&fulltext=Search&search=",
         qwant: "https://www.qwant.com/?q=",
+        alternativeto: "https://alternativeto.net/browse/search/?q=",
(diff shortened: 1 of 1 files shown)


bovine3dom:
Hey, sorry for the delay on this. Thanks for the PR.

I'd like to hear what other people think about adding extra searchurls. In my experience, I find that usually what I want to do is search Google for something, e.g. "github julia based web browsers" rather than use a searchurl. So far I've heard one person in favour and one person against.

Are you aware that you can add your own searchurls to your own installation with, e.g. `:set searchurls.bookspriceuk http://booksprice.co.uk/compare.do?inputData=`?

I suspect this might be better placed on the Wiki as a list of recommended searchurls.

ethylene-glycol:
I had added these search urls to my own installation.

I use both generic web search engines, such as DuckDuckGo and Google, and specific search urls. Which is better depends on the application, I think.

bovine3dom:
I've added these to the wiki - https://github.com/tridactyl/tridactyl/wiki/Example-searchurls. The general impression I got from users was that we already have too many searchurls and that we shouldn't be adding any more, so I'll close this.

Thanks for this PR :)

PR #2232, opened 2020-03-15
Insert mode fixes

mozbugbox:


src/content/controller_content.ts
```diff
@@ -172,17 +179,20 @@ function* ParserController() {
                 if (response.exstr) {
                     exstr = response.exstr
                     break
-                } else {
+                } else if (!["input", "insert", "ignore"].includes(newMode)) {
```
bovine3dom: I'm concerned by the appearance of "ignore" mode here. Ignore mode is just a "normal" mode like `normal` (i.e. it uses src/parsers/genericmode.ts) but with fewer binds. Why does it need a special case here when, e.g. `visual` does not?

See https://github.com/tridactyl/tridactyl/blob/6f9ecbbc777dc3ef77a7c7fec030fca1f55b3563/src/content/controller_content.ts#L98 for all the generic-mode parsers.

mozbugbox: I thought that the ignore mode basically passes all the keypress to the underlying content. Which mean that the ignore mode is just like the insert mode except that the target is the whole document. In that case, the suffix are not and should not be dealt with.

The `if` testing was just a precaution in case the suffix was somehow useful to the EventListener of it. Does visual mode make use of suffix?

bovine3dom: > ignore mode basically passes all the keypress to the underlying content.

It does this by having fewer keybinds; users can add as many as they like. The only magic is that ignore mode does not enter insert mode when a text box is entered.

I don't know what the suffix does at the moment. I'll look at the code more deeply and get back to you :)

bovine3dom: Ah, I see, suffix is just used here in the mode indicator: https://github.com/tridactyl/tridactyl/blob/ead0d0b914bc622afa82bed156ab3ea447d81c01/src/content.ts#L319

If you think updating it harms performance, we could instead update the suffix only if `config.get("modeindicatorshowkeys")` is true (almost everyone will have it turned off).

mozbugbox: The problem is that if someone want to listen on the suffix changed signal later on, they will find that the signal is not triggered for mystery reason.

bovine3dom: We can leave a comment above it in content/state_content.ts.

I think having it not update in a few hardcoded modes would be more surprising (there's nothing stopping you from adding multi-key binds in insert/input/ignore mode, e.g. behind a tmux style prefix).

If it is the updating `contentState` that is the problem, can we just only update it if there is a change? Most of the time keyEvents will be an empty string.

e.g.

```
let prevSuffix = ""
...
const suffix = keyEvents.map( ...
if (suffix !== prevSuffix) contentState.suffix = suffix
prevSuffix = suffix
```

mozbugbox: Yes, that should work. Look at the code, the `prevSuffix` check was actually in the patch. :)

PR #2234, opened 2020-03-17
Return keys=[] from parser so things don't blow up if exstr is ""

saulrh:
I'm not sure that this is actually what's going on, but it does look like a potential issue and I don't think this will break anything.


saulrh:
Note that this still appears to leave `unbindurl` broken; it just doesn't explode with an error after this.

PR #2259, opened 2020-03-26
Try to unblock overlapped hints

mozbugbox:


bovine3dom:
This seems to work well!

However, this exacerbates #367 as the de-prioritised useless hints are no longer covered up by the useful ones - see e.g. https://www.bbc.co.uk/news/uk-52052880#comp-most-popular in the "Most Read" side bar where in current Tridactyl only the useful links are shown, but in this PR some useless ones are shown too (on my machine these are the hints that appear below the real hints).

I think we should think about how we can work around this instead. Perhaps JS hints should be a different colour (e.g. grey? lighter red?) so that users can be trained to ignore them on sites where they don't work? Alternatively perhaps we could not try to correct for overlap with JS hints?

mozbugbox:
I agree that coloring JS hint differently will help. Not unblocking JS hint might not work well as user would always wonder what's on the blocked hint.

bovine3dom:
Thanks for your work and patience on this @mozbugbox! Merged :)

PR #2364, opened 2020-05-08
Add nmode mode

bovine3dom:
Useful for sending keys to the page in normal mode and sending keys to Tridactyl in ignore mode.

To my surprise, this mostly works.

Todo:

- remove duplication of code (especially where there are "borrowed from" comments - these should instead be moved to suitable files in /lib/ and exported to here and where they were borrowed from)
- deal with keys without ex-commands (these should count towards numcommands - currently it just pops up with an error) - recommended testing: 

```
www.github.com
<C-v>
/
# search box should be highlighted with no errors popping up
```
- improve docs

@tek said he would be happy to work on this.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -58,6 +58,7 @@ This is a (non-exhaustive) list of the most common normal-mode bindings. Type `:
 -   `Shift` + `Insert` — enter "ignore mode". Press `Shift` + `Insert` again to return to "normal mode".
 -   `ZZ` — close all tabs and windows, but only "save" them if your about:preferences are set to "show your tabs and windows from last time"
 -   `.` — repeat the last command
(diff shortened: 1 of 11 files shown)


tek:
:wave:

tek:
so the error came from `undefined` being used in the string for the returned excmd.
Behaviour is a bit weird, if I do the `<c-v>/` thing on github it selects the quick finder, but if I change the mapping to `n = 3`, it selects the search bar.

If I press `<c-v>j` in a repository though, it is processed by the page correctly.

I wasn't sure what to do about the code you marked as "borrowed from controller_content.ts".

What additional documentation should be written?

tek:
:confetti_ball:

PR #2384, opened 2020-05-17
Scrolling.ts refactor

MatiasStorm:
While looking at an issue with the scrolling I had a hard time following the code. So I have tried to refactor [scrolling.ts](../tree/master/src/content/scrolling.ts) to make the naming more explicit, at least to me I think.

Hope you can use the changes.

### DIFF
--- a/src/content/scrolling.ts
+++ b/src/content/scrolling.ts
@@ -3,12 +3,12 @@ import * as config from "@src/lib/config"
 type scrollingDirection = "scrollLeft" | "scrollTop"
 
 const opts = { smooth: null, duration: null }
-async function getSmooth() : Promise<string> {
+async function getSmooth(): Promise<string> {
     if (opts.smooth === null)
         opts.smooth = await config.getAsync("smoothscroll")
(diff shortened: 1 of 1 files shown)


src/content/scrolling.ts
```diff
@@ -115,35 +115,35 @@ const verticallyScrolling = new Map<Node, ScrollingData>()
...
  */
 export async function scroll(
-    x: number = 0,
-    y: number = 0,
+    xDistance : number = 0,
+    yDinstance : number = 0,
     e: Node,
     duration?: number,
-) {
+) : Promise<Boolean> {
```
bovine3dom, line 122:
```suggestion
) : Promise<boolean> {
```

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html :)

MatiasStorm:
You are very welcome! Yeah i started out thinking Vim-Vixen was pretty cool, until i found Tridactyl which I my opinion is way better. So I'm probably gonna stick with that.

PR #2409, opened 2020-05-22
Refactoring close all tabs functions

Dranaxel:
I modified the function so only one is used but i think there are some other modifications needed

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2331,35 +2331,26 @@ export async function tabclose(...indexes: string[]) {
     return done
 }
 
-/** Close all tabs to the right of the current one
+/** Close all tabs to the side specified
  *
  */
 //#background
-export async function tabclosealltoright() {
+export async function tabcloseallto(side) {
     const tabs = await browser.tabs.query({
         pinned: false,
         currentWindow: true,
     })
 
     const atab = await activeTab()
+    if (side == "right") {
     const ids = tabs.filter(tab => tab.index > atab.index).map(tab => tab.id)
-    return browser.tabs.remove(ids)
-}
-
-/** Close all tabs to the left of the current one
- *
- */
-//#background
-export async function tabclosealltoleft() {
-    const tabs = await browser.tabs.query({
-        pinned: false,
-        currentWindow: true,
-    })
-
-    const atab = await activeTab()
+	}
+	else if (side == "left"){
     const ids = tabs.filter(tab => tab.index < atab.index).map(tab => tab.id)
+	}
     return browser.tabs.remove(ids)
 }
+
 
 /** Restore the most recently closed item.
     The default behaviour is to restore the most recently closed tab in the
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -2331,36 +2331,27 @@ export async function tabclose(...indexes: string[]) {
     return done
 }
 
-/** Close all tabs to the right of the current one
+/** Close all tabs to the side specified
  *
  */
 //#background
-export async function tabclosealltoright() {
+export async function tabcloseallto(side) {
```
bovine3dom, line 2338:
```suggestion
export async function tabcloseallto(side: string) {
```

You need to give this a type so our python preprocessor knows what to convert its arguments to.

Dranaxel: Thank you, I saw it in the tests details but will this modification be enough ? I didn't reviewed all the code but i assume the tabcloseall functions are called somewhere else.

src/excmds.ts
```diff
@@ -2331,36 +2331,27 @@ export async function tabclose(...indexes: string[]) {
...
  */
 //#background
-export async function tabclosealltoright() {
+export async function tabcloseallto(side: string) {
     const tabs = await browser.tabs.query({
         pinned: false,
         currentWindow: true,
     })
 
     const atab = await activeTab()
```
bovine3dom, line 2344: `let ids`

src/excmds.ts
```diff
@@ -2331,36 +2331,27 @@ export async function tabclose(...indexes: string[]) {
...
+    if (side == "right") {
     const ids = tabs.filter(tab => tab.index > atab.index).map(tab => tab.id)
```
bovine3dom, line 2346: No const on either of these

src/excmds.ts
```diff
@@ -2331,36 +2331,27 @@ export async function tabclose(...indexes: string[]) {
     return done
 }
 
-/** Close all tabs to the right of the current one
+/** Close all tabs to the side specified
  *
  */
 //#background
-export async function tabclosealltoright() {
```
bovine3dom, line 2338: To reinstate these commands, in src/lib/config.ts, under exaliases, add `tabclosealltoright: "tabcloseallto right", tabclosealltoleft: "tabcloseallto left"`.

src/excmds.ts
```diff
@@ -2331,36 +2331,27 @@ export async function tabclose(...indexes: string[]) {
...
+export async function tabcloseallto(side: string) {
     const tabs = await browser.tabs.query({
         pinned: false,
         currentWindow: true,
     })
 
     const atab = await activeTab()
+    if (side == "right") {
```
bovine3dom, line 2345: Instead of this `if` statement I'd probably do something like this to further reduce duplication:

```js
if !(["left","right"]).includes(side)) throw("side must be left or right")
const comp = side == "right" ? tab => tab.index > atab.index : tab => tab.index < atab.index
const ids = tabs.filter(comp).map(tab => tab.id)
```

Issue #2432, opened 2020-05-27
Switch to GitHub actions from circle ci


bovine3dom:
This means people will be able to view the logs without logging in to circle ci.

bovine3dom:
Should be able to steal most of this: https://github.com/glacambre/firenvim/blob/master/.github/workflows/test.yml

bovine3dom:
Added in #4769

PR #2434, opened 2020-05-27
Add GitHub action for e2e tests

bovine3dom:
Presumably this won't work as it stands as it relies on `bash`, which probably isn't available on Windows.

Need to add the other three jobs, too: mozilla, linting and the unit tests - https://github.com/tridactyl/tridactyl/blob/4276cd68e968a4da7314e700a8e90ef3facefb00/.circleci/config.yml#L16

Related issue: #2432


bovine3dom:
It looks like Selenium isn't finding Firefox.

PR #2442, opened 2020-06-01
Enable tridactyl to run in chrome

MatiasStorm:
Resolves #1683


scripts/build.sh
```diff
@@ -4,6 +4,12 @@ set -e
 
 CLEANSLATE="node_modules/cleanslate/docs/files/cleanslate.css"
 TRIDACTYL_LOGO="src/static/logo/Tridactyl_64px.png"
+BROWSER="firefox"
+
+if [ $1 = "chrome" ]; then
```
bovine3dom, line 9: If called with no arguments, we get `scripts/build.sh: line 9: [: =: unary operator expected` here

bovine3dom:
See https://github.com/tridactyl/tridactyl/commit/3ff84e61661b955e478f523437cf73144e8ff1d5 for how to fix the CopyWebpackPlugin errors - when you merged master in, you'll have updated the plugin to a version with a new API.

bovine3dom:
Lots of Tridactyl seems to work fine on the "options" page. That's really cool!

What you said in the previous issue has worried me a little bit:

> I will now try to find all of the places where tridactyl fails in chrome and make issues for them. 

I can pretty much guarantee that you won't be able to find all of the places where Tridactyl doesn't work in Chrome - Tridactyl is just too big and has a massive number of configuration options.

Even if you did find everything that worked in Chrome but not Firefox, we'll almost certainly add something in the future that does break Tridactyl in Chrome.

I think it would be more productive to focus on allowing/encouraging Tridactyl developers to access `browser.` in a way that is safe across the browsers. So, e.g. `:tab ` completions, rather than just not working at all in Chrome because Chrome doesn't have container support, would not show container icons in Chrome.

Web extension API compatibility across browsers is available programmatically here https://github.com/mdn/browser-compat-data/tree/master/webextensions/api - I don't know if we could develop something that used that with TypeScript to warn us when we hadn't dealt with a property that might not exist. I think that's my ideal situation: if I tried to use `browser.firefoxOnlyProperty.function`, I should get a compiler error unless I used `browser.firefoxOnlyProperty?.function` and dealt with the possibility that it might be undefined.

I think @saulrh had some ideas about this, so I'm pinging him here :)

src/content/finding.ts
```diff
@@ -72,6 +72,12 @@ class FindHighlight extends HTMLSpanElement {
 
 }
 
+// TODO: Conditionally import customElements-polyfill needed for the chrome browser:
+if (!customElements) {
+    // tslint:disable-next-line
+    require("@webcomponents/custom-elements");
+}
```
bovine3dom, line 79:
```suggestion
```

I don't understand why this is needed on your machine. `customElements` is defined in Chrome. Tridactyl works fine on the options page on Chrome if I remove these lines. customElements have been in Chrome since version 54 - https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry#bcd:api.CustomElementRegistry - which was released in 2016.

MatiasStorm: I just tried to remove the lines again, and build with `yarn run build:chrome` and run with `yarn run run:chrome`, and i still get the error when i open a new tab and go to google.
Here is what it looks like:
Screenshot

My operating system is Ubuntu 18.04 and the chrome version is 83.

bovine3dom: Ah, I see. I hadn't spotted that Tridactyl was working (a little bit) on pages that weren't the options page. Odd that Chrome doesn't support customElements in web extensions!

package.json
```diff
@@ -47,12 +48,15 @@
...
     "web-ext-types": "^3.2.1",
+    "webextension-polyfill-for-webpack": "^0.0.1",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.11"
   },
   "scripts": {
     "build": "sh scripts/build.sh",
+    "build:chrome": "sh scripts/build.sh chrome",
     "run": "web-ext run -s build/ -u 'txti.es'",
+    "run:chrome": "web-ext run -t chromium -s build/ -u 'txti.es'",
```
saulrh, line 59: Might be worth adding `-t firefox-desktop` to the `run` command for clarity.

scripts/build.sh
```diff
@@ -4,6 +4,14 @@ set -e
 
 CLEANSLATE="node_modules/cleanslate/docs/files/cleanslate.css"
 TRIDACTYL_LOGO="src/static/logo/Tridactyl_64px.png"
+BROWSER="firefox"
+
+if [ -n "$1" ]; then
+  if [ "$1" = "chrome" ]; then
+    BROWSER="chrome"
+  fi
```
saulrh, line 12: Is there a reason to not just set `BROWSER = $1`?

saulrh:
> I think @saulrh had some ideas about this, so I'm pinging him here :)

Let me go see if I left any notes for myself...

[A comprehensive list of known incompatibilities](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Chrome_incompatibilities) maintained by Mozilla. It's mostly intended for chrome->ff ports, but it should help in both directions. It shouldn't be too hard to audit tridactyl's codebase for the slightly more subtle violations.

> I don't know if we could develop something that used that with TypeScript to warn us when we hadn't dealt with a property that might not exist.

Should be... vaguely possible to write a script that reads that and spits out a `.d.ts` that overrides the `browser` declarations with optionals in the right places? You'd lose a lot of the fancy type-checking that you get from `web-ext-types`, so it'd have to go in some kind of separate linting step or something. Not great.

I'm surprised that `web-ext-types` doesn't already do the right thing wrt parts of the spec that aren't implemented in some places.

Huh. Actually, looks like someone has already done this! https://github.com/Lusito/webextension-polyfill-ts

saulrh:
How do we want to structure the docs for tridactyl-on-chromium? Do we have them off on a separate page? Do we document inconsistencies inside specific commands?

We should probably think about designing some way to annotate `excmds` as working/not working on different platforms so we can raise error messages like "This is a known bug" rather than getting issue tracker spam.

bovine3dom:
I think I'd like one source of documentation for all versions of Tridactyl (partly so Chrome users can see what they're missing). We should probably be able to annotate ex-cmds as missing so that they do not show up in completions.

Ideally whatever we choose will be general enough to make supporting different Firefox versions (e.g. ESR) / Firefox for Android not too much extra work.

scripts/build.sh
```diff
@@ -4,6 +4,14 @@ set -e
 
 CLEANSLATE="node_modules/cleanslate/docs/files/cleanslate.css"
 TRIDACTYL_LOGO="src/static/logo/Tridactyl_64px.png"
+BROWSER="firefox"
+
+if [ -n "$1" ]; then
```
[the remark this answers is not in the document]
glacambre: `[[` isn't posix and build.sh uses `/bin/sh`, so I think the single brackets should stay.

bovine3dom:
I haven't had time to check that the polyfill you linked to does this, saulrh, but https://github.com/Lusito/forget-me-not/blob/develop/src/background/background.ts#L113 is exactly the sort of thing I want to compiler to persuade us to do :) (except with optional chaining as that's much more ergonomic).

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "sourceMap": true,
     "target": "es2017",
     "lib": ["es2017","es2019.array", "es2018.promise", "dom", "dom.iterable"],
-    "types": ["web-ext-types"],
+    "types": ["web-ext-types", "node"],
```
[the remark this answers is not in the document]
MatiasStorm: Yes this is only for the `require`, when i try to use `import` the compiler says that `import` can only be used in a namespace or module.
I also considered using dynamic import but that required me to set the `--module` flag. I tried setting it to `commonjs` which breaks everything, don't know that much about it so there might be some other value i can set it to, to make it work?

scripts/build.sh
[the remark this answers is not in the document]
MatiasStorm: I have updated `build.sh` to use `getopt`.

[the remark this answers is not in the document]
saulrh: Oops, didn't notice that we were using `/bin/sh`. >_>

tsconfig.json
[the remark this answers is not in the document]
saulrh: Dunno if there'd be a different setting that'd work. Looks fine to me, though maybe with a comment saying that the type import is for a `require` and if we get the polyfill working better we can ditch it.

package.json
saulrh: I like the explicit `run:firefox` target. :+1:

MatiasStorm:
Can this be merged into master or do i need to make more changes?

saulrh:
>  Merge branch 'master' into chrome-port 
Note for @MatiasStorm: I generally recommend trying to rebase branches instead of merging master into them. It's not necessary (or possibly even worth the time) in cases like this, but for higher-complexity projects it makes it a lot easier to follow other recommendations about clarity of history.

As an example of "history clarity recommendations": When I have a feature I'd like to add, the last thing I do before sending off a pull request is go back through my changes and re-group them into a totally new series of commits - sometimes even a series of independent pull requests - based on logical function rather than the coincidence of what order I wrote the changes in. This makes it so that the code can be reviewed, debugged, bisected, and rolled back [in the smallest possible units](https://google.github.io/eng-practices/review/developer/small-cls.html). This is substantially easier to do when I've been rebasing my changes onto main instead of merging main into my branch.

MatiasStorm:
Thank you @saulrh, that is noted!

webpack.config.js
```diff
@@ -1,63 +1,67 @@
...
+            help: "./src/help.ts",
+            newtab: "./src/newtab.ts",
+        },
+        output: {
+            filename: "[name].js",
+            path: __dirname + "/build",
+        },
+
+        // Enable sourcemaps for debugging webpack's output.
+        devtool: "source-map",
```
bovine3dom, line 43:
```suggestion
        devtool: "source-map",
        // devtool: "inline-source-map", // Uncomment me for more helpful error messages
```

bovine3dom:
> Can this be merged into master or do i need to make more changes?

One minor change - which I'm happy to do as I merge it - is that the `yarn.lock` file should be committed. That helps to ensure that our builds are reproducible as it is a record of all of the exact versions of dependencies used. I've just noticed that our CI would pass even if the lockfile changed on `yarn install` - I've fixed that on master now.

The other issue is that I'm slightly uncomfortable with merging this as it stands as the Chrome port doesn't work very well, so no-one will use it (yet). I don't want unused code to stay in Tridactyl for a long time. OTOH, I appreciate that un-merged code gets stale and is annoying to fix tp.

So, I'll merge this if you pinky-promise to tell me if you decide to abandon the Chrome port so that I then know to un-merge it :) . I am bad at noticing these things on my own (see https://github.com/tridactyl/tridactyl/commit/10b00b4f6849d5ef0093ebc63138a367c359f36c which survived unused in Tridactyl for a whole year).

MatiasStorm:
I have added `yarn.lock`.

I didn't really think about it that way but now that you say it, it might be wiser to wait merging this. At least until #2460 has been resolved.

bovine3dom:
Blocked by #2460 :)

saulrh:
I'd be willing to switch to Chrome as my daily driver until we have some users migrated, if that helps.

Marin-Kitagawa:
Hey,
I just want to know how the progress for the issue is. I really want to use Tridactyl in Chromium-based browsers. Is there a way to do it now? I can see the label as blocked. When shall we expect this?

asknet:
I'm curious to know if there is still a path forward to enable Tridactyl to run in Chrome. Thank you!

bovine3dom:
Yeah, pretty much what stupidcomputer said:

- For Manifest V2, we can port to Chromium pretty easily, we just need to play whack-a-mole with the linter, adding guards around Firefox only features such as containers. But Chrome is not accepting new Manifest V2 extensions, so this would not be very useful to people who want a public release.
- For Manifest V3, we would need to do a rewrite of Tridactyl's architecture. In theory we'll have to do that anyway eventually because Firefox wants to move to V3, but I am telling myself that Mozilla will not make the switch for a few years. Google used Manifest V3 as an excuse to tighten restrictions on code that changes at run-time, meaning things like `:js` would probably have to be disabled.

PR #2445, opened 2020-06-02
Experiment with webextension-polyfill-ts

bovine3dom:
I'd hoped that this would give me an error: `browser.contextualIdentities is undefined` since we're building for Chrome which has no such API. (It complains about "IdentityIcon" but I think that's just something that is missing from the type definitions - it doesn't complain about the useful stuff).

It doesn't, so I'm not sure it's fantastically useful for helping us to port to Chrome. @saulrh does this mean we have to roll our own, or am I misunderstanding how to use the package?

Edit: this follows discussion in #2442.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,7 +10,7 @@ module.exports = {
     "ts-jest": {
       tsConfig: {
         ...tsConfig.compilerOptions,
-        types: ["jest", "node", "web-ext-types"]
+        types: ["jest", "node"]
       },
       diagnostics: {
         ignoreCodes: [151001]
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
   },
   "scripts": {
     "build": "sh scripts/build.sh",
-    "run": "web-ext run -s build/ -u 'txti.es'",
+    "run": "web-ext run -s build/ -u 'txti.es' -t chromium",
     "forrest-run": "yarn run run",
     "watch": "echo 'watch is broken, use build instead'; exit 0;",
     "clean": "rm -rf build generated",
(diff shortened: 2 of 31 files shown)


saulrh:
Ah, okay, it doesn't handle that thing in particular because `contextualIdentities` isn't even part of the webextension standard, it's a totally mozilla-only thing. So, yeah, probably not useful the way I'd thought it'd be useful. Bah. :/

bovine3dom:
I wonder how hard it would be to make. It seems like a similar problem to Babel for targeting specific browsers and versions.

saulrh:
~~Well, the mozilla interface support data that you pointed out does have everything on `browser` rather than just the web-extensions stuff, so that'd be suitable.~~

~~The two big tasks that I see with the general approach:~~
1. ~~Compile and maintain a mapping from the compatibility data to the web-ext-types typings.~~
2. ~~Either:~~
  a. ~~Produce a modified version of web-ext-types that incorporates `Nullable[]` wherever an interface isn't present on one of our target browsers.~~
  b. ~~Produce and run type-checking against multiple modified versions of web-ext-types each incorporating a different browser's~~

Nevermind, I forgot about the interaction with the webextension polyfill. That'd be super gnarly - we'd have to produce not just a modified version of `web-ext-types`, but a modified version of `web-ext-types` ∪ `webextension polyfill types` that has nullables in the appropriate places. Ick.

It might be better to go through and do this by hand in a `.d.ts` file somewhere. :/

bovine3dom:
I'm not sure I see the issue. If we edited web-ext-types so that anything that wasn't in all browsers was the union of the real type and undefined, wouldn't that be good enough? Ideally with types based on the exact target.

Why do we have to care about the polyfill? I thought all it did was swap browser for chrome and change promises into callbacks.

PR #2449, opened 2020-06-03
introduce excmds unit tests

csabahenk:


bovine3dom:
Unit tests should be in the src/ directory :)

As we discussed on Riot it's probably worth renaming tests/excmds.test.ts to e2e.tests.ts as testing the ex-cmds is only a small part of what it does.

PR #2455, opened 2020-06-04
Autocommands for webrequests

### LINKED ISSUE #1885: WebRequest autocmd integration
- Want to allow users to replicate `set csp clobber` if they wish
- and expose as much WebRequest functionality as possible

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest

Ideally we'd integrate this with autocommands: `autocmd [HeadersReceived|BeforeRequest|etc] [url] [function]`.

There is a slight snag in that the ex-commands have to be synchronous for most events. We'll probably need to add jssync / jsbsync commands to make it usable, then `command mylistener jsbsync -p return {}` could work. We'd also have to make new functionality such that we could synchronously call synchronous ex-commands (e.g. add a sync version of acceptExCmd in src/lib/controller.ts).

bovine3dom:
Probably needs more thought, testing and documentation before merging.

Fixes #1885.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -25,6 +25,7 @@ import { AutoContain } from "@src/lib/autocontainers"
 import * as extension_info from "@src/lib/extension_info"
 import * as omnibox from "@src/background/omnibox"
 import * as R from "ramda"
+import * as webrequests from "@src/background/webrequests"
 
 // Add various useful modules to the window for debugging
 ;(window as any).tri = Object.assign(Object.create(null), {
@@ -41,6 +42,7 @@ import * as R from "ramda"
     request,
     state,
     webext,
+    webrequests,
     l: prom => prom.then(console.log).catch(console.error),
     contentLocation: window.location,
     R,
@@ -96,7 +98,7 @@ browser.runtime.onStartup.addListener(_ => {
         } else {
             native.run("hostname").then(hostname => {
                 for (const host of hosts) {
-                    if ((new RegExp(host)).exec(hostname.content)) {
+                    if (new RegExp(host).exec(hostname.content)) {
                         controller.acceptExCmd(aucmds[host])
                     }
                 }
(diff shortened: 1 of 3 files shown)


src/background.ts
```diff
@@ -138,6 +138,31 @@ browser.tabs.onActivated.addListener(ev => {
...
+    "Completed",
+    "ErrorOccured",
+    "HeadersReceived",
+    "ResponseStarted",
+    "SendHeaders",
+]) {
+    config.getAsync("autocmds", requestEvent).then(aucmds => {
+        if (!aucmds) return
+        const patterns = Object.keys(aucmds)
+        for (const pattern of patterns) {
```
bovine3dom, line 156: This should be moved out into a `registerWebRequestAutocmd` so that we can add a config listener that calls it (on a single event + pattern) every time a webrequest auto command is added.

It would be nice if we could deal with removal of web request aucmds too.

bovine3dom:
Almost ready to merge. One last issue to chase down: you have to `:autocmddelete` a webrequest autocmd _twice_ before it is actually removed.

src/background.ts
```diff
@@ -136,6 +138,57 @@ browser.tabs.onActivated.addListener(ev => {
...
+        R.mergeDeepWith(
+            (l, r) => (l == r ? undefined : r),
+            ...R.map(R.pickAll(webrequests.requestEvents), [a, b]),
+        ),
+    )
+
+    // Rather than thinking this through, exchange programmer time for user time
+    // and exploit the fact that deleted entries in b are the same as entries in a,
+    // so do the weirdMerge twice
+    const merged = weirdMerge(previous, weirdMerge(previous,current))
```
bovine3dom, line 170: Oh, obviously this doesn't work because I then swap all the undefineds in the RHS for any not-undefineds in the LHS.

Note to self: do less programming when tired

src/background.ts
```diff
@@ -136,6 +138,57 @@ browser.tabs.onActivated.addListener(ev => {
...
+    })
+}
+
+config.addChangeListener("autocmds", (previous, current) => {
+    // Unchanged entries become undefined
+    // New entries in b are left in place
+    // Deleted entries in b are replaced by ones in a
+    const weirdMerge = (a, b) => Object.entries(
+        R.mergeDeepWith(
+            (l, r) => (l == r ? undefined : r),
```
bovine3dom, line 162: Solution for the below - set to `null` instead?

Issue #2460, opened 2020-06-05
Check browser compatibility in tests


bovine3dom:
@MatiasStorm mentioned https://github.com/amilajack/eslint-plugin-compat on Riot. It might need some work before it works with TypeScript - https://github.com/amilajack/eslint-plugin-compat/issues/313.

Under the hood, it uses https://github.com/mdn/browser-compat-data which does have details on WebExtension compatibility, so that aspect might work out of the box. 

~~"Blocked" tag is because this approach relies on us switching to eslint first: #2363~~

mozbugbox:
There was an `addons-linter` from mozilla:

  https://github.com/mozilla/addons-linter

    Add-ons Linter - the Add-ons Linter is used on
    addons.mozilla.org and the web-ext tool. It uses
    browser-compat-data to check that the Firefox
    version that the add-on lists support for does in
    fact support the APIs used by the add-on.

and a ` webextension-polyfill` too:

  https://github.com/mozilla/webextension-polyfill

    If you choose to write your extension to use browser
    and promises, then Firefox also provides a polyfill
    that will enable it to run in Chrome:
    https://github.com/mozilla/webextension-polyfill.

bovine3dom:
Thanks @mozbugbox, we use

PR #2477, opened 2020-06-11
ESlint migration

### LINKED ISSUE #2363: Migrate from tslint to eslint
tslint has been deprecated as @noctuid reminded me in #2359.

MatiasStorm:
Resolves #2363.

I have managed to get `eslint` to work with pretty much all of the same rules as before.
However when I run `$(yarn bin)/eslint . --ext .ts` i get 3365 problems most of them concerning the use of any and missing return types.
I just wanted to know, before wasting time fixing them, if I have configured something the wrong way, since they have always been there? Or if you would prefer them to be fixed?
@bovine3dom


bovine3dom:
Could you change the lint CI job here https://github.com/tridactyl/tridactyl/blob/master/.circleci/config.yml to use eslint rather than tslint? That way I can see what the errors are exactly.

I'd rather not have 3k lines of code change as that will make merging things a nightmare.

What we could do is make `yarn run pretty` use stricter rules than the lint job, which would allow us to fix things while we make changes to the code.

MatiasStorm:
I have changed the CI job.

From what I can see `yarn run pretty` executes the `pretty.sh` script which runs tslint on all staged files. When you say you want stricter rules would that be for tslint, or should tslint be changed to eslint in `pretty.sh`?

bovine3dom:
Yes, provided that it has a --fix option like tslint.

I'm in the middle of moving so please forgive me if I'm a bit less responsive for the next few days : )

MatiasStorm:
I have changed `pretty.sh` to use eslint with the `--fix` option, such that we only use `eslint` for changed files.

I also changed `ci/circleci`'s linting back to only using tslint. 
Also, would it be a good idea to make  `ci/circleci` run eslint on changed files in PR's? If it is possible?

bovine3dom:
Yeah, that's possible and probably a good idea. I'd add another job, e.g. `eslint_migration` to circleci.

You'd need to adapt this script https://github.com/tridactyl/tridactyl/blob/0527c7496e027aca99da0a96aedd770b8a104c5d/scripts/common.sh#L6, swapping all the instances of `--cached` with `HEAD^` if it is called in that job (but leaving them as --cached for the pretty script).

bovine3dom:
I am wondering if this piecemeal approach to migrating is a good idea after all. It would probably make porting to Chrome harder as the compatibility linter won't check all of Tridactyl, which is what we need it to do.

So maybe it's better to run --fix on all of Tridactyl? What do you think?

bovine3dom:
I think this approach makes more sense than my original suggestion:

1. rules which Tridactyl failed have been turned off
2. circleci and pretty both use eslint
3. rather than trying to fix ~3k lines of code in one PR, we can slowly turn the rules back on (made very obvious by the comments in .eslintrc.js) with about one PR per enabled rule.
4. as we re-enable the rules we should convert the `/* tslint` comments into ones that eslint will accept.
5. I'll merge this PR as soon as I'm happy that `yarn run pretty` works OK.

This means that you can get on with the Chrome port by adding #2460 which would then apply to the whole code-base, just like when we turn more rules on for point 3.

Sorry for changing my mind :)

MatiasStorm:
No worries!
It sounds like a much better solution :)

I will start working on making `eslint-plugin-compat` support `typescript-eslint`, don't think it has been done yet.

scripts/common.sh
```diff
@@ -29,16 +29,35 @@ prettierUgly() {
...
 
-tslintUgly() {
+eslintUgly() {
     local acc=""
     local IFS=$'\n'
     local tmpdir
-    tmpdir=$(mktemp -d "tslint.XXXXXXXXX")
+    tmpdir=$(mktemp -d "eslint.XXXXXXXXX")
+
+    # TODO: need to add this directory to a tsconfig.json
```
bovine3dom, line 38: The pre-commit hook is broken currently - it always says changed files are broken.

We need to either add the temporary directory to the/a tsconfig programmatically once it is created or create a different temporary directory in the source tree that can be picked up by tsconfig (e.g. ./.tmp/ with tsconfig including ./.tmp/**ts)

PR #2534, opened 2020-06-24
Hint discrimination

### LINKED ISSUE #2265: JS hints should be a different colour
Blocks #2259 

I suspect we'll have to change the array of elements to hint in content/hinting.ts to an object of types of elements to arrays of elements to hint. That way we could extend it in the future if we wanted.

I am hoping we can use CSS arithmetic to automatically change the colour on other themes (probably just reduce saturation?).

bovine3dom:
Fixes #2265, unblocks #2259

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -375,11 +375,13 @@ class HintState {
 /** @hidden*/
 let modeState: HintState
 
+interface Hintables { elements: Element[]; style: string }
+
 /** For each hintable element, add a hint
  * @hidden
  * */
 export function hintPage(
-    hintableElements: Element[],
+    hintableElements: Hintables[],
     onSelect: HintSelectedCallback,
     resolve = () => {}, // eslint-disable-line @typescript-eslint/no-empty-function
     reject = () => {}, // eslint-disable-line @typescript-eslint/no-empty-function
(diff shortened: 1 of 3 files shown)


src/content/hinting.ts
```diff
@@ -630,7 +637,11 @@ class Hint {
         this.flag.style.cssText = `
             top: ${window.scrollY + top}px !important;
             left: ${window.scrollX + left}px !important;
+            ${extrastyle}
```
glacambre, line 640: When I thought about implementing this I was thinking more about adding extra "semantic" classes to the hints that themes could choose to display however they want. Are there issues with this approach?

bovine3dom:
Thoughts on colours for the js hints, esp for non-default themes, appreciated.

glacambre:
I think "[js yellow](https://www.google.com/search?q=js&tbm=isch)" might be a good choice.

bovine3dom:
I think it's a bit too hard to read:



vs current

glacambre:
The text would probably need to be black to have an acceptable level of 
contrast. Then we'd have to use black for regular hints too unless we 
want things to be ugly, and this would require using a lighter red. 
People would probably complain about this change.

We could also try cyan since that's the opposite of red but that'd be 
ugly and white on cyan is just as hard to read. I guess white on grey is 
the least bad option :) (hopefully people won't infer that grey means 
disabled).

PR #2546, opened 2020-06-28
Wait for completion filtering to finish before running next

bovine3dom:
This should improve e.g. `tft<Tab>` resulting in ft.com being selected rather than `<Tab>` firing too early and then being deselected.

It doesn't seem to make much difference in practice, suggesting that a lot of the work is being done outside of `filter` and isn't being awaited.

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -53,7 +53,9 @@ const logger = new Logger("cmdline")
 /** @hidden **/
 const commandline_state = {
     activeCompletions: undefined,
-    clInput: (window.document.getElementById("tridactyl-input") as HTMLInputElement),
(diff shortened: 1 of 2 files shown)


src/commandline_frame.ts
```diff
@@ -53,7 +53,9 @@ const logger = new Logger("cmdline")
 /** @hidden **/
 const commandline_state = {
     activeCompletions: undefined,
-    clInput: (window.document.getElementById("tridactyl-input") as HTMLInputElement),
+    clInput: window.document.getElementById(
```
bovine3dom, line 56: I'm sure we used to only run `prettier` on lines that had changed. This is annoying (and my fault, IIRC).

PR #2551, opened 2020-06-28
Fix #2548: don't rename hints when filtering by hintchars

### LINKED ISSUE #2548: `vimperator-reflow` hints behave very oddly
1. ~~`:set hintnames numeric`~~ not needed: `set hintchars 123457890`
2. `:set hintfiltermode vimperator-reflow`
3. `:open news.bbc.co.uk`
4. `f` - BBC in the top right is number 1
5. `1` - Your account is now number 1

Related: #2544

bovine3dom:
Fixes #2548.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -905,9 +905,6 @@ function filterHintsVimperator(query: string, reflow = false) {
     // Start with all hints
     let active = modeState.hints
 
-    // If we're reflowing, the names may be wrong at this point, so apply the original names.
-    if (reflow) rename(active)
-
     // Filter down (renaming as required)
     for (const run of partitionquery(query)) {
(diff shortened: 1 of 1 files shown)


src/content/hinting.ts
```diff
@@ -905,11 +905,8 @@ function filterHintsVimperator(fstr, reflow = false) {
     // Start with all hints
     let active = modeState.hints
 
-    // If we're reflowing, the names may be wrong at this point, so apply the original names.
-    if (reflow) rename(active)
```
bovine3dom, line 909: @cmcaine do you know why you thought the names might be wrong? We now hint JS-event hints separately so it is no longer (this) trivial to re-apply the original names.

In my limited testing it works fine without this line.

PR #2569, opened 2020-07-02
Add TST integration: mk III

bovine3dom:
The old PR had gone stale. This rebase + squash fixes that. Untested.


src/excmds.ts
```diff
@@ -737,6 +738,54 @@ export async function mktridactylrc(...args: string[]) {
     return conf
 }
 
+/** Writes current config to a file.
```
bovine3dom, line 741: I must have introduced this when merging / rebasing / squashing.

src/lib/objects.ts
```diff
@@ -0,0 +1,14 @@
+/** @hidden
+ * Merges two objects and any child objects they may have
+ */
+export default function mergeDeep(o1, o2) {
```
bovine3dom, line 4: lib/config should use this function

bovine3dom:
Todo: give saulrh and previous contributors some credit in the git log

bovine3dom:
I think this needs quite a bit of work - `interop/tst.ts` needs to be usable from the content script for any TST information to be used in completions (which, as I understand it, is a large part of what TST users want). `doTstIntegration` in particular needs to remain fast.

alexandru0-dev:
@bovine3dom what's the status of this pr?
it's abandoned but was asking if there is any chances that this will eventually restart as it's a feature i would like a lot
and would like to contribute if possible

bovine3dom:
I think I gave up with it because the stuff that it could do was not something anyone particularly wanted. The stuff that people wanted it to do (displaying the tree in `:tab ` completions) was much harder to implement and have it run quickly

I think I am telling myself that I will do it when the completions are rewritten as per #4709

PR #2577, opened 2020-07-04
Add escapehatch ex-command and global keybind for it

bovine3dom:
Todo: double check that this doesn't count as a new permission


src/excmds.ts
```diff
@@ -2931,6 +2932,51 @@ export async function shellescape(...quoteme: string[]) {
...
+ */
+//#background
+export async function escapehatch() {
+    const tabs = await browser.tabs.query({ currentWindow: true })
+    const tridactyl_tabs: browser.tabs.Tab[] = []
+    await Promise.all(
+        tabs.map(async tab => {
+            try {
+                // This doesn't actually return "true" like it is supposed to
+                await Messaging.messageTab(tab.id, "alive")
```
bovine3dom, line 2952: @farnoy I think `Messaging` is still broken - I expected this to return true but it just returns `undefined`

PR #2579, opened 2020-07-05
Add (slightly lame) user configurable browser wide binds

### LINKED ISSUE #2576: Ctrl-6 bind via commands API
It'd be really nice if this was accessible everywhere. Unclear if we can override the "choose the 6th tab" bind but worth a try.

bovine3dom:
Fixes #2576

Todo:

- [x] unbind
- [x] implement mapstrToMozMap in bind
- [x] docs (probably room for improvement, esp to explain how to unbind Ctrl+6 and the background ex-command weirdness)
- [ ] read from config and set `browser.commands.update` at startup to ensure sync between browsers?

Limitations:
- can have at most N binds active at the same time
- can only use background ex-strings on inaccessible pages which do not internally call content ex-commands - not sure how to tell users which is which other than "use common sense" + "trial and error" (edit: after more thought this isn't true - it's ex-commands that are and only call background ex-commands unless you know that you are on a tab where Tridactyl can run and then they can call content ex-commands)
- binds must be Mozilla/Chromium approved - need one or two modifiers (no more, no less) and a reasonably vanilla key.
- if users change a bind in "Manage extension shortcuts", behaviour is undefined

### DIFF
--- a/src/lib/binding.ts
+++ b/src/lib/binding.ts
@@ -50,16 +50,5 @@ export function parse_bind_args(...args: string[]): bind_args {
 
     result.excmd = args.join(" ")
 
-    // On second thought - I don't think we need this as browser.commands.update() will throw an error for us
-    //
-    // Check we have a Mozilla-compatible sequence for Commands binds
-    // TODO: check that the key is compatible too
-    // NB: +!! is a crazy way of coercing a boolean | undefined to a number
-    // if (result.mode == "browser" && (
-    //     keyseq.length > 1  ||
-    //     [+!!keyseq[0].ctrlKey, +!!keyseq[0].altKey, +!!keyseq[0].shiftKey, +!!keyseq[0].metaKey].reduce((l,r) => Number(l) + Number(r)) < 1 ||
-    //     [+!!keyseq[0].ctrlKey, +!!keyseq[0].altKey, +!!keyseq[0].shiftKey, +!!keyseq[0].metaKey].reduce((l,r) => Number(l) + Number(r)) > 2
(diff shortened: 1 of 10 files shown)


src/lib/config.ts
```diff
@@ -383,6 +383,10 @@ export class default_config {
         "<Space>": "hint.selectFocusedHint",
     }
 
+    browsermaps = {
+        "<C-,>": "escapehatch",
```
bovine3dom, line 387: Tempted to include a Ctrl-6 and Ctrl-^ (is that allowed?) bind here too, and remove it from the other modes.

src/lib/binding.ts
```diff
@@ -45,11 +45,21 @@ export function parse_bind_args(...args: string[]): bind_args {
...
+    //
+    // Check we have a Mozilla-compatible sequence for Commands binds
+    // TODO: check that the key is compatible too
+    // NB: +!! is a crazy way of coercing a boolean | undefined to a number
+    // if (result.mode == "browser" && (
+    //     keyseq.length > 1  ||
+    //     [+!!keyseq[0].ctrlKey, +!!keyseq[0].altKey, +!!keyseq[0].shiftKey, +!!keyseq[0].metaKey].reduce((l,r) => Number(l) + Number(r)) < 1 ||
+    //     [+!!keyseq[0].ctrlKey, +!!keyseq[0].altKey, +!!keyseq[0].shiftKey, +!!keyseq[0].metaKey].reduce((l,r) => Number(l) + Number(r)) > 2
+    // )) throw "Error: Incompatible key sequence specified for 'browser' mode binds. Use one or two modifiers with a single key, e.g. <C-.>"
+
```
bovine3dom, line 63:
```suggestion
```

bovine3dom:
Closing in favour of #2582

PR #2591, opened 2020-07-07
Feature/home ci

Dranaxel:
I created hidden scripts reproducing the circleci workflow, then we call them all with circleci script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,6 @@ commands:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
             - ./node_modules
-      - run: bash -c 'yarn run build'
 jobs:
   lint:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ commands:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
             - ./node_modules
-      - run: yarn run build
+      - run: bash -c 'yarn run build'
 jobs:
   lint:
     docker:
@@ -20,21 +20,19 @@ jobs:
       - run: sudo apt update
       - run: sudo apt install -qq shellcheck
       - commoncmd
-      - run: bash -c 'GLOBIGNORE="node_modules" shellcheck -e2012 **/*.sh'
-      - run: yarn run lint
-      - run: bash -c '"$(yarn bin)/eslint" src --ext .ts'
+      - run: bash -c './ci/lint.sh'
   unit:
     docker:
       - image: circleci/node:latest
     steps:
       - commoncmd
-      - run: bash -c '"$(yarn bin)/jest" src'
+      - run: bash -c './ci/unit.sh'
   mozilla:
     docker:
       - image: circleci/node:latest
     steps:
       - commoncmd
-      - run: bash -c 'cd build; "$(yarn bin)/web-ext" lint'
+      - run: bash -c 'ci/mozilla.sh'
 
 workflows:
   version: 2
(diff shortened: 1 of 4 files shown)


bovine3dom:
This is great, thanks!

I think I'd like the scripts to be more visible. I would suggest putting them in a new folder in the root of the project, e.g. `ci/`, so that it's obvious to people when a CI job has failed which script to run to replicate it on their machine.

Also - `.circleci/config.yml` doesn't seem to have any changes to it, so CircleCI isn't actually using these scripts. My intention was that it should - otherwise they'll inevitably fall out of synchronisation when someone updates the job but forgets to update the script.

Dranaxel:
Ok, I understand so you want circleci to call every script in each steps instead of a multi-line ?

bovine3dom:
e.g. https://github.com/tridactyl/tridactyl/blob/f047e5c895eda2f4cc2c9feb071ee518cacbd3cf/.circleci/config.yml#L23 and the three lines below it would become `bash -c "ci/lint.sh"`

all the setup - installing files etc. - should remain in `.circleci/config.yml`.

ci/commoncmd.sh
```diff
@@ -1,3 +1,2 @@
 #! /bin/sh
-yarn install --frozen-lockfile
 yarn run build
```
bovine3dom, line 2: I don't think you need to build stuff every time - IIRC only the mozilla job needs it

bovine3dom:
I think you just need to do `chmod +x` on the scripts :)

ci/lint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+SCRIPT_PATH=${0%/*}
+if [ "$SCRIPT_PATH" = "$0" ]; then
+	cd ..
+fi
+./ci/commoncmd.sh
+yarn install --frozen-lockfile
```
bovine3dom, line 7:
```suggestion
./ci/commoncmd.sh
yarn install --frozen-lockfile
```

I don't think you need either of these lines (certainly not the `yarn install` one as that's done in config.yml)

ci/lint.sh
```diff
@@ -0,0 +1,14 @@
...
+if ! [ -x "$(command -v shellcheck)" ]; then
+	GLOBIGNORE="node_modules" shellcheck -e2012 **/*.sh
+else
+	echo "You don't have shellcheck installed, we cannot validate shell scripts"
```
bovine3dom, line 11:
```suggestion
	echo "Warning: shellcheck is not installed, skipping shell scripts"
```

ci/lint.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+SCRIPT_PATH=${0%/*}
+if [ "$SCRIPT_PATH" = "$0" ]; then
+	cd ..
+fi
```
bovine3dom, line 5: This snippet is quite strange - it looks like you can only call it from `ci/` or the root folder.

Why not 

```suggestion
cd ${0%/*}/..
```
? IIRC that should always put you in the root folder.

.circleci/circle.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/sh
```
bovine3dom, line 1: I don't think this script works or is needed

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@ commands:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
             - ./node_modules
-      - run: yarn run build
+      - run: bash -c 'ci/commoncmd.sh'
```
bovine3dom, line 14: As I said below I don't think you need to build stuff every time, only for the mozilla job

PR #2606, opened 2020-07-10
Resurrect find mode completions

bovine3dom:
This needs quite a bit more work than I had anticipated - a lot of stuff that was added in e878b93fd was later binned.


src/completions/Find.ts
```diff
@@ -78,33 +88,31 @@ export class FindCompletionSource extends Completions.CompletionSourceFuse {
             incsearchonly = true
         }
 
-        // Note: the use of activeTabId here might break completions if the user starts searching for a pattern in a really big page and then switches to another tab.
-        // Getting the tabId should probably be done in the constructor but you can't have async constructors.
-        const tabId = await activeTabId()
-        const findings = await Messaging.messageTab(
```
bovine3dom, line 84: Ah, I see why this `messageTab` weirdness is here now - it is so that the iframe can access the content in https://github.com/tridactyl/tridactyl/blob/4885e30896a71a87600eb29531217924dc54a29d/src/content/finding.ts#L11

bovine3dom:
Related to #1048 - when we eventally add `/` as a default bind, we should also bind `<C-g/G>` to findnext / prev so that we don't punish ~4 years of muscle memory.

Issue #2597, opened 2020-07-08
Switch to clipboard API


bovine3dom:
https://blog.mozilla.org/addons/2018/08/31/extensions-in-firefox-63/#content-main

Less hacky so will hopefully fix #1239 and https://github.com/tridactyl/tridactyl/issues/1135.

We'd need to ask for the `clipboardRead` and `clipboardWrite` permissions, so we should do this the next time we ask for more permissions, e.g. https://github.com/tridactyl/tridactyl/pull/2359.

Rummskartoffel:
I started looking at this and noticed we have those permissions already, added by no-one other than yourself in c76d3a6 😜. I'll see if I can't get this clipboard API working.

PR #2617, opened 2020-07-14
<S-Del> closes highlighted tab in tab/taball commandline_frame

MyGitUsername:
Resolve Issue: https://github.com/tridactyl/tridactyl/issues/1815

`<S-Del>` in the "Tabs" an "All Tabs" buffer closes the highlighted tab and the commandline frame closes.  

The usefulness of this command is limited because if the users wishes to close multiple tabs in the Tab / All Tabs buffer, then the user needs to reopen the tab buffer (`b` or `B`) after each `<S-Del>` command.  

I can update the PR for the commandline frame to persist, if desirable.

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -194,7 +194,7 @@ commandline_state.clInput.addEventListener(
                 if (args.length === 0) {
                     commandline_state.fns[funcname]()
                 } else {
-                    commandline_state.fns[funcname](args)
+                    commandline_state.fns[funcname](args.join(" "))
                 }
 
                 prev_cmd_called_history = history_called
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -189,8 +189,14 @@ commandline_state.clInput.addEventListener(
             // This is definitely a hack. Should expand aliases with exmode, etc.
             // but this whole thing should be scrapped soon, so whatever.
             if (response.value.startsWith("ex.")) {
-                const funcname = response.value.slice(3)
-                commandline_state.fns[funcname]()
+                const [funcname, ...args] = response.value.slice(3).split(/\s+/)
+
+                if (args.length === 0) {
+                    commandline_state.fns[funcname]()
+                } else {
+                    commandline_state.fns[funcname](args)
+                }
+
                 prev_cmd_called_history = history_called
             } else {
                 // Send excmds directly to our own tab, which fixes the
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -91,12 +91,12 @@ function resizeArea() {
  * This is a bit loosely defined at the moment.
  * Should work so long as there's only one completion source per prefix.
  */
-function getCompletion() {
+function getCompletion(args_only = false) {
     if (!commandline_state.activeCompletions) return undefined
 
     for (const comp of commandline_state.activeCompletions) {
         if (comp.state === "normal" && comp.completion !== undefined) {
-            return comp.completion
+            return args_only ? comp.args : comp.completion
         }
     }
 }
--- a/src/completions/Tab.ts
+++ b/src/completions/Tab.ts
@@ -4,6 +4,7 @@ import { enumerate } from "@src/lib/itertools"
 import * as Containers from "@src/lib/containers"
 import * as Completions from "@src/completions"
 import * as config from "@src/lib/config"
+import * as Messaging from "@src/lib/messaging"
 
 class BufferCompletionOption extends Completions.CompletionOptionHTML
     implements Completions.CompletionOptionFuse {
@@ -67,10 +68,11 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
             "BufferCompletionSource",
             "Tabs",
         )
-
         this.sortScoredOptions = true
         this.updateOptions()
         this._parent.appendChild(this.node)
+
+        Messaging.addListener("tab_changes", () => this.reactToTabRemove())
     }
 
     async onInput(exstr) {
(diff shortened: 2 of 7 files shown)


bovine3dom:
This is great, much nicer than I had expected!

We should probably change the `<C-o>yy` bind to use this once merged.

I _think_ the command line should probably persist rather than being closed. What do you think?

src/lib/commandline_cmds.ts
```diff
@@ -155,6 +155,39 @@ export function getCommandlineFns(cmdline_state) {
...
+            }
+
+            // Create new cmd to execute excmd with the completion
+            // args, discarding the completion func
+            const cmdToExec = excmd + " " + args.join(" ")
+
+            // Save non-secret commandlines to the history.
+            if (
+                !browser.extension.inIncognitoContext &&
+                !(excmd === "winopen" && args[0] === "-private")
```
bovine3dom, line 180: I'm wondering how this deals with `composite winopen -private` - I think usually we just discard anything that contains "winopen -private" but I might be wrong.

src/commandline_frame.ts
```diff
@@ -189,8 +189,14 @@ commandline_state.clInput.addEventListener(
...
             // but this whole thing should be scrapped soon, so whatever.
             if (response.value.startsWith("ex.")) {
-                const funcname = response.value.slice(3)
-                commandline_state.fns[funcname]()
+                const [funcname, ...args] = response.value.slice(3).split(/\s+/)
+
+                if (args.length === 0) {
+                    commandline_state.fns[funcname]()
+                } else {
+                    commandline_state.fns[funcname](args)
```
bovine3dom, line 197:
```suggestion
                    commandline_state.fns[funcname](...args)
```

If I try to rebind `<C-o>yy` to this:

```patch
index f26a0575..cd14523d 100644
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -120,7 +120,7 @@ export class default_config {
         "<S-Tab>": "ex.prev_completion",
         "<Space>": "ex.insert_space_or_completion",

-        "<C-o>yy": "ex.copy_completion",
+        "<C-o>yy": "ex.execute_ex_on_completion clipboard yank",
     }

     /**
```

I get `# controller while accepting: Error: Not an excmd: clipboard,yank` when using it. 

I think this is probably the source of the problem but I haven't tested it :)

MyGitUsername: Good catch.  This should be fixed in the latest commit.

src/lib/commandline_cmds.ts
MyGitUsername: > I _think_ the command line should probably persist rather than being closed. What do you think?

Yes, I agree.  Can you please advise regarding the best approach to implement a persistent cmd line?  It seems that even without `cmdline_state.fns.hide_and_clear()`, the cmdline hides on `return ownMessageTab()`.

Also, I noticed <S-Del> highlighted tab is not working yet in the All Tabs buffer (B).  I will fix this in the next commit.

Thanks for the feedback and help.

bovine3dom: I'm honestly not sure what we should do with `accept_line`. We should have the same behaviour everywhere. That probably means we should have a special "shouldStoreExString" function that they both use. I'm trying to work out if it's more annoying to have e.g. a search query that includes winopen -private disappear or have a sensitive composite command stored.

I think as part of this PR the current behaviour should be kept but the logic should be separated into its own function in e.g. src/lib/, so that if we change our minds we remember to change it everywhere. Sorry for changing my mind :)

I'll look into the command line hiding.

src/lib/commandline_cmds.ts
```diff
@@ -155,6 +155,42 @@ export function getCommandlineFns(cmdline_state) {
             return messageOwnTab("controller_content", "acceptExCmd", [command])
         },
 
+        execute_ex_on_completion: (excmd: string) => {
+            const command = cmdline_state.getCompletion()
+
+            cmdline_state.fns.hide_and_clear()
```
bovine3dom, line 161: So you're saying even without this line the command line hides?

MyGitUsername: Interestingly, without line 161 `cmdline_state.fns.hide_and_clear()`,  the command line hides in the case of `<C-o>yy` but not in the case of `<S-Del>`.  Hiding the command line upon ` <C-o>yy` is consistent with the previous behavior.

Another consideration: even when the cmdline persists after deleting a tab from the "Tabs" or "Tabs All" buffer with `<S-Del>`, the deleted tab is not dynamically removed from the tab list.  To see the updated list of tabs, I exit the cmd mode and press 'b' again.  I'm working on fixing this, so that the tab buffer dynamically updates the tab list if a tab is deleted.

bovine3dom: There's two possible causes for that:

- we're using `clipboard yank` rather than `yank` which gives feedback via the command line
- we use the command line to copy text - see #2597 

For the update, you could either:
- make composite work with `ex.` so we can call refresh_completions (or whatever it is called) or
- add a listener that updates the completions as soon as the tabs change

bovine3dom: The listener is a little tricky now that I think about it, as the events are fired in the background script but our completions run in the content script.

I think I'd recommend the following:

1. For each of the tab events we care about here, https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs#Events in src/background.ts register a listener with them
2. That listener should send a message to all the tabs telling them that the tabs have been updated - I think we have the `messageAllTabs` function for this
3. Listen for that message in src/completions/Tab{,All}.ts and update accordingly. For a first pass I'd ignore the content of the message and just re-run the normal update process whenever a message about a tabs update is received; we can optimise later if we need to.

---

That is slightly involved, though, so it might be better to merge this as-is and make a new issue for that. What do you think?

src/lib/commandline_cmds.ts
MyGitUsername: The latest commit adds an a "store_ex_string" function to avoid code duplication, making it easier to change the logic if we change our minds.

src/lib/commandline_cmds.ts
MyGitUsername: Sounds good to make a new issue to add the tab event listener.

I did add another commit to move the duplicate logic that checks if we should store the ex string command in the command history to its own function.

src/completions/Tab.ts
```diff
@@ -238,4 +240,32 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
         }
         return this.updateDisplay()
     }
+
+    /**
+     * Update the list of possible tab options and select (focus on)
```
bovine3dom, line 245: This makes me uneasy, I think it could cause confusion.

I would store the current options, update the options, then do a diff between them to see if the focused option was deleted, then focus the one after (I think this makes more sense as you are pressing delete and not backspace). If the option was not deleted focus should be kept on it, even if it has moved.

Ramda is probably a useful library for this; we use it in a few other places.

If that's too much work I'm happy to look at it myself :)

MyGitUsername: Sounds like a plan - I can make these changes.

MyGitUsername: Sorry for any inconvenience with the delay here - I pushed an update to address the above concerns.

bovine3dom:
I'll try to review this today or tomorrow :)

PR #2619, opened 2020-07-14
fix(autocontain): Fix part of RegExp that matches subdomains.

aureooms:
Otherwise `autocontain abc.de ...` will match `xyabc.de`. Bug or feature?

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3503,7 +3503,7 @@ export function autocmd(event: string, url: string, ...excmd: string[]) {
  * Automatically open a domain and all its subdomains in a specified container.
  *
  * This function accepts a `-u` flag to treat the pattern as a URL rather than a domain.
- * For example: `autocontain -u ^https?://[^/]*youtube\.com/ google` is equivalent to `autocontain youtube\.com google`
+ * For example: `autocontain -u ^https?://([^/]*\.|)youtube\.com/ google` is equivalent to `autocontain youtube\.com google`
  *
  * For declaring containers that do not yet exist, consider using `auconcreatecontainer true` in your tridactylrc.
  * This allows tridactyl to automatically create containers from your autocontain directives. Note that they will be random icons and colors.
@@ -3531,7 +3531,7 @@ export function autocontain(...args: string[]) {
     let [pattern, container] = args
 
     if (!urlMode) {
-        pattern = `^https?://[^/]*${pattern}/`
+        pattern = `^https?://([^/]*\\.|)${pattern}/`
     }
 
     return config.set("autocontain", pattern, container)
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -3531,7 +3531,7 @@ export function autocontain(...args: string[]) {
     let [pattern, container] = args
 
     if (!urlMode) {
-        pattern = `^https?://[^/]*${pattern}/`
+        pattern = `^https?://([^/]*\\.|)${pattern}/`
```
bovine3dom, line 3534:
```suggestion
        pattern = `^https?://([^/]*\.|)${pattern}/`
```
It's either wrong here and in config.ts or wrong in the docstring :)

bovine3dom:
In theory I think this is a good idea.

I'm not sure what the behaviour is if a site matches two autocontain patterns (which is what would happen if this was merged and someone ran `autocontain` in their RC file). I foresee two possibilities: 1) Tridactyl explodes or 2) one autocontain takes precedence over the other, leading to confusion when one tries to update the autocontain container later on. So I/we need to look at that before merging it.

It might be safer to add a new flag that uses this new, sane regex and quietly deprecate the old one.

src/excmds.ts
aureooms: Two backslashes are needed in JavaScript strings. I do not know if the tridactyl config parser escapes/unescapes content.

```
Welcome to Node.js v14.5.0.
Type ".help" for more information.
> new RegExp('\.').test('a')
true
> new RegExp('\\.').test('a')
false
> new RegExp('\\.').test('.')
true
> new RegExp(`\.`).test('a')
true
> new RegExp(`\\.`).test('a')
false
```

aureooms:
> I'm not sure what the behaviour is if a site matches two autocontain patterns (which is what would happen if this was merged and someone ran `autocontain` in their RC file). I foresee two possibilities: 1) Tridactyl explodes or 2) one autocontain takes precedence over the other, leading to confusion when one tries to update the autocontain container later on. So I/we need to look at that before merging it.

Currently if multiple matches occur no container is assigned. See below https://github.com/tridactyl/tridactyl/blob/b40c0ea24250b9c1d47feb20b0e89534f6e5a5e1/src/lib/autocontainers.ts#L277

> It might be safer to add a new flag that uses this new, sane regex and quietly deprecate the old one.

Yes.

src/excmds.ts
bovine3dom: Eugh, that's annoying. Our docs are wrong in lots of places, then. We don't do anything to escape things.

bovine3dom:
Closing in favour of #2629 (I managed to mess up pushing here somehow). Thanks :)

Issue #2636, opened 2020-07-17
`<C-6>` on Windows appears to fire twice


bovine3dom:
Tridactyl version: 1.19.1pre4292-8937866c
Firefox version: Mozilla Firefox 78.0.2
Operating system: win

Could be specific to my profile. I `unset {nmaps, imaps, inputmaps, exmaps}` and don't appear to have any Ctrl-6 binds other than the browser one.

You see the previous tab flash and you return to the current tab.

Ctrl-Shift-6 appears to work fine.

I haven't had time to test it in a fresh profile yet.

bovine3dom:
It's a Firefox bug. Firefox doesn't back off from Ctrl-6 being switch-to-sixth-tab if it's set as a browser bind (unlike if a website asks it to preventDefault which it does quite happily).

Shipping a different default browser bind per platform is fairly easy (although we'd then need to work out how to have OS specific config), but we'll then break the muscle memory of lots of users.

I think the best approach would be to leave it unbound for Windows users and somehow have the `<C-6>` n+lots of others+map still appear for Windows users, which again requires OS-specific default config.

This is a can of worms. :( 🥫🐍🐍🐍 

(I'm worried about telling Mozilla about the bug as they might decide that we can't bind to ctrl-6 anywhere, which would make me very sad).

PR #2664, opened 2020-07-25
Add platform specific defaults

bovine3dom:
For dealing with platform-specific bugs like #2636

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -18,6 +18,7 @@
  */
 import * as R from "ramda"
 import * as binding from "@src/lib/binding"
+import * as platform from "@src/lib/platform"
 
 /* Remove all nulls from objects recursively
  * NB: also applies to arrays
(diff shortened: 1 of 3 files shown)


src/lib/platform.ts
```diff
@@ -0,0 +1,15 @@
...
+// Will return undefined if it can't work it out
+export function getPlatformOs(): browser.runtime.PlatformOs {
+    const platform = navigator.platform
+    const mapping = {
+        "win":  "Windows",
+        "openbsd": "BSD",
+        "mac": "Mac",
+        "linux": "Linux",
+    }
+    return R.keys(R.filter(x=>platform.includes(x), mapping))[0] as keyof typeof mapping
```
bovine3dom, line 14: There must be a neater way to do this in Ramda

src/lib/config.ts
```diff
@@ -1059,8 +1060,23 @@ export class default_config {
...
+
+
+/** @hidden
+ * Merges two objects and removes all keys with null values at all levels
+ */
+export const mergeDeepCull = R.pipe(mergeDeep, removeNull)
+
 /** @hidden */
-export const DEFAULTS = o(new default_config())
+export const DEFAULTS = platform.getPlatformOs() == "win" ? mergeDeepCull(o(new default_config()), windows_defaults) : o(new default_config())
```
bovine3dom, line 1079: Should probably make this more extensible. The truly sane thing to do would be to have a `set --platform` setting like `set --url` (i.e. `:seturl` as it actually exists), but I'm not sure I can be bothered.

PR #2679, opened 2020-07-29
Increase readline conformity of editor functions

### LINKED ISSUE #2678: The word related editor functions are not fully readline compliant
Tridactyl version: 1.19.1pre4342-134bc4d1
Firefox version: Mozilla Firefox 78.0.2
Operating system: linux

It's documented in `src/lib/editor.ts` that the `(backward|forward)_word` and `(backward_)?kill_word` functions are intended to mimic `readline` behaviour. However, their current behaviours don't match what one expects from readline in a few non-intuitive ways.

Assuming `set wordpattern [^\/\.]`, and denoting cursor with pipe (|):

The `(forward|backward)_word` variants leap twice instead of expected just once:

`backward_word`:
```
text     => https://developer.moz|illa.org/

expected => https://developer.|mozilla.org/
actual   => https://|developer.mozilla.org/
```

`forward_word`:
```
text     => https://developer.moz|illa.org/

expected => https://developer.mozilla|.org/
actual   => https://developer.mozilla.|org/

------------------

text     => https://developer.mozilla|.org/

expected => https://developer.mozilla.org|/
actual   => https://developer.mozilla.|org/
```

For the `kill_word` variants, aside from the word itself `readline` also cleans up leading/trailing non-wordpattern chars but tridactyl leaves them:

`backward_kill_word`:

```
text     => https://developer.|mozilla.org/

expected => https://|mozilla.org/
actual   => https://|.mozilla.org/
```

`kill_word` also deletes one more char at boundary2 unlike readline:

```
text     => https:|//developer.mozilla.org/

expected => https:|.mozilla.org/
actual   => https://|mozilla.org/
```

Additionally, the `getWordBoundaries` function in `src/lib/editor_utils.ts` searches in the reverse direction, when no match is found in the main direction, this can lead to some funky behaviours.

natrys:
The kill_word variants now cleanup trailing/leading non-word chars, and
don't kill anything in reverse direction if no word found in main
direction.

The word movement functions now only leap once instead of twice.

---

bovine3dom edit: fixes #2678

### DIFF
--- a/src/lib/editor.ts
+++ b/src/lib/editor.ts
@@ -285,15 +285,18 @@ export const kill_whole_line = wrap_input(
 export const kill_word = wrap_input(
     needs_text((text, selectionStart, selectionEnd) => {
         const boundaries = getWordBoundaries(text, selectionStart, false)
-        if (selectionStart > boundaries[0] && selectionStart < boundaries[1])
+        if (selectionStart < boundaries[1]) {
             boundaries[0] = selectionStart
-        // Remove everything between the newline and the caret
-        return [
-            text.substring(0, boundaries[0]) +
-                text.substring(boundaries[1] + 1),
-            boundaries[0],
-            null,
-        ]
+            // Remove everything between the newline and the caret
+            return [
+                text.substring(0, boundaries[0]) +
+                    text.substring(boundaries[1]),
+                boundaries[0],
+                null,
+            ]
+        } else {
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks for this! Does this fix #2678 completely?

natrys:
Yes, I think so.

PR #2693, opened 2020-08-01
Use a MutationObserver to set up event handlers in content frames

farnoy:
This allows tridactyl to catch events in nested document trees,
like those in Google Docs for example. Protecting slashes is also
done in every nested document tree now.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -53,24 +53,12 @@ messaging.addListener("alive", async () => true)
 
 // Hook the keyboard up to the controller
 import * as ContentController from "@src/content/controller_content"
-import { getAllDocumentFrames } from "@src/lib/dom"
 
 const guardedAcceptKey = (keyevent: KeyboardEvent) => {
     if (!keyevent.isTrusted) return
     ContentController.acceptKey(keyevent)
 }
 function listen(elem) {
-    elem.removeEventListener("keydown", guardedAcceptKey, true)
-    elem.removeEventListener(
-        "keypress",
-        ContentController.canceller.cancelKeyPress,
-        true,
-    )
-    elem.removeEventListener(
-        "keyup",
-        ContentController.canceller.cancelKeyUp,
-        true,
-    )
     elem.addEventListener("keydown", guardedAcceptKey, true)
     elem.addEventListener(
         "keypress",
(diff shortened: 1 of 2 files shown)


src/content.ts
```diff
@@ -53,24 +53,12 @@ messaging.addListener("alive", async () => true)
...
-    elem.removeEventListener(
-        "keypress",
-        ContentController.canceller.cancelKeyPress,
-        true,
-    )
-    elem.removeEventListener(
-        "keyup",
-        ContentController.canceller.cancelKeyUp,
-        true,
-    )
```
farnoy, line 73: These are added again below, in the same exact form, so removing them first is not necessary.

src/content.ts
```diff
@@ -84,9 +72,24 @@ function listen(elem) {
...
+                    listen(node.contentDocument);
+                    observer.observe(node.contentDocument, {subtree: true, childList: true})
+                    if (leavegithubalone !== "true")
+                        node.contentDocument.body.addEventListener("keydown", protectSlash)
+                }
+    }
+}
+
+const iframeMutationObserver = new MutationObserver(iframeMutationCallback)
+iframeMutationObserver.observe(document, {subtree: true, childList: true})
```
farnoy, line 92: I'm not sure what guarantees Firefox makes about the order of execution. This observer is set up as soon as navigation happens, which from my brief experiments is enough to consistently catch all changes done from loading the actual page. It might not be bulletproof though, especially if we start using top level async-await.

src/content.ts
```diff
@@ -84,9 +72,24 @@ function listen(elem) {
     )
 }
 listen(window)
-document.addEventListener("readystatechange", _ =>
-    getAllDocumentFrames().forEach(f => listen(f)),
```
farnoy, line 88: This registered event handlers on `<iframe>` elements, which are still a part of the same document tree as root. I don't think they ever captured anything.

src/content.ts
```diff
@@ -84,9 +72,24 @@ function listen(elem) {
...
+
+async function iframeMutationCallback(mutations: MutationRecord[], observer: MutationObserver) {
+    const leavegithubalone = await config.getAsync("leavegithubalone")
+
+    for (const mutation of mutations) {
+        if (mutation.type === "childList")
+            for (const node of mutation.addedNodes)
+                if ((node instanceof HTMLIFrameElement) || (node instanceof HTMLFrameElement)) {
+                    listen(node.contentDocument);
+                    observer.observe(node.contentDocument, {subtree: true, childList: true})
```
farnoy, line 84: This sets up event listeners on the inner document tree, separate from the parent. It also registers the observer in that document tree for nested iframes. It potentially has the same problem as the root observer - it needs to run before other things. From my testing, this observer is registered as soon as the `<iframe>` tag is added, so before the frame actually loads, but it may be imperfect.

src/content.ts
```diff
@@ -84,9 +72,24 @@ function listen(elem) {
     )
 }
 listen(window)
-document.addEventListener("readystatechange", _ =>
-    getAllDocumentFrames().forEach(f => listen(f)),
-)
+
+async function iframeMutationCallback(mutations: MutationRecord[], observer: MutationObserver) {
+    const leavegithubalone = await config.getAsync("leavegithubalone")
```
farnoy, line 77: I don't like that this function is async right now. The ideal thing would be to process all mutations synchronously and observe nested trees right away. Since `leavegithubalone` is not something that changes often, maybe we could cache it somewhere?

bovine3dom: You can use config.get if you want, that is synchronous. getAsync is merely more likely to be right :)

bovine3dom:
@farnoy could you explain in more detail what bug this fixes?

On Matrix, you said

> This PR will fix scrolling j/k in google docs for example, and will protect slash in iframes too (if configured by the user)

But it doesn't seem to work for me. `:unfocus` doesn't work either. Could you provide detailed steps to reproduce the bug in Tridactyl without this patch?

Thanks :)

PR #2726, opened 2020-08-10
Fix source_quiet parameter

xoi:
`source_quiet` function parameter does not match the command arguments.
This causes `tridactylrc` loading at `TriStart` to fail.

The involved commit is 4781ac1.


bovine3dom:
Wow, I don't know how this went unnoticed for two months. Thanks!

PR #2841, opened 2020-09-24
Fix #2839: parse keys iff user config loaded

### LINKED ISSUE #2839: Default `bind d tabclose` resurrects itself over my `bind d scrollpage 0.5`
Tridactyl version: 1.20.1pre4591-6d0810f2
Firefox version: Mozilla Firefox 81.0
Operating system: linux

Sometimes while using tridactyl, pressing key `d` closes a tab i was working on instead of scrolling down, per my updated bindings. It happens only sometimes, and so far I've only seen it:
1. after closing other tab, if I press `d` very quickly when the focus is being moved to another tab
2. after opening a new tab, but in that short time before the focus is moved to it, pressing `d`
3. when a tab is generally loading, pressing `d`

Maybe important info is that I'm also running TST, Stylish and Violentmonkey which slow down responsiveness of my 50+ tabs multi-window firefox, especially on github which I experience in a customized way. That leaves some room for this bug to manifest, as the page is loading slower.

*   Steps to reproduce:

1. have `bind d scrollpage 0.5` in your `tridactylrc`
2. open two tabs
3. on one tab navigate to another link
4. QUICK! press `d` repeatedly! pressss ittt!
5. if the tab is still alive, navigate back `<C-[>` per my config, and quickly press `d`
6. keep on navigating back and forward that one tab and very quickly press `d` while it's loading...
7. voila! the old binding that was not supposed to be used has been applied

https://github.com/tridactyl/tridactyl/issues or anywhere really.. haven't tested with some special pages, but tridactyl doesn't work on special pages anyways i guess..

https://gist.github.com/maricn/a86720274072a9899d747524e8aacd84/71904ac60cd3e1577ba09eb7d06facea8397ecb9 (rev. 4)

*   Contents of ~/.tridactylrc or ~/.config/tridactyl/tridactylrc (if they exist):

```
sanitise tridactyllocal tridactylsync

" General Settings
set configversion 1.9
set update.lastchecktime 1593077346454
set smoothscroll true
set modeindicatorshowkeys true
set modeindicator false

" Theme
" set theme maricn-base16-gruvbox-dark-hard
" set colours maricn-base16-gruvbox-dark-hard
colourscheme maricn-base16-gruvbox-dark-hard

" Binds
unbind <F1>
unbind <C-f>
unbind <C-a>
unbind <C-d>
unbind <C-u>
unbind .
unbind p
unbind s
unbind S
unbind D
unbind U

bind d scrollpage 0.5
bind u scrollpage -0.5

" conflicts with <C-d> as "Bookmark this page"
" bind <C-d> scrollpage 0.7
" bind <C-u> scrollpage -0.7

bind <C-[> back
bind <C-]> forward

bind i mode ignore
bind --mode=ignore <escape> mode normal

bind t tabnew
bind J tabnext
bind K tabprev
bind x tabclose
bind --mode=ignore ^ tab #

bind / fillcmdline find
bind n findnext
bind N findnext -1
bind <Escape> composite mode normal ; hidecmdline ; nohlsearch

bind --mode=ex <ArrowRight> ex.insert_completion

" For syntax highlighting see https://github.com/tridactyl/vim-tridactyl
" vim: set filetype=tridactyl
```

bovine3dom:
@maricn could you see if this PR fixes your issue? The bind should just not do anything until your config has loaded. I can't replicate the issue so I'm not 100% sure I've fixed it.

1. clone this branch
2. `yarn run install` and `yarn run build`
3. In Firefox, go to about:addons, disable Tridactyl
4. Go to about:debugging, click "This Firefox", load temporary addon, choose build/manifest.json
5. On a tab Tridactyl can access run `:source` and wait a bit
6. Try to reproduce #2839 

Thanks : )

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -37,7 +37,7 @@ const CONFIGNAME = "userconfig"
 /** @hidden */
 const WAITERS = []
 /** @hidden */
-let INITIALISED = false
+export let INITIALISED = false
 
 /** @hidden */
 // make a naked object
--- a/src/lib/keyseq.ts
+++ b/src/lib/keyseq.ts
@@ -418,6 +418,9 @@ export function translateKeysInPlace(keys, conf): void {
 export function keyMap(conf): KeyMap {
     if (KEYMAP_CACHE[conf]) return KEYMAP_CACHE[conf]
 
+    // Fail silently and pass keys through to page if Tridactyl hasn't loaded yet
+    if (!config.INITIALISED) return new Map()
+
     const mapobj: { [keyseq: string]: string } = config.get(conf)
     if (mapobj === undefined)
         throw new Error(


maricn:
This branch solves the issue. But if i got you right, you wouldn't merge this to mainline?

I double checked against the current beta release since I had a minor update of RC file, and the issue is present there, but not on this branch with the fix.

PR #2843, opened 2020-09-24
Fix #2810: run shellcheck if it is installed

### LINKED ISSUE #2810: CI seems to think shellcheck isn't installed
https://app.circleci.com/pipelines/github/tridactyl/tridactyl/2561/workflows/e45b74e9-083c-4fd7-8f1b-02a804fe3590/jobs/8354/parallel-runs/0/steps/0-107

bovine3dom:
I suspect #!/bin/sh doesn't support `command -v` or something. Need to check that ci/lint actually runs shellcheck for real.

### DIFF
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 cd ${0%/*}/..
-if ! [ -x "$(command -v shellcheck)" ]; then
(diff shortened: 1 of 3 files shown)


bovine3dom:
This hasn't worked. Further investigation needed - will probably need to print out the output of `which` and maybe `ls` the paths in path.

Issue #2776, opened 2020-08-31
bmarks command opens as search rather than matching bookmark


nwilleyltk:
*   Brief description of the problem:  bmarks command opens search page instead of matching bookmark

*   Steps to reproduce:

    1. Type :bmarks to invoke the bmarks command
    2. Search for a known bookmark  
    3. When it is the only bookmark shown, type the enter key
    4. On my computer the duckduckgo search page opens with my search term instead of the bookmark

*   URL of the website the bug happens on: N/A

nwilleyltk:
[Config as Gist](https://gist.github.com/nwilleyltk/0e12418e6fe958a6aa6465150943d595)

bovine3dom:
Thanks for telling us about this bug :)

We simply need to preselect a bookmark as we do with `:tabs`.

bovine3dom:
This was fixed #2863 :)

Issue #1370, opened 2019-02-18
Select first item by default in open/tabopen/etc


bpmcl:
Recently migrated from Vimium, and really liking Tridactyl so far, but I have one small configuration option I would like to suggest.

Most of the time when I am using functions like `open` and `tabopen`, I know exactly what I am looking for, so I can type a couple of letters and the first item in the list is what I wanted. Rather than having to hit Tab to select it, an option to have it selected by default would be useful, instead of always performing a web search by default.

(Additionally, being able to navigate up and down the list with the arrow keys would be nice.)

bovine3dom:
Up and down you can do - `bind --mode=ex <Tab>` will tell you what to bind to, then `bind --mode=ex <[...I can't remember what the up arrow is called, maybe ArrowUp... It's listed on MDN somewhere]> [the thing you just found]`. Repeat for `<S-Tab>` and whatever the down arrow is called.

We can definitely do the first one - that's how buffer behaves - but I think it would annoy most people as space inserts the highlighted completion, so you wouldn't be able to do `tabopen why do people use emacs`.

bpmcl:
I don't think it should be the default setting for everyone, but a configuration option to allow it to be customised would be useful.

Thanks for the help on the arrow key binding 👍

bovine3dom:
No it's not possible yet - we have `:set completions.History.autoselect true` _but_ the code for that isn't in History completions yet. It works for bookmarks, tabs, RSS and undo completions though.

It would not be hard to add.

https://github.com/tridactyl/tridactyl/blob/ced62d5fca6ea5aa986975cf5cf1765844498363/src/completions/Rss.ts#L35

something like that should go in History.ts

stevium:
@bovine3dom On the other hand, is there a way to show *currently opened* tab as selected one when typing `:tab` ? I almost always find myself looking for adjacent tabs to the currently opened one, and being able to press `ArrowUp` or `ArrowDown` to get to the start or end of the list doesn't really help much :)

stevium:
Oh wow, that was exactly what I was looking for! Works like a charm!

PR #2856, opened 2020-09-29
Add a default bookmark completion if none is selected

Trixxo:
Issue [#2776](https://github.com/tridactyl/tridactyl/issues/2776).
This selects the topmost option from the bookmark list as a default. Space still works.

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -402,8 +402,9 @@ export function editor_function(fn_name, ...args) {
 
 import * as SELF from "@src/commandline_frame"
 Messaging.addListener("commandline_frame", Messaging.attributeCaller(SELF))
-
 import { getCommandlineFns } from "@src/lib/commandline_cmds"
+import { KeyEventLike } from "./lib/keyseq"
+
 commandline_state.fns = getCommandlineFns(commandline_state)
 Messaging.addListener(
     "commandline_cmd",
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -61,6 +61,7 @@ const commandline_state = {
     completionsDiv: window.document.getElementById("completions"),
     fns: undefined,
     getCompletion,
+    getdefaultCompletion,
     history,
     /** @hidden
      * This is to handle Escape key which, while the cmdline is focused,
@@ -101,6 +102,18 @@ function getCompletion(args_only = false) {
     }
 }
 commandline_state.getCompletion = getCompletion
+
+/** @hidden **/
+function getdefaultCompletion(args_only = false) {
+    if (!commandline_state.activeCompletions) return undefined
+
+    for (const comp of commandline_state.activeCompletions) {
+        if (comp.state === "normal" && comp.defaultCompletion !== undefined) {
+            return comp.defaultCompletion
+        }
+    }
+}
+commandline_state.getdefaultCompletion = getdefaultCompletion
 
 /** @hidden **/
 export function enableCompletions() {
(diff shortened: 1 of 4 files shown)


bovine3dom:
Thanks for this PR! It looks like you had to delve quite deep into Tridactyl : )

I'm heading off for the night but my first reaction to this is that there's more code than I was expecting. I _think_ it might be as simple as calling the right constructor with `shouldsetstatefromScore = true`, as we do for Tab and RSS completions: https://github.com/tridactyl/tridactyl/blob/5e8f94ff06bd5f143732f2dabd32b08dabf1e65a/src/completions/Tab.ts#L93

Disclaimer: things I write just before going to bed are wrong even more often than usual. I'll try to look in more detail tomorrow.

Trixxo:
I thought that it could be good if `insert_space_or_completion`, would be ignored if you don't actively select an option. This way you can still press space. If you can hint me towards a better way of achieving that or you would rather have the other functionality, i would try to implement it.

bovine3dom:
Oh, I hadn't thought about that at all. I'm really not sure which is better. This behaviour might confuse users - does the default completion get highlighted?

Could you explain your thinking a little more?

Trixxo:
It doesn't get highlighted. i saw you mentioning here [#1370](https://github.com/tridactyl/tridactyl/issues/1370#issuecomment-464824348), that a default selection could annoy users. I felt the same when testing this, but i understand that it happens less frequently when selecting bookmarks. I thought that being able to fill the commandline with the current completion should only happen when an option is actually selected.

bovine3dom:
Yeah. Maybe we could highlight it in grey to make it clear that it will be entered if nothing is selected?

I think we need to think/work on this a bit more before it can be merged. I'd like more of the code to be in completions rather than the command line I think.

Trixxo:
> Thanks for this PR! It looks like you had to delve quite deep into Tridactyl : )
> 
> I'm heading off for the night but my first reaction to this is that there's more code than I was expecting. I _think_ it might be as simple as calling the right constructor with `shouldsetstatefromScore = true`, as we do for Tab and RSS completions:
> 
> https://github.com/tridactyl/tridactyl/blob/5e8f94ff06bd5f143732f2dabd32b08dabf1e65a/src/completions/Tab.ts#L93
> 
> Disclaimer: things I write just before going to bed are wrong even more often than usual. I'll try to look in more detail tomorrow.

Should i just do a new PR for this?

bovine3dom:
No, I think you're right and that it would be annoying if space inserted the thing. Probably.

Let me think about it for a bit : )

Trixxo:
I already did the other version, so maybe it could be used for now :D

bovine3dom:
I'm thinking that maybe the right way to do this is to have a config setting of an array of excmds that get autoselected completions. Or maybe a completionselection object: `set completionselection.tabopen auto|manual|weak` where weak is the grey behaviour, manual means only if user presses tab, auto is the current `:tab` behaviour?

By all means open the PR, but I won't merge it until I've thought about it.

Thanks for your work on this, I really appreciate it : )

Trixxo:
Yes that makes sense. Thank you for the thoughtful consideration, I will see if i can help somewhere else.

bovine3dom:
Please do open that PR as it will save me from having to do some thinking : )

(Doubly so if it helps you get a nice T shirt - https://hacktoberfest.digitalocean.com/)

bovine3dom:
I've been thinking about this quite a bit; I think essentially I want completions to be rewritten almost completely - #2865 - and integrate the ideas from this PR then. That might take a long time so I can only say that I'll merge this (or something that implements equivalent functionality) someday. Thanks again for the PR - it really is useful : )

PR #2863, opened 2020-09-30
select first bookmark completion automatically

Trixxo:


bovine3dom:
This PR should be fairly easy to merge. I want it configurable such that `set completions.bmarks.autoselect true` enables it. It might mean that we need to rewrite a little bit of the completions but it shouldn't be too bad.

I'll aim to get it done sometime next week.

Issue #2883, opened 2020-10-08
[feature request] :undo tab variation that works exactly like <C-t> (= DOES NOT extend to other windows)


m2habert:
Tridactyl version: 1.20.2pre4649-496ab135
Firefox version: Mozilla Firefox 77.0.1
Operating system: mac

FFx's <C-t> ( <cmd-t> on mac ) reopens closed tabs from current focused windows but, whenever reaches its limits, it does nothing.

bovine3dom:
https://github.com/tridactyl/tridactyl/blob/5e8f94ff06bd5f143732f2dabd32b08dabf1e65a/src/excmds.ts#L2450 dead easy to add if anyone is interested : )

PR #2888, opened 2020-10-09
Add 'undo tab_strict' to only restore tabs previously open in current window

fluem:
Issue #2883

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2480,10 +2480,11 @@ export async function tabcloseallto(side: string) {
     current window unless the most recently closed item is a window.
 
     Supplying either "tab" or "window" as an argument will specifically only
-    restore an item of the specified type.
+    restore an item of the specified type. Supplying "tab_strict" only restores
+    tabs that were open in the current window.
 
     @param item
-        The type of item to restore. Valid inputs are "recent", "tab" and "window".
+        The type of item to restore. Valid inputs are "recent", "tab", "tab_strict" and "window".
     @return
         The tab or window id of the restored item. Returns -1 if no items are found.
  */
@@ -2495,6 +2496,16 @@ export async function undo(item = "recent"): Promise<number> {
     if (item === "tab") {
         const lastSession = sessions.find(s => {
             if (s.tab) return true
+        })
+        if (lastSession) {
+            browser.sessions.restore(lastSession.tab.sessionId)
+            return lastSession.tab.id
+        }
+    } else if (item === "tab_strict") {
+        const lastSession = sessions.find(s => {
+            if (s.tab && s.tab.windowId === current_win_id) {
+                return true
+            }
         })
         if (lastSession) {
             browser.sessions.restore(lastSession.tab.sessionId)
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2492,63 +2492,25 @@ export async function tabcloseallto(side: string) {
 export async function undo(item = "recent"): Promise<number> {
     const current_win_id: number = (await browser.windows.getCurrent()).id
     const sessions = await browser.sessions.getRecentlyClosed()
+    const reUndoType = /^tab|tab_strict|window|recent$/
 
-    if (item === "tab") {
-        const lastSession = sessions.find(s => {
-            if (s.tab) return true
-        })
-        if (lastSession) {
-            browser.sessions.restore(lastSession.tab.sessionId)
-            return lastSession.tab.id
-        }
-    } else if (item === "tab_strict") {
-        const lastSession = sessions.find(s => {
-            if (s.tab && s.tab.windowId === current_win_id) {
-                return true
-            }
-        })
-        if (lastSession) {
-            browser.sessions.restore(lastSession.tab.sessionId)
-            return lastSession.tab.id
-        }
-    } else if (item === "window") {
-        const lastSession = sessions.find(s => {
-            if (s.window) return true
-        })
-        if (lastSession) {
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -2500,6 +2501,16 @@ export async function undo(item = "recent"): Promise<number> {
             browser.sessions.restore(lastSession.tab.sessionId)
             return lastSession.tab.id
         }
+    } else if (item === "tab_strict") {
+        const lastSession = sessions.find(s => {
```
bovine3dom, line 2505: Thanks for this PR!

I think we have a bit too much code repeated here now. Would you mind refactoring it so that the if statement set a filter function (i.e. the s => bit) and then move the actual filtering and restoration out of the ifs?

If not I'll get round to it in a few days time : )

fluem: Agreed. I simplified it quite a lot. Were you thinking something along the line of this?

src/excmds.ts
```diff
@@ -2492,63 +2492,25 @@ export async function tabcloseallto(side: string) {
...
-    } else {
-        throw new Error(`[undo] Invalid argument: ${item}. Must be one of "tab", "window", "recent"`)
+    // Pick the first session object that is a window or a tab from this window ("recent"), a tab ("tab"), a tab
+    // from this window ("tab_strict"), a window ("window") or pick by sessionId.
+    const session = sessions.find(s =>
+        (item === "recent" && (s.window || (s.tab && s.tab.windowId === current_win_id))) ||
+        (item === "tab" && s.tab) ||
+        (item === "tab_strict" && s.tab && s.tab.windowId === current_win_id) ||
+        (item === "window" && s.window) ||
+        (!isNaN(parseInt(item, 10)) && (s.tab || s.window).sessionId === item)
```
bovine3dom, line 2508: rather than the Regex check above you can just throw an error here - if it is a NaN throw an error.

Slightly more elegant as if we add more options we only have to do it in one place.

fluem: Sorry, I can't see how that would work. We would still have to check for item==="tab", "window", etc. in case there were no recently closed sessions, wouldn't we?

bovine3dom: Oh, yeah, you would have to split the argument parsing logic from the actual filtering.

e.g. `predicate = item === "recent" ? s => (s.window || ... ) : item === "tab" ? s => ... : s => throw new Error(...)`

I'm on a phone and will be for the next couple of days, so please forgive my brevity : )

src/excmds.ts
```diff
@@ -2492,52 +2493,28 @@ export async function undo(item = "recent"): Promise<number> {
...
+            ? s => s.tab && s.tab.windowId === current_win_id
+            : item === "window"
+            ? s => s.window
+            : !isNaN(parseInt(item, 10))
+            ? s => (s.tab || s.window).sessionId === item
+            : s => {
+                  throw new Error("blah")
+              } // this won't throw an error if there isn't anything in the session list, but I don't think that matters
+
+    const session = sessions.find(predicate)
```
bovine3dom, line 2513: @fluem do you see what I meant now? : )

src/excmds.ts
```diff
@@ -2492,52 +2493,28 @@ export async function undo(item = "recent"): Promise<number> {
...
+            : item === "tab"
+            ? s => s.tab
+            : item === "tab_strict"
+            ? s => s.tab && s.tab.windowId === current_win_id
+            : item === "window"
+            ? s => s.window
+            : !isNaN(parseInt(item, 10))
+            ? s => (s.tab || s.window).sessionId === item
+            : s => {
+                  throw new Error("blah")
```
bovine3dom, line 2510: this needs a less stupid error message, I did this only for testing : )

bovine3dom (changes requested):

src/excmds.ts
fluem: Great, thanks. Now I understand the syntax.

bovine3dom: Oh, the ternary `x ? "this" : "that"` isn't what I meant - I meant shifting the argument parsing ("did the user ask for recent, tab_strict, etc.") out of the filter. It would have been totally fine to instead do

```
let predicate
if (item === "recent") predicate = s => ...
else if (item == ...
else throw new Error(...)
```

and, probably, it would have been better to do it that way since it would throw an error even if there weren't any sessions to restore.

fluem: I like that it does not throw an error, if there are no sessions to restore.

src/excmds.ts
fluem: I will throw in the existing error. How do I accept your requested changes, before doing more changes? Have not had much experience with collaboration on GitHub.

fluem: Ok, figured it out.

src/excmds.ts
```diff
@@ -2507,9 +2507,8 @@ export async function undo(item = "recent"): Promise<number> {
             : !isNaN(parseInt(item, 10))
             ? s => (s.tab || s.window).sessionId === item
             : s => {
-                  throw new Error("blah")
+                  throw new Error(`[undo] Invalid argument: ${item}. Must be one of "recent, "tab", "tab_strict" or "window"`)
```
bovine3dom, line 2510: this isn't quite true as you can do `:undo [sessionId]` (a number, accessible to humans via the undo completions). Too tired to word it better now : )

fluem: Clarified the error message.

bovine3dom (approved):

PR #2901, opened 2020-10-13
Optional autoselect

bovine3dom:
Unblocks #2863

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -982,6 +982,24 @@ export class default_config {
     bmarkweight = 100
 
     /**
+     * General completions options - NB: options are set according to our internal completion source name - see - `src/completions/[name].ts` in the Tridactyl source.
+     */
+    completions = {
+        Tab: {
+            /**
+             * Whether to automatically select the closest matching completion
+             */
+            autoselect: "true",
+        },
+        Rss: {
+            autoselect: "true",
+        },
+        Sessions: {
+            autoselect: "true",
+        },
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


src/lib/config.ts
```diff
@@ -981,6 +981,24 @@ export class default_config {
      */
     bmarkweight = 100
 
+    /**
+     * General completions options - NB: options are set according to our internal completion source name - see - `src/completions/[name].ts` in the Tridactyl source.
```
bovine3dom, line 985: I should probably add an "experimental" warning here, because I'm still toying with totally rewriting completions to use config rather than object-oriented-bad-smells, and then this would probably move to an excmd -> setting mapping rather than completion source -> setting jobbie. Or maybe it wouldn't, since we'll still want the concept of completion sources so that different excmds can use the same ones.

src/lib/config.ts
```diff
@@ -981,6 +981,24 @@ export class default_config {
...
+     */
+    completions = {
+        Tab: {
+            /**
+             * Whether to automatically select the closest matching completion
+             */
+            autoselect: "true",
+        },
+        Rss: {
```
bovine3dom, line 994: These were broken before. Presumably this change will upset someone

thedaviddelta:
Hey! Amazing project, really liking it!
Cloud you add this into every other completion category? Or even an option that enables this for each one? I'd really like to have automatic selection inside the finder as I have in Vim, specially in my most used one "open"/"tabopen".
Thanks!

Issue #2690, opened 2020-08-01
Hide mode indicator in ignore mode


ChocolateOverflow:
Tridactyl version: 1.19.1pre3928-53d0d189
Firefox version: Mozilla Firefox 78.0.2
Operating system: linux

I'd like to be able to turn the mode indicator on & off without reloading the page or browser. When I'm on a site where Ignore Mode is preferred, I don't use tridactyl's functions on that site much so the mode indicator can also be removed. If Ignore Mode is the only mode without the indicator, that's an indication of Ignore Mode in and of itself. This also means I can easily toggle the indicator on & off as I switch to & from Ignore Mode.

This would be an easy solution for times where I'd like to quickly hide the mode indicator, most of which are when I'm doing something in full-screen like watching videos. Such applications often have their own bindings & functions so I tend to use Ignore Mode anyway.

bovine3dom:
Should be straightforward:

https://github.com/tridactyl/tridactyl/blob/134bc4d1ee6fdffda425788cccda6eebbb099670/src/content.ts#L218

Add a setting to src/lib/config, e.g. modeindicatormodes, mapping modes to true/false. If modeindicator is false it should override it.

ChocolateOverflow:
Is there a way to do this without modifying the source code? I prefer to keep all my configs separate from the source & avoid conflicts caused by modifications which make package management a huge pain.

bovine3dom:
Not possible without modifying the source. I'd merge the patch I described if it found its way to me :)

fluem:
Volunteering. 
Any suggestions about the default setting?
Show modeindicator for all modes except ignore mode, but configurable for all modes along the lines of 
```
    modeindicatormodes: { [key: string]: "true" | "false" } = {
        normal: "true",
        insert: "true",
        input: "true",
        ignore: "false",
        ex: "true",
        hint: "true",
        visual: "true"
    }
```

bovine3dom:
I think the default should be true for all - the mode indicator was added specifically because people wanted to know when they were in ignore mode : )

Thanks for looking into this.

bovine3dom:
this issue is technically fixed by #2902 but I'm leaving it open until the toggling of the setting is also fixed

ChocolateOverflow:
Checking issue #2902, I tried running `:set modeindicatormodes {"ignore":"false"}` and the indicator went away. However, it doesn't come back when I switch back to normal mode or any other mode for that matter. It does come back when I refresh the page though.

[...]

PR #2902, opened 2020-10-13
Allow visability of modeindicator to be configured for each mode

fluem:
Issue #2690 
This introduces a new setting ```modeindicatormodes``` that allows toggling whether the modeindicator is displayed or hidden for each mode separately. The setting is ignored, if ```modeindicator``` is set to ```false```. By default the modeindicator is displayed in every mode.

For example, if we ```:set modeindicatormodes {"hint":"false"}```, it hides the modeindicator when entering hint mode with ```f```.
Exiting hint mode does not display the modeindicator for normal mode - only after reloading the page. (I don't see yet, why that is.)

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -345,7 +345,9 @@ config.getAsync("modeindicator").then(mode => {
         statusIndicator.className +=
             " TridactylMode" + statusIndicator.textContent
 
-        if (config.get("modeindicator") !== "true") statusIndicator.remove()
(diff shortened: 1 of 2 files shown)


bovine3dom:
I think I wrote the mode indicator so I can say this: it's written really badly. It could do with some TLC - that's probably why it doesn't update when exiting hint mode. I wouldn't worry about it too much.

fluem:
I'll leave that for another time then. I don't have much experience with browser extensions yet and I am still getting used to the code.

Issue #2733, opened 2020-08-12
Text shuffler


bovine3dom:
Just for fun: https://www.newscientist.com/letter/mg16221887-600-reibadailty/

Shuffle letters except for first and last in each word. Like rot-13 but slows comprehension less (in my experience).

---

Edit: scratchpad -

```js
reibadailty = (str) => {const arr = str.split(""); for (let i = arr.length - 2; i > 1; i--){const j = Math.floor(Math.random()*i+1);const t = arr[i]; arr[i] = arr[j]; arr[j] = t}; return arr.join("")}
element.textContent = element.textContent.split(" ").map(reibadailty).join(" ")
```

This gets 90% of the way there, but we need to split on any non-a-z character while still retaining those characters (so `.split(/[^a-z]/i)` won't cut it).
```
hint -Fc * element => {const reibadailty = (str) => {const arr = str.split(""); for (let i = arr.length - 2; i > 1; i--){const j = Math.floor(Math.random()*i+1);const t = arr[i]; arr[i] = arr[j]; arr[j] = t}; return arr.join("")}; element.textContent = element.textContent.split(" ").map(reibadailty).join(" ")}
```

bovine3dom:
Added in #2913

PR #2913, opened 2020-10-16
Jumble text

mariuszdb:
I'm creating this PR to implement a text shuffler https://github.com/tridactyl/tridactyl/issues/2733 (I called it "jumble").
I decided to treat a word as a sequence of one of: only lowercase letters or only uppercase letters or an uppercase letter followed by only lowercase letters (word or WORD or Word). For example, a variable in code may be named "someVariableName" and it makes little sense to treat it like one word so it's treated like 3 words ("some", "Variable" and "Name") and the letters are shuffled separately in each one. The same goes for company names and other things written in camelCase/PascalCase. Characters that are not letters are never shuffled.
I do this by splitting the text using this regex: ([^a-zA-Z]|[A-Z][a-z]+).
I based the solution on rot13_helper.
I'm attaching a few screenshots of websites with the text shuffled.
image
image
image

Just for fun.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -129,7 +129,7 @@ import * as DOM from "@src/lib/dom"
 import * as CommandLineContent from "@src/content/commandline_content"
 import * as scrolling from "@src/content/scrolling"
 import { ownTab } from "@src/lib/webext"
-import { wrap_input, getLineAndColNumber, rot13_helper } from "@src/lib/editor_utils"
+import { wrap_input, getLineAndColNumber, rot13_helper, jumble_helper } from "@src/lib/editor_utils"
 import * as finding from "@src/content/finding"
 import * as toys from "./content/toys"
 import * as hinting from "@src/content/hinting"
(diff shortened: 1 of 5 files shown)


bovine3dom:
If we follow the camel case logic I don't think acronyms should be shuffled. A decent heuristic could be: a word in all caps of four letters or fewer is an acronym.

src/excmds.ts
```diff
@@ -4363,6 +4363,22 @@ export function rot13(n: number) {
         body.currentNode.textContent = rot13_helper(t, n)
     }
 }
+/**
+ * Perform text jumbling.
+ *
+ * Shuffles letters except for first and last in all words in text nodes in the current tab. Only characters in
+ * the ASCII range are considered.
```
bovine3dom, line 4370: Should probably add a reference to a site that explains the phenomenon : )

src/lib/config.ts
```diff
@@ -239,6 +239,7 @@ export class default_config {
         x: "stop",
         gi: "focusinput -l",
         "g?": "rot13",
+        "g!": "jumble",
```
bovine3dom, line 242: Can we add an ex alias for `reibadailty`? I liked that joke

src/excmds.ts
mariuszdb: Added

src/lib/config.ts
mariuszdb: Added

mariuszdb:
> If we follow the camel case logic I don't think acronyms should be shuffled. A decent heuristic could be: a word in all caps of four letters or fewer is an acronym.

Added, I was thinking of skipping uppercase words altogether but I guess this is ok.

bovine3dom:
This is my new favourite thing, thank you

Issue #2811, opened 2020-09-10
tab (buffer) list should have have a symbol show and select which tab is playing audio


allanwind:
Tridactyl version: 1.20.1
Firefox version: Mozilla Firefox 68.12.0
Operating system: linux

*   Brief description of the problem:

It would be really helpful if the tab (b) list had a symbol similar to the current % and # that both shows and allow you to quick select the tab with audio currently playing. Firefox marks that tab with a speaker symbol both in the list of tabs on top, and in the "List all tabs" drop-down.

*   Steps to reproduce:

1. Open youtube.com, whatever, and play a video.  Observe the speaker symbol on the tab, and in the tab drop-down list.
2. Open another two new tabs (empty is fine)
3. No press b to see the list of tabs and you cannot tell from the list which is the one with audio.  If have enough tabs, and move sufficiently far away that tab is off-screen.  I don't know of keyboard shot-cut for the "List all tabs" per help page, and Ctrl-L + tabs doesn't navigate to it either.  I am asking for both a visual indicator but also a way to select it.

bovine3dom:
This would be great, but I have no idea how we'd go about doing it. I don't think Mozilla provides an API.

I wonder if TST has already looked into it.

bovine3dom:
Aha, they do provide an API: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/Tab so this is easy : )

allanwind:
I am afraid that I know nothing about add-on development (just a 
happy long-time user):

a. Could you iterate through browser.tabs when you press b to find 
the tab with auditable: true?

b. Then there is this event that seems interesting:

https://developer.mozilla.org/en-US/docs/Mozilla/Gecko/Chrome/API/Browser_API/mozbrowseraudioplaybackchange

bovine3dom:
Since they provide an API it's quite easy technically (my first reply was wrong).

I'm not sure what UI we should provide for switching to tabs with audio playing. Do you have any ideas?

allanwind:
On 2020-09-10 14:36:34, Oliver Blanthorn wrote:
>I'm not sure what UI we should provide for switching to tabs with audio playing. Do you have any ideas?

Maybe I am missing what you are asking...

For show I was thinking of adding a symbol like % and # are 
currently being displayed in the buffer view.  The wrinkle is
that % and # are mutually exclusive while the audio symbol would 
not be (i.e. display both if needed)

[...]

PR #2917, opened 2020-10-18
Add more status symbols to tab list

fluem:
Issues #1338 and #2811 
Introduces tab status symbols ```P```,```A```,```M```,```D``` for pinned, audible, muted, and discarded tabs. If ```completions.Tab.statusstylepretty``` is set to ```true```, capital letters are replaced with Unicode characters. The plain uppercase letters are kept in a hidden column, so that they can still be used for filtering.

Filtering tabs by using ```P```,```A```,```M```,```D``` works whether ```statusstylepretty``` is set or not, although there are still some cases where I don't understand the filtering; will try to work these out.


src/completions/Tab.ts
```diff
@@ -24,17 +24,33 @@ class BufferCompletionOption
...
+            if (tab.audible) pre += "\uD83D\uDD0A"
+            if (tab.mutedInfo.muted) pre += "\uD83D\uDD07"
+            if (tab.discarded) pre += "\u2296"
+        } else {
+            pre = preplain
+        }
 
         // Push prefix before padding so we don't match on whitespace
         this.fuseKeys.push(pre)
+        this.fuseKeys.push(preplain)
```
bovine3dom, line 53: Hmm, if we do this it won't do the case sensitive matching.

I was imagining that we would do something more like the matching we do for numbers.

fluem: Yes, agreed. I don't fully understand how the completion code works yet. Testing out some combinations of tabs does not always give me the selection I expected.

bovine3dom:
I really thought I had merged this and was wondering why I wasn't getting the symbols in the tab list, I assumed something had broken it and was about to file an issue...

Bumping it in the hope that I will remember to review and merge it soon

bovine3dom:
LGTM, just waiting for the tests to pass. Sorry it took such an absurd amount of time to review!

PR #2919, opened 2020-10-18
Allow users to specify which commands get Tab completions

### LINKED ISSUE #2918: `:set` can't handle deep settings whose values are arrays
```
set this.is.an.array [1,2] # set in the config as an array to give it a fair shot
get this.is.an.array # = ["[","1",",","2","]"]
```

Probably already a known issue

bovine3dom:
1) this was outrageously simple, why didn't we do this earlier?
2) I don't like the fact that setting is a space separated string. I'd like to fix #2918 and then update it to be an array. Probably have a `setappend` convenience command for appending to an array, too.
3) will add equivalent for other completions

### DIFF
--- a/src/completions/Tab.ts
+++ b/src/completions/Tab.ts
@@ -68,7 +68,7 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
 
     constructor(private _parent) {
         super(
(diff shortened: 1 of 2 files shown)


bovine3dom:
I'm toying with just keeping it as a space-separated string since the array thing seems to be tricky and this PR has now been held up for quite a while.

Issue #604, opened 2018-05-23
[Feature Request] Local marks


MoSal:
As available in Vim and Vimperator.

## Example 

 * `mb`: create local mark `b` (save the current position of the page).
 * `'b` or `` `b ``: scroll to the saved position.

-------

## Vimperator documentation for reference

    m a-zA-Z
                Set mark at the cursor position. Marks a-z are local to the buffer, whereas
                A-Z are valid between buffers.

    ` or ' a-zA-Z
                Jump to the mark. Marks a-z are local to the buffer, whereas A-Z are valid
                between buffers.

    :delm[arks]!
            Delete the specified marks. Marks are presented as a list.

            Examples:
                :delmarks Aa b p deletes marks A, a, b and p
                :delmarks b-p deletes all marks in the range b to p
                :delmarks! deletes all marks for the current buffer

    :ma or :mark a-zA-Z
            Mark current location within the web page.

    :marks [arg]
                Show all location marks of the current web page. If
                [arg] is specified then limit the list to those
                marks mentioned.

seren:
I would love this capability as well. The "quickmark" functionality which seems to have supplanted it isn't very useful for my use.

liviu-:
Are there any suggested solutions for this?

bovine3dom:
Are you offering to have a crack at implementing it? I'd suggest looking at how `jumpprev` works. It should be fairly straightforward.

PR #2927, opened 2020-10-20
Local and global marks

mariuszdb:
I am creating this issue to resolve https://github.com/tridactyl/tridactyl/issues/604. This PR adds local marks, global marks and a jump to the location before the last mark jump (\`\`). Global marks are saved between restarts, local marks are not (I assumed they could start piling up after a while). I based this loosely on the implementation in Vimium https://github.com/philc/vimium/blob/f1a58259562e6b5c905b11ac05b0cc0deef38976/background_scripts/marks.js. 
m[a-z] - add a local mark,
m[A-Z] - add a global mark,
\`[a-z] - jump to a local mark,
\`[A-Z] - jump to a global mark,
\`\` - jump to the location before the last mark jump


src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+    }
+    const currentTabId = await activeTabId();
+    state.beforeJumpMark = { scrollX: window.scrollX, scrollY: window.scrollY, tabId: currentTabId };
+    try {
+        const tab = await getTabByTabId(mark.tabId);
+        goToTab(tab.id).then(() => {
+            browserBg.tabs.executeScript({
+                code: `window.scrollTo(${mark.scrollX},${mark.scrollY})`
+            });
+            return fillcmdline_tmp(3000, `Jumped to mark '${key}'`);
```
mariuszdb, line 1039: If the mark is in another tab, the message will appear in the tab where the jump is made from. I think adding this in a handler for webNavigation.onCompleted or tabs.onUpdated might fix this, but I wasn't able to access the event handlers from the content script via browserBg proxy  (the shim function is `return browser[api][func](...args)` but I need something like `return browser[api][event][func](...args))`. Any ideas?

bovine3dom:
goodness me, people have asked for this for ages. You'll be very popular : )

src/excmds.ts
bovine3dom: We should probably just add an "execute ex string in tab" command and use that.

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+        return fillcmdline_tmp(3000, `Jumped to mark '${key}'`);
+    }
+    return fillcmdline_tmp(3000, `Mark '${key}' is not set`);
+}
+
+/**
+ * Jumps to a global mark. If the tab with the mark no longer
+ * exists, jumps to another tab with the same url or creates it first if such tab does not exist.
+ * @param key the key associated with the mark
+ * @hidden
```
bovine3dom, line 1023: I don't think this or the local function should be hidden - for one, the help they have is useful to users.

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+        // If there are no matching tabs, open a new one and update the mark's tabId for future use in this session
+        if (!matchingTabs.length) {
+            return openInNewTab(mark.url).then((tab) => {
+                mark.tabId = tab.id;
+                state.globalMarks = globalMarks;
+                browserBg.tabs.executeScript({
+                    code: `window.scrollTo(${mark.scrollX},${mark.scrollY})`
+                })
+                return fillcmdline_tmp(3000, `Jumped to mark '${key}'`);
+            });
```
bovine3dom, line 1053: I think this function could be extracted and reused ~3 times:

```
const scrolltab = tab => {
...
}
openInNewTab(mark.url).then(scrolltab)
```

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+        return jumpGlobalMark(key);
+    }
+    return jumpLocalMark(key);
+}
+
+/**
+ * Jumps to a local mark.
+ * @param key the key associated with the mark
+ * @hidden
+ */
```
bovine3dom, line 1005: Every function in excmds.ts should be explicitly annotated with `//#{content,background,both}{,_helper}`. I suspect setting marks should all be `//#content`, jumping to local marks should be `//#content`, but jumping to global marks should be `//#background`.

src/lib/webext.ts
```diff
@@ -65,6 +65,14 @@ export async function activeTabContainerId() {
     return (await activeTab()).cookieStoreId
 }
 
+export async function getTabsByUrl(url: string) {
+    return await browserBg.tabs.query({url});
+}
+
+export async function getTabByTabId(tabId: number) {
+    return await browserBg.tabs.get(tabId);
```
bovine3dom, line 73: These functions are so simple that it's probably not worth having them - I'd just use browserBg directly.

src/state.ts
```diff
@@ -28,10 +28,29 @@ class State {
...
+        }>> = new Map()
+    beforeJumpMark: {
+        scrollX: number,
+        scrollY: number,
+        tabId: number
+    } = undefined
 }
 
+type MarkKey = string;
+type MarkUrl = string;
```
bovine3dom, line 50: I would personally just use `string` directly - do you see any major advantage to making custom types?

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+    const globalMarks = await State.getAsync("globalMarks");
+    const mark = globalMarks.get(key);
+    if (!mark) {
+        return fillcmdline_tmp(3000, `Mark '${key}' is not set`);
+    }
+    const currentTabId = await activeTabId();
+    state.beforeJumpMark = { scrollX: window.scrollX, scrollY: window.scrollY, tabId: currentTabId };
+    try {
+        const tab = await getTabByTabId(mark.tabId);
+        goToTab(tab.id).then(() => {
```
bovine3dom, line 1035:
```
Messaging.messageTab(tab.id, "controller_content", "acceptExCmd", [`composite scrollto ${mark.scrollX} ${mark.scrollY}; fillcmdline_tmp 3000 Jumped to mark '${key}'`])
```

I think this is how I'd solve your problem. We should probably make a little `acceptExCmd(destination: tabId | "background", ex_string)` helper function so we don't have to remember this strange incantation every time.

bovine3dom (changes requested):
Thanks so much for this PR; it will make a lot of people happy. I think it's 90% of the way there. I've made some initial comments here.

Because it's quite a large PR I'll probably find more things to bikeshed later, so please don't assume these comments are exhaustive : )

src/excmds.ts
mariuszdb: I removed `@hidden`s

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
     })
 }
 
+/**
+ * Jumps to a local mark, a global mark, or the location before the last mark jump.
+ * [a-z] are local marks, [A-Z] are global marks and '`' is the location before the last mark jump.
+ * @param key the key associated with the mark
+ */
+//#content
+export async function jumpmark(key: string) {
```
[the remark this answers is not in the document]
mariuszdb: I changed the names to `markjump, markjumplocal, markjumpglobal, beforemarkjump, markadd, markaddlocal, markaddglobal

src/excmds.ts
```diff
@@ -961,6 +976,168 @@ export function jumpprev(n = 1) {
...
+    if (key.length !== 1) {
+        return;
+    }
+    if (key === "`") {
+        return jumpBeforeJumpMark()
+    }
+    if (!/[a-z]/i.exec(key)) {
+        return;
```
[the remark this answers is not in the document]
mariuszdb: Added error messages.

src/excmds.ts
mariuszdb: done

src/excmds.ts
mariuszdb: I added //#content annotations to all the functions including markjumpglobal because it gets current url now.

src/lib/webext.ts
mariuszdb: Funcions removed

src/excmds.ts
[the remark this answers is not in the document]
bovine3dom: Perhaps `markjumpbefore`? If it's going in reverse

src/state.ts
mariuszdb: Removed. My intention was to make it clear what the keys in the maps are,

src/excmds.ts
mariuszdb: Thanks that works but when opening a new tab from a global mark the message still appeared in the old tab for me. I made the scroll tab like this: `browserBg.tabs.executeScript(//scrolling).then(//printing) which seems to make the message appear in a new tab for me most of the time.

src/excmds.ts
```diff
@@ -4380,6 +4572,18 @@ export function jumble() {
         body.currentNode.textContent = jumble_helper(t)
     }
 }
+
+/**
+ * Run a command in a tab with an identifier specified by destination or in background if destination is "background".
```
bovine3dom, line 4577: Should probably specify that destination is an internal mozilla tab id and not the 1-based tab index we assign them

src/excmds.ts
```diff
@@ -961,6 +974,185 @@ export function jumpprev(n = 1) {
...
+    }
+    return markjumplocal(key);
+}
+
+/**
+ * Jumps to a local mark.
+ * @param key the key associated with the mark
+ */
+//#content
+async function markjumplocal(key: string) {
```
bovine3dom, line 1004: I think all these functions need to be exported. I'm not sure what happens if they aren't

src/excmds.ts
```diff
@@ -961,6 +974,185 @@ export function jumpprev(n = 1) {
...
+}
+
+/**
+ * Scrolls to a given position in a given tab and prints a message in it.
+ */
+//#content
+export async function scrolltab(tabId: number, scrollX: number, scrollY: number, message: string) {
+    browserBg.tabs.executeScript({
+        code: `window.scrollTo(${scrollX},${scrollY})`
+    }).then(() => acceptExCmd(tabId, [`fillcmdline_tmp 3000 ${message}`]));
```
bovine3dom, line 1105: Is this where scrolling within acceptExCmd meant that the message appeared in the wrong tab? I'll try to look into it

bovine3dom:
I'll try to get round to this within the next couple of days. If I don't manage it, it'll probably be next week. Sorry for the wait : )

src/excmds.ts
bovine3dom: @mariuszdb I think I still need an answer to this - I'd prefer to avoid executeScript as we don't use it anywhere else. Thanks!

src/excmds.ts
```diff
@@ -961,6 +974,183 @@ export function jumpprev(n = 1) {
...
+            scrolltab(tab.id, mark.scrollX, mark.scrollY, `Jumped to mark '${key}'`)
+        })
+    }
+
+    // the tab with mark's tabId doesn't exist or it has a different url than the mark's url
+    async function onTabNoLongerValid() {
+        const matchingTabs = await browserBg.tabs.query({url: mark.url})
+        // If there are no matching tabs, open a new one and update the mark's tabId for future use in this session
+        if (!matchingTabs.length) {
+            return openInNewTab(mark.url).then(updateMarkAndScroll())
```
bovine3dom, line 1057: On my machine, this doesn't scroll to the right place once the tab is opened. We probably need to wait for the DocLoad event. I'll try to look into it today

bovine3dom: A broken sketch here:

```
            return openInNewTab(mark.url).then(tab => {
                mark.tabId = tab.id
                state.globalMarks = globalMarks
                browserBg.tabs.executeScript(
                    tab.id, {
                        code: `
                        window.addEventListener("DOMContentLoaded", () => window.scrollTo(${mark.scrollX}, ${mark.scrollY}))
                        `
                    }
                )
            })
```

I think "open tab and wait for it to load" is such a common operation that we should probably update `openInNewTab` to have an option that waits for the DOMContentLoaded event to fire before returning the promise.

bovine3dom:
The capital vs lowercase mark distinction needs a bit of work I think. Probably as simple as making the "mark not set" message say "No local mark on this tab" and "No global mark".

The messages could be a little prettier too - probably just need `# ` like all our others

bovine3dom (changes requested):
This works now but I need to polish it a little more before I'm happy to merge it.

src/state.ts
```diff
@@ -136,6 +130,15 @@ notBackground &&
         if (message.command == "stateUpdate") {
             const property = message.args.property
             const value = message.args.value
+            // Ensure we don't accidentally store anything sensitive
+            const inIncognitoContext = message.args.inIncognitoContext
+            if (inIncognitoContext && PERSISTENT_KEYS.includes(property)) {
+                overlay[property] = value
```
bovine3dom, line 136: What is this line for?

(Thanks for looking at this, by the way!)

src/excmds.ts
```diff
@@ -4709,6 +4902,18 @@ export function jumble() {
         body.currentNode.textContent = jumble_helper(t)
     }
 }
+
+/**
+ * Run a command in a tab with an internal firefox identifier specified by destination or in background if destination is "background".
+ */
+//#content
+export function acceptExCmd(destination: number | "background", ex_string: string[]) {
```
bovine3dom, line 4910:
```suggestion
export function acceptExCmd(destination: number | "background", ...ex_string: string[]) {
```

without this we can't run it interactively from the Tridactyl command line

src/excmds.ts
```diff
@@ -1025,6 +1026,183 @@ export function jumpprev(n = 1) {
...
+        // the mark's tab is no longer valid
+    }
+}
+
+/**
+ * Scrolls to a given position in a given tab and prints a message in it.
+ */
+//#content
+export async function scrolltab(tabId: number, scrollX: number, scrollY: number, message: string) {
+    acceptExCmd(tabId, [`composite scrollto ${scrollX} ${scrollY}; fillcmdline_tmp 3000 ${message}`])
```
bovine3dom, line 1154:
```suggestion
    await acceptExCmd(tabId, [`scrollto ${scrollX} ${scrollY}`])
    acceptExCmd(tabId, [`fillcmdline_tmp 3000 ${message}`])
```

I'm not sure why I prefer this but I do.

src/excmds.ts
```diff
@@ -1025,6 +1026,183 @@ export function jumpprev(n = 1) {
...
+        const currentTabId = await activeTabId()
+        state.beforeJumpMark = { url: window.location.href.split("#")[0], scrollX: window.scrollX, scrollY: window.scrollY, tabId: currentTabId }
+        goToTab(tabId).then(() => scrolltab(tabId, scrollX, scrollY, "# Jumped to the last location before a mark jump"))
+    } catch (e) {
+        // the mark's tab is no longer valid
+    }
+}
+
+/**
+ * Scrolls to a given position in a given tab and prints a message in it.
```
bovine3dom, line 1150: the `tabId` distinction will confuse users, I think

src/excmds.ts
```diff
@@ -1140,18 +1140,19 @@ export async function markjumpbefore() {
...
 }
 
 /**
- * Scrolls to a given position in a given tab and prints a message in it.
+ * Scrolls to a given position in a tab identified by tabId and prints a message in it.
  */
 //#content
 export async function scrolltab(tabId: number, scrollX: number, scrollY: number, message: string) {
-    acceptExCmd(tabId, [`composite scrollto ${scrollX} ${scrollY}; fillcmdline_tmp 3000 ${message}`])
+    await Messaging.messageTab(tabId, "controller_content", "acceptExCmd", [`composite scrollto ${scrollX} ${scrollY}`])
```
bovine3dom, line 1154:
```suggestion
    await Messaging.messageTab(tabId, "controller_content", "acceptExCmd", [`scrollto ${scrollX} ${scrollY}`])
```

: )

dlobue:
What's the status of this PR? I'd love to use this feature! I find myself double-tapping `'` all the time, even though I know it doesn't do anything yet!

bovine3dom:
I have absolutely no recollection of where we got to with it. It has a "needs review" so it's probably my fault.

I'll try to look at it soon. If I don't, please poke me : )

src/state.ts
bovine3dom: I think I'm going to remove this line - otherwise you could

1. set the property to something sensitive
2. reject the write
3. set another property to something not at all sensitive
4. write it all including the sensitive property

Please let me know if I'm missing something : )

mariuszdb:
I'm not sure where this was left either. I pulled in your changes and the changes from master and the check for incognito wasn't working since browser.windows isn't available in content scripts, so I changed markadd to be a background script. I played around a bit and the marks seem to work fine. The only thing I noticed is that sometimes when you open a new tab via a global mark, the page isn't scrolled to the saved coordinates (I checked this using youtube.com).

wetlife:
This PR appears to pass all checks. May we please get the feature for regular tabs and delay incognito functionality if needed? Is there anything I can do to help?

bovine3dom:
@wetlife I was worried about this - https://github.com/tridactyl/tridactyl/pull/2927#discussion_r720685925 - but it looks like it has gone now.

I'll try to review it when I'm not on holiday any more. Honestly the most helpful thing you can do is remind me in a week or two :)

wetlife:
I hope your vacation became a persistent joy. Is it possible to release the local/global marks feature?

bovine3dom:
Hahaha, it was, thanks for the reminder, I'll look at it tomorrow :)

src/state.ts
```diff
@@ -76,13 +102,6 @@ const state = new Proxy(overlay, {
 
         // Persist "sets" to storage in the background for some keys
         if (PERSISTENT_KEYS.includes(property)) {
-            // Ensure we don't accidentally store anything sensitive
-            if (browser.extension.inIncognitoContext) {
-                console.error(
-                    "Attempted to write to storage in private window.",
-                )
-                return false
-            }
```
bovine3dom, line 85: I'm pretty nervous about this change - I need to look at it really carefully and make sure it is safe :)

readme.md
```diff
@@ -127,6 +127,12 @@ If you want to be able to use `<C-f>` to search for things, use `<C-f>` after op
 
 If you want to use Firefox's default `<C-b>` binding to open the bookmarks sidebar, make sure to run `unbind <C-b>` because Tridactyl replaces this setting with one to go to the previous part of the page.
 
+#### Marks
+
+-   `m a-zA-Z` — set a local mark (lowercase letter), or a global mark (uppercase letter)
+-   `` ` a-zA-Z `` — jump to a local mark (lowercase letter), or a global mark (uppercase letter)
+-   ``` `` ``` — jump to the location before the last mark jump
+
```
bovine3dom, line 135: Should probably add this to the tutor too

src/excmds.ts
```diff
@@ -1047,6 +1048,187 @@ export function jumpprev(n = 1) {
...
+ * Scrolls to a given position in a tab identified by tabId and prints a message in it.
+ */
+//#content
+export async function scrolltab(tabId: number, scrollX: number, scrollY: number, message: string) {
+    await Messaging.messageTab(tabId, "controller_content", "acceptExCmd", [`scrollto ${scrollX} ${scrollY}`])
+    Messaging.messageTab(tabId, "controller_content", "acceptExCmd", [`fillcmdline_tmp 3000 ${message}`])
+}
+
+/**
+ * Adds a global or a local mark. Assigns a global mark to a key or a local mark the current url and a key.
```
bovine3dom, line 1181: should explain here that local marks are lowercase and global marks are upper case

bovine3dom: >  Assigns a global mark to a key or a local mark the current url and a key.

Seems a bit mangled, I'm not sure what it is trying to say after ... or a local mark

bovine3dom (commented):
This seems to work great but there's a few things I want to check before merging it.

I'll try really hard to get them done within the next week or so. Sorry it's taking so long!

fiskhest:
Is there an update on this PR? @bovine3dom

bovine3dom:
Hi @fiskhest the PR is 99% ready - all the features work - but I need to sit down and seriously look at this change https://github.com/tridactyl/tridactyl/pull/2927/files#r793643221 to see if it introduces any security bugs, it's the part of Tridactyl that makes sure we never write activity in private windows to storage so it's quite important :)

bovine3dom (approved):

seren:
Thanks so much @bovine3dom and @mariuszdb for all the work that went into this!

PR #2932, opened 2020-10-21
WIP #2918: open a can of worms to fix setting arrays in nested objects

bovine3dom:
1) provide an array in a nested object in src/lib/config.ts
2) `set that.array.to [1,2]`
3) the "to" setting disappears

`fallback` seems to be called and return the right value so the problem must be upstream somewhere. `:js tri.config.set("that","array","to",[1,2])` does work fine though. : /

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3413,25 +3413,33 @@ function validateSetArgs(key: string, values: string[]) {
     const file = Metadata.everything.getFile("src/lib/config.ts")
     const default_config = file.getClass("default_config")
     const md = default_config.getMember(target[0])
-    if (md !== undefined) {
-        const strval = values.join(" ")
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -3413,25 +3413,33 @@ function validateSetArgs(key: string, values: string[]) {
...
+            return values.join(" ")
+        } else {
+            throw new Error("Unsupported setting type!")
+        }
+    }
     if (md !== undefined) {
         const strval = values.join(" ")
         // Note: the conversion will throw if strval can't be converted to the right type
-        if (md.type.kind === "object" && target.length > 1) {
+        if (md.type.kind === "object" && target.length == 2) {
```
bovine3dom, line 3431: it feels obvious that this was the problem, but it clearly wasn't the only problem :(

PR #2934, opened 2020-10-22
Morgan/kill but reversable

### LINKED ISSUE #2794: ;k but reversible
;k is our kill-element hint mode. You can't unkill elements so if you make a mistake you have to refresh.

Could probably hide by adding a random classname which is hidden. Can then unhide by removing that CSS, taking from a stack.

Inspired by https://www.reddit.com/r/ProgrammerHumor/comments/im2cij/hacker_man_100/g40jont?utm_medium=android_app&utm_source=share&context=3

morganrconnolly:
adds hintmode excmd `-K` which restores killed elements in LIFO. `-K` for unKill
fixes #2794

### DIFF
--- a/src/content/state_content.ts
+++ b/src/content/state_content.ts
@@ -20,7 +20,6 @@ export class PrevInput {
 class ContentState {
     mode: ModeName = "normal"
     suffix = ""
-    killStack: Element[] = []
 }
 
 export type ContentStateProperty =
@@ -28,7 +27,6 @@ export type ContentStateProperty =
     | "cmdHistory"
     | "prevInputs"
     | "suffix"
-    | "killStack"
 
 export type ContentStateChangedCallback = (
     property: ContentStateProperty,
@@ -46,9 +44,7 @@ export function addContentStateChangedListener(
 }
 
 export const contentState = (new Proxy(
-    { mode: "normal", 
-      killStack: [] 
-    },
+    { mode: "normal" },
     {
         get(target, property: ContentStateProperty) {
             return target[property]
--- a/src/content/state_content.ts
+++ b/src/content/state_content.ts
@@ -20,6 +20,7 @@ export class PrevInput {
 class ContentState {
     mode: ModeName = "normal"
     suffix = ""
+    killStack: Element[] = []
 }
 
 export type ContentStateProperty =
@@ -27,6 +28,7 @@ export type ContentStateProperty =
     | "cmdHistory"
     | "prevInputs"
     | "suffix"
+    | "killStack"
 
 export type ContentStateChangedCallback = (
     property: ContentStateProperty,
@@ -44,7 +46,9 @@ export function addContentStateChangedListener(
 }
 
 export const contentState = (new Proxy(
-    { mode: "normal" },
+    { mode: "normal", 
+      killStack: [] 
+    },
     {
         get(target, property: ContentStateProperty) {
             return target[property]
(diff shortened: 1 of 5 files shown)


bovine3dom (commented):
This is great, thanks!

I replied to you on Matrix about the killStack but I'll reproduce it here:

```
> <@mci:matrix.org> Hello! I'm I'd like to work on this issue https://github.com/tridactyl/tridactyl/issues/2794 . Is content/state_content.ts the place to store the "killStack" ? 

Hi! Thanks for that PR. It looks good.
I wouldn't personally have put it there - we use it for stuff that needs to be shared between multiple scripts and particularly for state that other things need to be told about when there is a change. I would just do

"""
//#content_helper
const KILL_STACK = Element []
"""

above the hint function.
It doesn't really matter though : )
```

morganrconnolly:
@bovine3dom  Thank you for the feedback! I'm much more fond of this PR now. I was able to implement all of your comments, but let me know if you see any other possible improvements. In particular, let me know if you think there are better names for the new reversible kill hint mode, now `-K`, and/or the command to restore killed elements , now `:restorekill` .

src/excmds.ts
```diff
@@ -3967,23 +3970,7 @@ export function setnull(...keys: string[]) {
...
-        - -f [text] hint links and inputs that display the given text
+        - -p copy an element's text to the clipboard -h select an element (as if you click-n-dragged over it) -P copy an element's title/alt text to the clipboard -r read an element's text with text-to-speech -i view an image -I view an image in a new tab -k delete an element from the page -K deletes an element from the page; deleted elements can be restored using :restorekilled.  -s save (download) the linked resource -S save the linked image -a save-as the linked resource -A save-as the linked image -; focus an element and set it as the element or the child of the element to scroll -# yank an element's anchor URL to clipboard -c [selector] hint links that match the css selector `bind ;c hint -c [class*="expand"],[class="togg"]` works particularly well on reddit and HN this works with most other hint modes, with the caveat that if other hint mode takes arguments your selector must contain no spaces, i.e. `hint -c[yourOtherFlag] [selector] [your other flag's arguments, which may contain spaces]` -f [text] hint links and inputs that display the given text
```
bovine3dom, line 3973: it looks like some formatting was lost here

src/content/state_content.ts
```diff
@@ -20,13 +20,15 @@ export class PrevInput {
 class ContentState {
     mode: ModeName = "normal"
     suffix = ""
+    killStack: Element[] = []
```
bovine3dom, line 23: I don't think we need any of the changes to this file any more

src/excmds.ts
```diff
@@ -4907,4 +4903,13 @@ export async function extoptions(...optionNameArgs: string[]) {
     return winopen("-popup", selectedExtension.optionsUrl)
 }
 
+/** Restores elements killed using ;K. 
+ *
+ *NB: The most recently killed element that has not already been restored will be restored.
+ */
+//#content
+export async function restorekill() {
```
bovine3dom, line 4911: Yeah, I don't know what to call this either. In general we have a convention of naming things `[object][action]`, so under that this would be `killrestore` ... but I don't like that either : )

I'll try to think about it a bit more.

bovine3dom (commented):
Thanks for this. Just a few minor things and then the hard problem of picking a name. I'm happy to do all of them (at some point).

src/excmds.ts
fluem: How about ```resurrect()```, ```revive()``` or ```elementrestore()```?

bovine3dom: I like elementrestore the most out of those ... still not 100% happy with it though.

I'll try really hard to get round to look seriously at this PR next week : )

bovine3dom:
LGTM - I'll merge once CI has passed. Thanks, and sorry it took so long to review :)

PR #2970, opened 2020-11-03
WIP: add waitForDom option to openInNewTab

bovine3dom:
Unblocks #2927

### DIFF
--- a/src/lib/webext.ts
+++ b/src/lib/webext.ts
@@ -130,6 +130,7 @@ export async function openInNewTab(
         related: false,
         cookieStoreId: undefined,
     },
+    waitForDom = false,
 ) {
     const thisTab = await activeTab()
     const options: Parameters<typeof browser.tabs.create>[0] = {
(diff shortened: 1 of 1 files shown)


src/lib/webext.ts
```diff
@@ -165,15 +166,31 @@ export async function openInNewTab(
...
-    } else {
-        // load in background and then activate, per issue #1993
-        return browserBg.tabs
-            .create(options)
-            .then(newtab => browserBg.tabs.update(newtab.id, { active: true }))
+    // load in background
+    let tab = browserBg.tabs.create(options)
+
+    if (kwargs.active) {
+        // activate once tab has been opened, per issue #1993
```
bovine3dom, line 173: I haven't checked that this behaviour is the same - probably ought to since it's a crucial bit of code.

bovine3dom:
I have no recollection of making this PR.

I think I added it in a nicer way in 85129c3463e560a484171349c46b3ac817224f27 anyway.

PR #3017, opened 2020-11-21
Fix #3016

### LINKED ISSUE #3016: js -r does not find the specified file on Windows
Tridactyl version: 1.20.2
Firefox version: Mozilla Firefox 83.0
Operating system: win

*   **Brief description of the problem:** On Windows, `js -r` consistently fails with "controller in excmd: Error: Couldn't read js file &lt;filename&gt;".

*   **Steps to reproduce:**
    1.  On Windows, create a js file in the same directory your .tridactylrc is located in
    1. Run `:js -r <filename>`

*   **URL of the website the bug happens on:** N/A

*   **Config:** N/A

This is caused by the .tridactylrc path being split on forward slashes, while the native messengers returns a path separated by backslashes: https://github.com/tridactyl/tridactyl/blob/bff998aeb759d4a92f2b77c61be43fc38b6b3aa4/src/excmds.ts#L4681-L4682 I'll be submitting a PR shortly with a proposed fix.

Rummskartoffel:
I've no idea whether this is the best way to deal with this – if, for instance, the native messenger were ever made to always normalise paths to use forward slashes, even on Windows, this'd break –, but since this is the *third* issue I've filed, I think I can at least *try* to be helpful instead of just complaining.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4744,7 +4744,7 @@ async function js_helper(str: string[]) {
     if (doSource) {
         let sourcePath = jsContent
         if (fromRC) {
-            const sep = (await browserBg.runtime.getPlatformInfo()).os == "win" ? "\\" : "/"
+            const sep = "/"
             const rcPath = (await Native.getrcpath()).split(sep).slice(0, -1)
             sourcePath = [...rcPath, sourcePath].join(sep)
         }
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4745,7 +4745,7 @@ async function js_helper(str: string[]) {
         let sourcePath = jsContent
         if (fromRC) {
             const sep = "/"
-            const rcPath = (await Native.getrcpath()).split(sep).slice(0, -1)
+            const rcPath = (await Native.getrcpath("unix")).split(sep).slice(0, -1)
             sourcePath = [...rcPath, sourcePath].join(sep)
         }
         const file = await Native.read(sourcePath)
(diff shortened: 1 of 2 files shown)


bovine3dom:
Updating the native messenger itself is tricky because it doesn't update automatically (despite our attempts to do so), but I'd still prefer it if this was a bit more general.

Could you add an option to https://github.com/tridactyl/tridactyl/blob/eca1e1614b8d0e51dcb7369a3b4026fed72ad209/src/lib/native.ts#L61 that swaps backslashes for forward slashes on windows? Very similar to your code here but in that function instead. E.g. `getRcPath(pathstyle: "unix" | "auto" = "auto")`.

FWIW, filing issues is helpful. Tridactyl has more bugs than I can find/notice on my own : )

Rummskartoffel:
Unfortunately, I had to skip the commit hooks, because they didn't work correctly (probably because of Windows or Git Bash or both). I'm not really familiar with how these things work, so someone else who knows what they're doing might have to manually re-run those on the files I changed.

bovine3dom:
We should probably just skip the hooks on anything but Linux. Mostly what they do is ensure that the `lint` test passes : )

bovine3dom:
@Rummskartoffel I won't have access to a Windows machine for a very long time, so could you triple check that this works and fixes your issue? I'll test it doesn't break Linux but I can't do much more than that. Thanks!

bovine3dom:
It works fine on my Linux machine and the code looks fine. @Rummskartoffel can you confirm that you've tested this on Windows? : )

Rummskartoffel:
Yes, I have and found no issues.

PR #3027, opened 2020-11-23
Fix #1651

### LINKED ISSUE #1651: `:nativeinstall` is broken on Windows
Tridactyl version: 1.15.0pre2902-ff73610
Firefox version: Mozilla Firefox 67.0
Operating system: win

Need to do something along the lines of

```
$TempScript = Invoke-WebRequest 'https://raw.githubusercontent.com/cmcaine/tridactyl/master/native/win_install.ps1'; $ScriptBlock = [Scriptblock]::Create($TempScript.Content); Invoke-Command -ScriptBlock $ScriptBlock -ArgumentList ($args + @('Tag') + @('master')) 
```

which gives the error
```
Invoke-Command : Cannot convert the "Tag" value of type "System.String" to type "System.Management.Automation.SwitchParameter".
At line:1 char:177
+ ... t.Content); Invoke-Command -ScriptBlock $ScriptBlock -ArgumentList ($ ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [Invoke-Command], PSInvalidCastException
    + FullyQualifiedErrorId : ConvertToFinalInvalidCastException,Microsoft.PowerShell.Commands.InvokeCommandCommand
```

so we just need to work out how to pass `-Tag master` as an argument.

Rummskartoffel:
Because PowerShell's syntax is weird, I wasn't able to find a way to properly escape a `-Command` in string form, so I decided to use a `ScriptBlock` instead, which has the slight disadvantage of only working when pasted into PowerShell.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -694,7 +694,7 @@ export async function nativeinstall() {
     if ((await browser.runtime.getPlatformInfo()).os === "win") {
         const installstr = (await config.get("win_nativeinstallcmd")).replace("%WINTAG", "-Tag " + tag)
         await yank(installstr)
-        done = fillcmdline("# Installation command copied to clipboard. Please paste and run it in PowerShell (other shells won't work) to install the native messenger.")
+        done = fillcmdline("# Installation command copied to clipboard. Please paste and run it in cmd.exe (other shells won't work) to install the native messenger.")
     } else {
         const installstr = (await config.get("nativeinstallcmd")).replace("%TAG", tag)
         await yank(installstr)
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -694,7 +694,7 @@ export async function nativeinstall() {
     if ((await browser.runtime.getPlatformInfo()).os === "win") {
         const installstr = (await config.get("win_nativeinstallcmd")).replace("%WINTAG", "-Tag " + tag)
         await yank(installstr)
-        done = fillcmdline("# Installation command copied to clipboard. Please paste and run it from cmd.exe, PowerShell, or MinTTY to install the native messenger.")
+        done = fillcmdline("# Installation command copied to clipboard. Please paste and run it in PowerShell (other shells won't work) to install the native messenger.")
     } else {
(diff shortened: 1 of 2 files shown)


bovine3dom:
Thanks for this! In theory this runs automatically when there's an update to Tridactyl.

https://github.com/tridactyl/tridactyl/blob/eca1e1614b8d0e51dcb7369a3b4026fed72ad209/src/excmds.ts#L815

https://github.com/tridactyl/tridactyl/blob/eca1e1614b8d0e51dcb7369a3b4026fed72ad209/src/excmds.ts#L4877

I think `native.run` is cmd.exe on most systems so this would break. We could take the approach of saying that native update is broken on Windows like we do for OSX, but we should do something : )

Rummskartoffel:
> I think `native.run` is cmd.exe on most systems so this would break. We could take the approach of saying that native update is broken on Windows like we do for OSX, but we should do something : )

Ugh, I hadn't considered that. I'm pretty sure it's *always* cmd.exe, so this wouldn't work for any Windows users. Thing is, I couldn't for the life of me figure out how to properly escape everything when passing the command in as a string and not as a ScriptBlock, at least when pasting into PowerShell. I *might* be able to figure something out that'll work in cmd.exe only instead, though.

bovine3dom:
Looks good to me : )

Again, I've no access to a Windows machine so could you confirm that you've tested that this works? E.g. if you set the Tridactyl version and version mame in src/manifest.json to something really old like 1.2.0, does an old version of the native messenger get installed after :nativeinstall?

Thanks : )

Rummskartoffel:
Good thing you asked again, otherwise I wouldn't have noticed that missing quote (side note: I really hate the cmd.exe syntax for *totally unrelated* reasons).
I tried changing the version and it worked, though I couldn't go as far back as 1.2.0, because the native helper seems to have had a different directory structure back then.

bovine3dom:
Just want to make sure that the e2e windows test failure isn't repeatable. If it passes all the tests I'll merge it :)

PR #3030, opened 2020-11-25
Remove outdated :set help text

Rummskartoffel:
I recently noticed this help text and thought to myself "wait, wasn't one of the first things I did after installing setting some searchurls?". I tried it out again and it appears `:set`'s documentation is indeed outdated.


bovine3dom:
The great thing about keeping documentation right next to the function is that it's not possible for the documentation to get out of date 🤔

Issue #3022, opened 2020-11-22
git hooks are doubly broken on Windows


Rummskartoffel:
1. The dev dependency "shared-git-hooks" causes `yarn install` to fail because it receives a permission error when trying to create its symlinks. This can be prevented by running as admin, but we might as well just achieve the same effect by putting a `preinstall` script into package.json (`git config core.hooksPath hooks` does the trick) instead of relying on a partially broken npm package that hasn't been updated in ~2 years. See also: kilianc/shared-git-hooks#8
1. `post-checkout` seems to work fine (or at least it doesn't report any errors), but `pre-commit` creates a file called "nul" (which, btw, is a reserved file name on Windows that you can't normally create – or delete) that contains a bunch of linter errors, whenever I try to commit changes (to .ts files, at least). The first time I made a commit to (a fork of) this repository, *something* also created two weirdly named directories ("lint.*&lt;random letters&gt;*", IIRC), but I don't remember whether that was the hook or `yarn run pretty`, which I dutifully ran after the hook told me to.

bovine3dom:
We should definitely investigate why 1) doesn't happen on CI. Half the point of it was to stop us breaking building on Windows.

Rummskartoffel:
> 
> 
> We should definitely investigate why 1) doesn't happen on CI. Half the point of it was to stop us breaking building on Windows.
As far as I could find out, that is because GitHub Actions always run as admin, so they can create those symlinks just fine.

Rummskartoffel:
I've started looking into this a little more, and it turns out "nul" is basically the MS-DOS version of /dev/null, so I guess Git Bash tried to translate the write to /dev/null done by eslint in the pre-commit hook and wrote to a normal file instead of the device file for some reason. The strangely-named directories I mentioned apparently came from that hook as well, as they are *supposed* to be created in .temp/, but for some reason they just... weren't.
I'm also gonna go and see if I can't make these hooks work in Git Bash/MinGW, because it turns none of them worked after all, which I somehow hadn't noticed until today, and I think the convoluted shared-git-hooks thing is pretty unnecessary.

Rummskartoffel:
Another update: making the hooks *run* was not an issue, but making them *work* could prove more difficult. I've found out the following things:
* under MinGW, node's `fs.writeFileSync("/dev/null", "foo")` works as expected, and `fs.writeFileSync("nul", "foo")` creates a normal file called "nul", arguably also expected.
[...]

PR #3033, opened 2020-11-26
Make git hooks work on MinGW etc.

Rummskartoffel:
This pull request deals with #3022 in the following ways:
1. Replace shared-git-hooks:
    * with the package.json lifecycle script "install", which locally sets the git hooks search directory to `<working_directory>/hooks` when `yarn install` is run
    * with an `echo` in each of the scripts that informs the user when the hook runs.
1. Add a check to the pre-commit hook that just skips it when on MinGW, MSYS, or Cygwin<sup>1</sup>.

[1] I'm fairly certain it'll be broken on pure MSYS, which, as far as I know, MinGW and in turn Git Bash are based on, as well. I'm not sure about Cygwin, so we *could* leave out that check for now – but I don't think it matters. Either it turns out it would've worked, leading to marginally less prettified code, or it turns out it doesn't work either, leading to a Cygwin user getting slightly confused, then finding #3022 and this PR and adding that check as well. Both are rather minuscule dangers.

### DIFF
--- a/hooks/post-checkout
+++ b/hooks/post-checkout
@@ -2,6 +2,8 @@
 
 # stolen from https://gist.github.com/sindresorhus/7996717
 
+echo "Running post-checkout hook..."
+
 changed_files="$(git diff-tree -r --name-only --no-commit-id $1 $2)"
 
 
--- a/hooks/post-merge
+++ b/hooks/post-merge
@@ -2,6 +2,8 @@
 
 # stolen from https://gist.github.com/sindresorhus/7996717
 
+echo "Running post-merge hook..."
+
 changed_files="$(git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD)"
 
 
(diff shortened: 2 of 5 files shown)


hooks/post-checkout
```diff
@@ -2,6 +2,8 @@
 
 # stolen from https://gist.github.com/sindresorhus/7996717
 
+echo "Running post-checkout hook..."
```
bovine3dom, line 5: I'm not keen on this verbosity, do you think it adds useful information? I change between branches reasonably frequently (and I sync them up so I'm not triggering yarn install).

Rummskartoffel: shared-git-hooks prints something like this whenever it triggers, so I added it for the time being to replace that functionality. I agree that it's probably not necessary for post-checkout and post-merge, but I'd like to keep it in pre-commit because otherwise that would run for pretty long without any indication of what's happening.

PR #3046, opened 2020-12-01
Add remark to :command doc about creating excmds

### LINKED ISSUE #3026: Is there a way to define aliases with arguments?
Tridactyl version: 1.20.2pre4788-35ebb97d
Firefox version: Mozilla Firefox 78.0.2
Operating system: linux

It appears that `alias`es are handled by string-replacing the alias in the excmd string, so there's no good way to give them arguments - anything you write after the alias is necessarily tacked on to the end of the excmd string. You can sort of handle it by using `js`, structuring the alias like `(function(args){ ... })(`, and then remembering that you have to use js arguments and close the paren, but this is quite icky. It'd be nice if there was a way to define aliases with arguments.

This may have to wait on having a real parser.

Rummskartoffel:
Resolves #3026.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3266,7 +3266,8 @@ export async function taball(id: string) {
  * entering `:tn 43` will expand to `:tabnext_gt 43`.
  *
  * Note that this is only for excmd->excmd mappings. To map a normal-mode
- * command to an excommand, see [[bind]].
+ * command to an excommand, see [[bind]]. For defining your own excommands,
+ * see [[js]].
  *
  * See also:
  *  - [[comclear]]
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3263,11 +3263,11 @@ export async function taball(id: string) {
  *  - `command hello t` This will expand recursively into 'hello'->'tabopen'
  *
  * Commands/aliases are expanded as in a shell, so, given the commands above,
- * entering `:tn 43` will expand to `:tabnext_gt 43`.
+ * entering `:tn 43` will expand to `:tabnext_gt 43`. You can use this to create
+ * your own ex-commands in conjunction with [[js]], specifically `js -p` and `js -d`.
  *
- * Note that this is only for excmd->excmd mappings. To map a normal-mode
- * command to an excommand, see [[bind]]. For defining your own excommands,
- * see [[js]].
+ * Note that this is only for excmd -> excmd mappings. To map a normal-mode
+ * command to an excommand, see [[bind]].
  *
  * See also:
  *  - [[comclear]]


src/excmds.ts
```diff
@@ -3266,7 +3266,8 @@ export async function taball(id: string) {
  * entering `:tn 43` will expand to `:tabnext_gt 43`.
```
bovine3dom, line 3266: I think this could be a bit clearer:

1. `js [...]` is still and ex-command, this makes it sound like it isn't
2. the key problem with #3026 is that it wasn't clear that `command`s can accept arguments - I think really what we need to do is add a bit to this to say something along the lines of "if you need to control where the argument goes in the ex-command, use [[js]] with the -d or -p options"

Thanks for taking the time to look at this. What do you think? : )

bovine3dom: For clarity the solution that saulrh couldn't find in #3026 was `command loudecho js -d€ window.alert(JS_ARGS.join(" "))€`.

Rummskartoffel: Damn, I may have managed to completely miss the point of the issue... so if *now* I understand correctly, the desired effect was to create an alias with composite that passes the arguments to any part of the composite command other than the last, i. e. something like `composite <command> <arguments go here> | <command>`?

bovine3dom: Not quite.

There's three levels to this:

An ex-alias is a shorthand for a longer ex-command, e.g.

```
:command top scrollto 0
:top # takes you to the top of the page
```

An ex-alias lets you rename an ex-command:

```
:command t tabopen
:t news.bbc.co.uk # opens BBC News
```

An ex-alias lets you create a new ex-command:

```
:command tabclosethenrecent composite tabclose ; tab #
:tabclosethenrecent # closes your tab then takes you to the most recent one
```

Under the hood `:command` works exactly the same in all cases. #3026 is asking how to control where the arguments given (like in the second case) go.

For example say I have 

```
scrollto x y
```

and I want to make a new command

`otllorcs y x` which is the does the same as `scrollto x y`.

An answer is

```
:command otllorcs js -d€ tri.excmds.scrollto(JS_ARG[1],JS_ARG[0])€
:scrollto 100 0
:otllorcs 0 100
```

Is that clearer? It's not really anything to do with composite but it could be used with it.

Rummskartoffel: I think I've got it now. Let's see if I can think of a good way of putting it into the doc.

Rummskartoffel: Nope. *I* know exactly what I mean, but I can't think of any good wording.

bovine3dom:
@Rummskartoffel I think this is closer to being useful, what do you think?

Rummskartoffel:
Looks fine to me. I still feel a vague sense of dissatisfaction with `:command`'s whole doc, but that may just be because I've spent too much time thinking about it.

Issue #2513, opened 2020-06-19
Re-enable the last few eslint rules


bovine3dom:
We have a few sensible eslint rules turned off. Fixing these probably isn't too difficult but might be a little time-consuming.

Recommended workflow: 

```
# In one terminal
ls .eslintrc.js src/**.js **.ts | entr -s '$(yarn bin)/eslint --fix src/'
# In another: make your fixes and re-enable rules
```

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L85

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L108

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L124

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L125

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L126

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L127

- [ ] https://github.com/tridactyl/tridactyl/blob/3a80fd03cbd84111958ee200cb9fd91c171f40b7/.eslintrc.js#L157

MyGitUsername:
I can start chipping away at this issue, one rule at a time.

bovine3dom:
That would be great : )

MyGitUsername:
If anyone else is interested in contributing to this issue, please feel welcome :)  
I haven't made any real progress here yet.  I'm planning to enable just one rule at a time and commit in small chunks to keep the work manageable as advised.

Rummskartoffel:
I think `@typescript-eslint/explicit-function-return-type` might also be worth considering adding to the backlog.

bovine3dom:
What's wrong with letting typescript work it out for us?

If there is an option to throw an error if it can't work it out and falls back to `any`, I'd much prefer that.

Rummskartoffel:
[...]

PR #3061, opened 2020-12-05
Explicitly annotate state.getAsync

Rummskartoffel:
cf. 0fb5430 and its comments.
I had to look up how to do this, and while it looks a bit weird, this does what we want here, allowing different return types for `getAsync` based on what the argument is.
On the topic of explicit annotations, I'd like to do some more annotating elsewhere in the codebase as well once this PR is done, because I've encountered quite a number of `any`s that could be avoided, which somewhat defeats the purpose of using types in the first place.

### DIFF
--- a/src/state.ts
+++ b/src/state.ts
@@ -91,7 +91,7 @@ const state = new Proxy(overlay, {
     },
 })
 
-export async function getAsync(property) {
+export async function getAsync<K extends keyof State>(property: K): Promise<State[K]> {
     if (notBackground())
         return browser.runtime.sendMessage({
             type: "state",
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -279,9 +279,7 @@ async function history(n) {
 
     // Check for matches in history, removing duplicates
     const matches = R.reverse(
-        // for some reason ramda loses the type information here
-        // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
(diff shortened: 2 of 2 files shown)


bovine3dom:
#2513 is probably related : )

src/state.ts
```diff
@@ -91,7 +91,7 @@ const state = new Proxy(overlay, {
     },
 })
 
-export async function getAsync(property) {
+export async function getAsync<K extends keyof State>(property: K): Promise<State[K]> {
     if (notBackground())
```
bovine3dom, line 95: If this is false, I don't think it returns a promise. We should probably wrap it in one for consistency? OTOH it's unlikely to ever be called from the background.

Rummskartoffel: It's an async function, so it can only return promises. Everything that isn't already a promise gets wrapped in one, while existing promises are returned as-is.

bovine3dom: Eugh, of course it is. I should not try to read code late at night. Thanks : )

Rummskartoffel:
> 
> 
> #2513 is probably related : )

I do think it'd be a good long-term goal to strictly enforce return type annotations for named functions at least. It might even be worth considering moving towards setting `noImplicitAny` in tsconfig.json to cover variables and parameters as well (because typescript-eslint doesn't seem to have very good rules for that), but I reckon it'd take a whole lot of work to get that to compile.

Issue #2749, opened 2020-08-22
Tridactyl breaks for the current tab when piping commands that use fillcmdline into yank or clipboard yank


Rummskartoffel:
Tridactyl version: 1.20.1
Firefox version: Mozilla Firefox 79.0
Operating system: win

*   Brief description of the problem: when using `composite` to pipe into `yank` or `clipboard yank`, Tridactyl breaks for the current tab until it is reloaded. It will still work on other tabs, but switching away from the broken tab and back again doesn't fix it.

*   Steps to reproduce:

    1. run a command such as `:composite get searchurls.google | yank`
    2. try to use any key binding (including `<C-,>`) or run any command

*   URL of the website the bug happens on: happens everywhere

*   Config: https://gist.github.com/Rummskartoffel/43a4ab6054be554ca29314727bab660b

bovine3dom:
I keep forgetting to check if I can reproduce this. Apparently I don't look at the "needs review" label as often as I thought

bovine3dom:
I can reproduce this but I think the blame should be shared equally between Tridactyl and you : )

Tridactyl shouldn't break when you do it, but you shouldn't be doing this - `get` is meant for interactive usage and opens the command line with the result. The return value of it is not meaningful.

The way to do what you want to do is via `js`:

```
composite js tri.config.get("searchurls","google") | yank
```

Note that where there was a `.`, we have split the arguments.


---

We could add a warning to `get`'s documentation. I think trying to stop Tridactyl from breaking might be too big an ask...

Rummskartoffel:
I don't think just telling users not to do it is the right way of dealing with this. At the very least, the command should display an error message, but after taking a quick look at the code, I think making `get` pipeable shouldn't be too hard. I'll try and come up with something.

bovine3dom:
Making get return a result wouldn't be hard. The thing that's hard (and I think the thing that's breaking stuff) is that `yank` uses the command line to copy things, so if you're trying to use the command line to display something at the same time, weird things will happen.

#2597 covers stopping `yank` from doing that.

Rummskartoffel:
Oh, now I see. In that case, this issue isn't specific to `get`, but happens more generally with piping `fillcmdline` or any command that uses `fillcmdline` into `yank`. I've tried it, and `composite fillcmdline abcde | yank` has the same effect as using `get`.

bovine3dom:
[...]

PR #3078, opened 2020-12-09
Update yank to use Clipboard API

### LINKED ISSUE #2597: Switch to clipboard API

Rummskartoffel:
(Hopefully) resolves #2597 and a step forward regarding #2749. I haven't extensively tested this yet – though with how simple this code is, I don't think there is a lot to test here.
I've removed `commandline_frame`'s `setClipboard`, `getClipboard`, and `applyWithTmpTextArea`, which are now obsolete. I've also done a small bit of cleaning up in the existing clipboard manipulation functions.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3103,10 +3103,10 @@ async function setclip(data: string) {
             promises = [setclip_selection()]
             break
         case "clipboard":
-            promises = [setclip_api(data)]
+            promises = [setclip_webapi(data)]
             break
         case "both":
-            promises = [setclip_selection(), setclip_api(data)]
+            promises = [setclip_selection(), setclip_webapi(data)]
             break
     }
     return Promise.all(promises)
@@ -3117,7 +3117,7 @@ async function setclip(data: string) {
  * @hidden
  */
 //#background_helper
-async function setclip_api(data: string) {
+async function setclip_webapi(data: string) {
     return window.navigator.clipboard.writeText(data)
 }
 
@@ -3130,7 +3130,7 @@ async function setclip_api(data: string) {
 export async function getclip(from?: "clipboard" | "selection") {
     if (from === undefined) from = await config.getAsync("putfrom")
     if (from === "clipboard") {
-        return getclip_api()
+        return getclip_webapi()
     } else {
         return Native.clipboard("get", "")
     }
@@ -3141,7 +3141,7 @@ export async function getclip(from?: "clipboard" | "selection") {
  * @hidden
  */
 //#background_helper
-async function getclip_api() {
+async function getclip_webapi() {
     return window.navigator.clipboard.readText()
 }
 
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3115,6 +3115,8 @@ async function setclip(data: string) {
 /**
  * Copies a string to the clipboard using the Clipboard API.
  * @hidden
+ *
+ * Has to be a background helper as it's only available on HTTPS and background pages. We want to be able to copy stuff to the clipboard from HTTP pages too.
  */
 //#background_helper
 async function setclip_webapi(data: string) {
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3095,18 +3095,18 @@ export function yank(...content: string[]) {
 //#background_helper
 async function setclip(data: string) {
     // Function to avoid retyping everything everywhere
-    const setclip_selection = () => Native.clipboard("set", data)
+    const setclip_selection = data => Native.clipboard("set", data)
 
-    let promises: Array<Promise<any>> = []
+    let promises: Promise<any>[]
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -3089,48 +3088,63 @@ export function yank(...content: string[]) {
...
+    return window.navigator.clipboard.writeText(data)
+}
+
 /**
  * Fetches the content of the clipboard/selection buffer depending on user's preferences
  *
  * Exposed for use with [[composite]], e.g. `composite getclip | fillcmdline`
  */
 //#background
-export async function getclip(fromm?: "clipboard" | "selection") {
```
bovine3dom, line 3125: `fromm` has been annoying me for years. I assumed it was because from was reserved. Has it just been a typo all these years?

src/excmds.ts
```diff
@@ -3089,48 +3088,63 @@ export function yank(...content: string[]) {
...
             break
     }
     return Promise.all(promises)
 }
 
+/**
+ * Copies a string to the clipboard using the Clipboard API.
+ * @hidden
+ */
+//#background_helper
```
bovine3dom, line 3119: It was surprising to me that this was a background command. It's presumably because of this:

> When using content scripts, the Clipboard API is available only for HTTPS pages. As a workaround, use messaging between your content scripts and the background script.

which we should document, I think : )

src/excmds.ts
```diff
@@ -3089,48 +3088,63 @@ export function yank(...content: string[]) {
...
+async function setclip_api(data: string) {
```
bovine3dom, line 3120:
```suggestion
async function setclip_webapi(data: string) {
```

The _api confused me as the primary selection also uses an API (arguably moreso since it's crossing a process boundary) : )

src/excmds.ts
Rummskartoffel: `from` *is* used as a keyword in import statements, but it [isn't reserved](https://www.ecma-international.org/ecma-262/#sec-keywords-and-reserved-words), probably because it can never be confused with an identifier. I've seen a few of these weirdly spelt names already – just today, I stumbled upon a `constructorr` in a context where `constructor` probably wouldn't have been reserved –, so it may just have been added by someone who *thought* it was reserved.

src/excmds.ts
Rummskartoffel: Well, about that. I first tried `content_helper` – I needed to access `window`, so that seemed the right place to go –, but that gave me a compile error. So I tried `content`, which gave me a run time `TypeError: <something> is not a function`. It just happened to work with `background_helper` when I tried it.

Rummskartoffel: By which I mean: I don't really understand how these macros work, I just winged it.

bovine3dom: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard is the page that tells you why you got a runtime error : )

No idea why you got a compile error with content_helper.

Rummskartoffel: Hmm. I thought I'd tried it on Tridactyl's new tab page, but I guess it must've been on a non-https page, unless `moz-extension://` doesn't count as https.
The error I got also didn't look like something Firefox would throw in that case, though. I'd assume Firefox would either throw something like a generic `Error: the Clipboard API is only available on HTTPS connections`, a `TypeError: window.navigator.clipboard.writeText is not a function`, or a `TypeError: window.navigator.clipboard is undefined`, but I remember that wasn't it. I think it was something along the lines of `TypeError: object[foobar] is not a function`.

bovine3dom: oh, it's because `setclip` is a background function and you can't call content commands from the background.

src/excmds.ts
```diff
@@ -3095,18 +3095,18 @@ export function yank(...content: string[]) {
 //#background_helper
 async function setclip(data: string) {
     // Function to avoid retyping everything everywhere
-    const setclip_selection = () => Native.clipboard("set", data)
+    const setclip_selection = data => Native.clipboard("set", data)
 
-    let promises: Array<Promise<any>> = []
+    let promises: Promise<any>[]
```
Rummskartoffel, line 3100: The reason I used `Array<>` instead of `[]` was because we have a linter rule that prohibits `[]` on complex types or something like that. Maybe we should get rid of that?

bovine3dom: Oh, yeah, I'll just turn it off.

PR #3085, opened 2020-12-10
Add image-to-clipboard hinting

### LINKED ISSUE #3079: Image to clipboard hint mode
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/clipboard

Not sure how to get the image to the background script, though.

Rummskartoffel:
Resolves #3079. So far, this only adds `:yankimage`, but the only thing left to do now is to add a default alias and/or bind.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3218,6 +3218,25 @@ export async function clipboard(excmd: "open" | "yank" | "yankshort" | "yankcano
     return done.then(() => undefined)
 }
 
+/** Copy an image to the clipboard.
+
+    @param url
+        Absolute URL to the image to be copied. You can obtain an absolute URL from a relative one using [tri.urlutils.getAbsoluteURL](_src_lib_url_util_.html#getabsoluteurl).
+*/
+//#background
+export async function yankimage(url: string) {
+    const absoluteUrl = UrlUtil.getAbsoluteURL(url, document.baseURI)
+    const image = await window.fetch(absoluteUrl)
+    const blob = await image.blob()
+    // Blob.type returns a MIME type like image/jpeg, but the Clipboard API expects only the second half
+    const imageType = blob.type.split("/")[1]
+    if (!["jpeg", "png"].includes(imageType)){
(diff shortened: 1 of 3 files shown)


src/excmds.ts
```diff
@@ -3218,6 +3218,25 @@ export async function clipboard(excmd: "open" | "yank" | "yankshort" | "yankcano
...
+export async function yankimage(url: string) {
+    const absoluteUrl = UrlUtil.getAbsoluteURL(url, document.baseURI)
+    const image = await window.fetch(absoluteUrl)
+    const blob = await image.blob()
+    // Blob.type returns a MIME type like image/jpeg, but the Clipboard API expects only the second half
+    const imageType = blob.type.split("/")[1]
+    if (!["jpeg", "png"].includes(imageType)){
+        throw new Error("Cannot copy image to clipboard: formats other than JPEG and PNG are not supported")
+    } else {
+        browser.clipboard.setImageData(await blob.arrayBuffer(), imageType as browser.clipboard._SetImageDataImageType)
```
bovine3dom, line 3236: What happens if an unsupported image type is given to the this? Presumably it throws a similar error. If that's the case I'd rather we didn't duplicate the error handling as that way if they ever add more formats we'll support it for free.

Rummskartoffel: The error thrown is:
> controller in excmd:  Error: Type error for parameter imageType (Invalid enumeration value "gif") for clipboard.setImageData. logging.ts:42:32
log logging.ts:42
acceptExCmd controller.ts:31
It's rather technical, but I expect your average Tridactyl user should be able to cope with that.

bovine3dom: I'd just wrap it in a try-catch, if imageType is mentioned in the error, throw a new error saying that it's an unsupported filetype and give the mimetype.

bovine3dom:
`:hint -cF img i => tri.excmds.yankimage(tri.urlutils.getAbsoluteURL(i.src))` seems to work a treat. I'll leave it to you to think of a bind : )

PR #3122, opened 2020-12-20
Fix :viewconfig for --default or --user with no key

Rummskartoffel:
The changes introduced in fa3d82c5 cause `:viewconfig` to throw a `TypeError` when using it with either of the flags, but no key, because it tries to call `.split` on `undefined`.


bovine3dom:
How did this one slip past me -_-

Thanks! Will merge once a test passes.

PR #3125, opened 2020-12-21
Persist gi

### LINKED ISSUE #3124: focusinput last input should persist until restart
Reckon we can use getSelector https://github.com/tridactyl/tridactyl/blob/9017049d235937140b7bdbb9825b0de751c3be93/src/lib/dom.ts#L301
instead of storing the element. Then we can share that across page loads. If the selector doesn't work, fall back to any old input.

This could make working with `gi` in two windows on different pages a bit more annoying, but I think my use case of

1. change the issue filter on github and accept it
2. realise that you messed up the filter and try to change it again

is probably more common

bovine3dom:
Fix #3124


src/state.ts
```diff
@@ -28,6 +28,7 @@ class State {
         },
     ]
     last_ex_str = "echo"
+    lastFocusInputSelectors: Map<number, string> = new Map() // used for focusinput, not for changing tabs like prevInputs
```
bovine3dom, line 31: I can't think of a good reason why this should be separate from prevInputs. I'll probably merge the two.

Issue #3131, opened 2020-12-25
Tridactyl breaks semi-permanently when using :seturl with a syntax error in the RegEx pattern


Rummskartoffel:
Tridactyl version: 1.20.4pre4968-2ead7df9
Firefox version: Mozilla Firefox 84.0.1
Operating system: win

Merry Christmas!

*   Brief description of the problem:

    When using `:seturl` with syntactically incorrect RegEx, Tridactyl immediately becomes completely unusable. Any kind of interaction has no effect save for writing swathes of `SyntaxError`s regarding the RegEx supplied to `:seturl` to the console. Additionally, the new tab page appears blank. Manually calling excmds through `window.tri` on one of Tridactyl's pages also doesn't work, same with the "tri" keyword in the address bar.
    The only way to remedy this problem I've found is to manually modify the config through `window.tri.browserBg.storage.local`. Restarting does not help, even when using an rc file that calls `:sanitise tridactyllocal`.

*   Steps to reproduce:

    Run any `:seturl` command with syntactically incorrect RegEx, e.g. `:seturl [ foo`. `:bindurl` has the same effect.

*   URL of the website the bug happens on:

    Mostly irrelevant. The only difference I've observed is that non-Tridactyl pages, i.e. any page on which `window.tri` is not defined, don't write the errors to the console, but they can still be seen in the extension toolbox.  

*   Config and rc file: N/A

This specific error originates from `config.getURL`, where the pattern set in the config is fed into `String.prototype.match` without prior validation or error handling. https://github.com/tridactyl/tridactyl/blob/2ead7df95a576d252ff50a8fe0e69c16e1d8e133/src/lib/config.ts#L1238
However, the larger issue in my opinion is that an error while loading the config can break the entire extension. A broken config should should just result in loading the default config and displaying an error message.

bovine3dom:
Can probably catch them easily by just making a new RegEx here https://github.com/tridactyl/tridactyl/blob/73ef32497d6b1fc7975d703753eca63de69e8af6/src/excmds.ts#L3434 and here https://github.com/tridactyl/tridactyl/blob/73ef32497d6b1fc7975d703753eca63de69e8af6/src/excmds.ts#L3527

but yeah, we should make Tridactyl robust to broken configs.

And merry Christmas! 🦌🦌🦌🛷🎅🎁🎁🎁

PR #3134, opened 2020-12-27
Prevent setURL from setting invalid RegEx patterns

Rummskartoffel:
Addresses the immediate problem with #3131 by preventing users from setting sub-configs with syntactically incorrect RegEx patterns. The more architectural issue of a broken config still remains. 

I figured throwing from `config.setURL` would be better than from the excmds. This way, we not only prevent the screwing over of oneself using `:seturl` and `:bindurl`, but also when using `config.setURL` in a `:js` alias. Of course, `config.set` or `browser.storage.local.set` can still insert invalid RegEx, but, at that point, you're probably doing it on purpose.

Lastly, the completely unrelated change in the platform defaults for Linux is because Prettier said so.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1142,8 +1142,10 @@ const platform_defaults = {
     },
     linux: {
         nmaps: {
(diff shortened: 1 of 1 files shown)


Rummskartoffel:
@bovine3dom 👀

bovine3dom:
Sorry, I'm AFK for a couple more days : )

PR #3142, opened 2020-12-29
Make ff_cmdline work on Windows

Rummskartoffel:
It certainly ain't pretty, and it takes around 800ms to execute on my end, but since the native messenger isn't a direct descendant of the main Firefox process, we need to work our way up the tree.
Currently, this splits right through quoted arguments containing spaces. I've done it that way for now because the non-Windows part of the function does the same, so I figured it might not be a problem (and I really couldn't be bothered to try and wrestle with cmd's quoting rules again when I might not need to).

### DIFF
--- a/src/lib/native.ts
+++ b/src/lib/native.ts
@@ -425,17 +425,17 @@ export async function clipboard(
  * That's rather slow, so this function shouldn't be called unless really necessary.
  */
 export async function ff_cmdline(): Promise<string[]> {
+    let output: MessageResp
     if ((await browserBg.runtime.getPlatformInfo()).os === "win") {
-        const output = await run(
+        output = await run(
             "powershell -Command " +
                 '"$ppid = Get-CimInstance -Property ProcessId,ParentProcessId Win32_Process | Where-Object -Property ProcessId -EQ $PID | Select-Object -ExpandProperty ParentProcessId;' +
                 "$pproc = Get-CimInstance -Property ProcessId,ParentProcessId,Name,CommandLine Win32_Process | Where-Object -Property ProcessId -EQ $ppid;" +
                 "while ($pproc.Name -notmatch 'firefox') { $ppid = $pproc.ParentProcessId;" +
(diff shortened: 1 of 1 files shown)


bovine3dom:
> it takes around 800ms to execute on my end

Do you have python on your PATH? If you don't, you'll be using the compiled native executable which is very slow.

src/lib/native.ts
```diff
@@ -416,18 +416,29 @@ export async function clipboard(
...
-        throw new Error(
-            `Error: "ff_cmdline() is currently broken on Windows and should be avoided."`,
+        const output = await run(
+            "powershell -Command " +
+                '"$ppid = Get-CimInstance -Property ProcessId,ParentProcessId Win32_Process | Where-Object -Property ProcessId -EQ $PID | Select-Object -ExpandProperty ParentProcessId;' +
+                "$pproc = Get-CimInstance -Property ProcessId,ParentProcessId,Name,CommandLine Win32_Process | Where-Object -Property ProcessId -EQ $ppid;" +
+                "while ($pproc.Name -notmatch 'firefox') { $ppid = $pproc.ParentProcessId;" +
+                '$pproc = Get-CimInstance Win32_Process | Where-Object -Property ProcessId -EQ $ppid}; Write-Output $pproc.CommandLine"',
         )
+        return output.content.trim().split(" ")
```
bovine3dom, line 436: any reason why you duplicated this rather than moving swapping line 443 and 444?

bovine3dom:
I don't have a way of testing this. LGTM - pinky promise me that it does really work on Windows now and I'll merge it. Thanks : )

src/lib/native.ts
Rummskartoffel: Nope, I just forgot to change it.

Rummskartoffel:
I do have Python in my PATH, but that's probably not the cause of the slowness, anyway. I think the main cause is probably `Get-CimInstance`, but there's no getting around that, as far as I can tell. I've tried remedying it a little by only retrieving the properties we actually need, but it's still pretty slow.
The only way I see this failing would be trying to use it on a Windows version that doesn't have PowerShell, all of which are long EOL, seeing as it's included since at least Windows 7.

PR #3154, opened 2021-01-01
Update webpack to version 5

bovine3dom:
Supersedes #3151.

The removal of all the node polyfills could have broken something - probably worth testing this a little bit thoroughly before merging.


bovine3dom:
I'll leave working out why the jobs aren't running for another day.

Issue #3150, opened 2020-12-31
:editor's internals could use some updating


Rummskartoffel:
There are currently several problems with `:editor`'s internals:
1. There is no default that works on a clean installation of Windows. Not much we can do about that, but we can at least add Notepad as a last resort.
1. TUI editors are only returned when a terminal is found to run them. This is desirable on non-Windows platforms, but not on Windows, since Windows automatically opens a terminal for any terminal application run with `start`, which is a cmd built-in and therefore always available.
1. On Windows, it is unnecessary to check for X11-dependent terminal emulators such as xterm or st, since they aren't going to be available anyway. Checking for them adds unnecessary performance overhead.
1. `inpath` currently uses the command line utilities `which` and `where`, depending on platform. It may be faster to use Python's `shutil.which` instead, to avoid having to call an external binary.
1. Similarly, it may be possible to improve `firstinpath`'s performance by sending the array to the native messenger and processing it there, instead of spawning the messenger anew for every element.

I've started looking into 1 through 3 and might shoot a PR soon™.

Lastly, I've found out that Alacritty [supports](https://github.com/alacritty/alacritty/blob/master/CHANGELOG.md#user-content-040) `--class` now, so we can update accordingly.

bovine3dom:
I am wondering if we should ditch python and go for something with less overhead and better binary support, like `Nim` perhaps? I don't think it should be as slow as it currently is.

I think a major reason we picked python was so that people could edit the messenger themselves if they wanted, but I don't think that is happening.

Rummskartoffel:
While I don't feel qualified to otherwise comment on possible language choices for a possible native messenger rewrite – or the rewrite itself, for that matter –, Nim in particular [supports cross-compiling](https://nim-lang.org/docs/nimc.html#crossminuscompilation-for-windows), which might help prevent such issues as #797.

Rummskartoffel:
Three are done, the other three would require changes to the native messenger. With how much you're working on that rewrite, @bovine3dom, I'm guessing it wouldn't really be worth still doing that in the Python version?

bovine3dom:
I hope it won't be worth doing : )

Nim should be quick enough that we won't care about how long things take.

If cross-compiling is somehow worse than Python we might still stick with Python.

Rummskartoffel:
I've tried cross-compiling to Windows and it works like a charm. It just depends on the native messenger being written in a compatible way, which I'll be putting a few remarks about on your draft PR.

[...]

PR #3163, opened 2021-01-02
Get Tridactyl ready for the nim native messenger

### LINKED ISSUE #3159: Make native messenger less slow and easier to distribute
As I mentioned in #3150.

I've been playing with Nim. Here's what I have so far:

```nim
# compile with 
# nim c --verbosity:0 native_main.nim
# (add -d:release for real builds)

# test with
# time printf '%c\0\0\0{"cmd": "run", "command": "echo $PATH"}' 39 | ./native_main

# compare with
# time printf '%c\0\0\0{"cmd": "run", "command": "echo $PATH"}' 39 | python ~/.local/share/tridactyl/native_main.py

import json
import options
import os
# https://nim-lang.org/docs/json.html

type 
    MessageResp* = object
        cmd*, version*, content*, error*, command*: Option[string]
        code: Option[int]

# let a = MessageResp(cmd: some(""))
# echo(a)

proc getMessage(): MessageResp =

    # length of the string - not required AFAICT
    var length: int32
    discard readBuffer(stdin, addr(length), 4)

    let command = to(parseJson(readAll(stdin)),MessageResp)
    
    # unclear how to get the output of this
    # execShellCmd(command.command.get())

    return command

echo(%* getMessage()) # %* converts the object to JSON
```

Currently runs in ~4ms compared to ~100ms for our Python native messenger. (`time python -c "1+1"` takes 50ms alone FWIW).

bovine3dom:
Fix #3159 and probably loads of other issues with native.

Currently 10x faster on Linux - probably 10-100x faster on Windows.

---

Should probably only build this on demand since most developers won't have `nim` installed and it will probably only need changing once every ~4 years like the current messenger.

Once it's working, will need to work out how to cross-compile and distribute the artifacts comfortably.


native/native_main.nim
```diff
@@ -0,0 +1,71 @@
...
+
+proc getMessage(): MessageResp =
+
+    # length of the string - not required AFAICT
+    var length: int32
+    discard readBuffer(stdin, addr(length), 4)
+
+    var command = to(parseJson(readAll(stdin)),MessageResp)
+    
+    command.content = some(execProcess(command.command.get(), options={poUsePath,poEvalCommand,poStdErrToStdOut}))
```
bovine3dom, line 31: poUsePath probably has no effect when used alongside poEvalCommand.

native/native_main.nim
```diff
@@ -0,0 +1,71 @@
...
+    MessageResp* = object
+        cmd*, version*, content*, error*, command*: Option[string]
+        code: Option[int]
+
+# let a = MessageResp(cmd: some(""))
+# echo(a)
+
+proc getMessage(): MessageResp =
+
+    # length of the string - not required AFAICT
```
cmcaine, line 25: We'd need to read only this many bytes if Firefox reuses a single process for multiple messages. But I think we have FF start a new native messenger process for every message.

cmcaine:
If Nim is a pain to cross compile, we might be better off with zig. It's another small language but it supports cross compilation very well.

glacambre:
Random thought: won't shipping binaries instead of python scripts cause notarization issues with MacOS?

bovine3dom:
> If Nim is a pain to cross compile, we might be better off with zig. It's another small language but it supports cross compilation very well.

I knew there was another small three-letter language that did cross-compiling well but I couldn't remember what it was.

---

I had forgotten about OSX. We update the messenger so rarely that we could keep the Python one around.

native/native_main.nim
bovine3dom: Yeah, we start a new native messenger each time at the moment. We could switch to a daemon based method but I'm hoping this will be quick enough to mean that we don't need to.

bovine3dom:
It looks like with OSX we can just add an `xattr` command to our install script to skip signature verification:

https://github.com/hashicorp/terraform/issues/23033#issuecomment-548688203

native/native_main.nim
```diff
@@ -0,0 +1,89 @@
...
+        of "run":
+            command.content = some(execProcess(msg.command.get(), options={poEvalCommand,poStdErrToStdOut}))
+            command.code = some(0) # need to use something other than execProcess to capture this
+
+        of "version":
+            command.version = some(VERSION)
+
+    return command
+
+write(stdout, %* handleMessage(getMessage())) # %* converts the object to JSON
```
bovine3dom, line 53: This apparently isn't sufficient for Firefox to understand it. Slightly perplexed as in my terminal the output appears similar, albeit with e.g. `"cmd": null` for all the missing values in Nim. Further investigation needed.

cmcaine:
There's also this thing for cross compiling on windows and osx: https://github.com/pololu/nixcrpkgs

native/native_main.nim
```diff
@@ -0,0 +1,97 @@
...
+
+let message = handleMessage(getMessage())
+
+# this works fine V so I reckon the problem must be with the "nulls" it spits out
+# could just write our own JSON? ... is that so mad?
+# let message = "{\"version\": \"0.1.11\"}" #$ %* handleMessage(getMessage()) # %* converts the object to JSON
+
+# but this doesn't work -_-
+# bleugh
+# let message = "{\"version\": \"0.2.0\"}" #$ %* handleMessage(getMessage()) # %* converts the object to JSON
```
bovine3dom, line 56: I don't like computers

cmcaine: https://www.youtube.com/watch?v=jpoFKZlQPV4

bovine3dom:
I had a quick look at Zig, @cmcaine, but its documentation isn't as good as Nim's IMO. I'll keep it as Plan B if deploying Nim is too difficult

cmcaine:
FWIW, last time I looked, the master docs were a bit nicer. The standard library docs are still quite sparse, though. They could do with some more docstrings and examples.

Anyway, good luck :) Ping me for review or comment if you want my ill-considered opinions!

native/native_main.nim
bovine3dom: Fixed it - I waited 15 hours and tried again and now it works. hooray

native/native_main.nim
```diff
@@ -0,0 +1,240 @@
...
+            close(strm)
+            quit(0)
+
+        let message = readStr(strm, length)
+        write(stderr, "Got message: " & message & "\n")
+        var raw_json = parseJson(message)
+
+        # Compatibility with Python native messenger:
+        # rename env's _var_ key to _variable_ coz _var_ is reserved in Nim
+        trySwapJsonKey(raw_json, "var", "variable")
```
Rummskartoffel, line 58: This isn't necessary, because Nim has [stropping](https://nim-lang.org/docs/manual.html#lexical-analysis-stropping), e.g.:
```nim
msg.`var`.get()
```

native/native_main.nim
```diff
@@ -0,0 +1,240 @@
...
+            write(stderr, "TODO: NOT IMPLEMENTED\n")
+
+        of "getconfigpath":
+            reply.content = findUserConfigFile()
+            reply.code = some(0)
+            if not isSome(reply.content):
+                reply.code = some(1)
+
+        of "run":
+            # this seems to use /bin/sh rather than the user's shell
```
Rummskartoffel, line 107: In the case of Windows, this appears to not use any normal shell at all, so you don't have access to any built-ins.

native/native_main.nim
bovine3dom: Oh, nice, thanks. I've never heard of that before. I thought Nim was being a bit greedy reserving common variable names : )

native/native_main.nim
```diff
@@ -0,0 +1,244 @@
...
+        of "env":
+            reply.content = some(getEnv(msg.`var`.get()))
+
+        of "list_dir":
+            write(stderr, "TODO: NOT IMPLEMENTED\n")
+
+        of "win_firefox_restart":
+            write(stderr, "TODO: NOT IMPLEMENTED\n")
+
+        of "ppid":
```
Rummskartoffel, line 177: This causes a compile error when targetting Windows, because it's part of the POSIX module and therefore not available. As far as I can tell, there isn't any good alternative, either, so we'd have to "manually" use the Windows API to make this work on Windows.
It might be best to just platform-gate it and avoid using `ppid` on Windows – as far as I understand it, Windows' process IDs aren't as useful as on *nix anyway.

native/native_main.nim
bovine3dom: Hmm, we should probably use `getenv "SHELL` on everything but Windows and hard-code `cmd.exe` there...

native/native_main.nim
bovine3dom: Apparently some of the `posix` module does work on Windows - https://github.com/nim-lang/Nim/issues/9372#issuecomment-564350044 - but yeah, we can just gate it since it's only used by `ff_cmdline` if the platform isn't Windows.

native/native_main.nim
Rummskartoffel: I only found out about it pretty much by pure chance, but I didn't expect it to actually come in handy.

Rummskartoffel:
See also https://github.com/tridactyl/tridactyl/issues/3150#issuecomment-754108912.

I compared the binaries produced by cross-compiling from Linux to Windows and by natively compiling with MinGW-w64 x32 after finally getting that to work. There doesn't appear to be any difference in performance. Regarding size, as of 5460b29, the cross-compiled binary is 294kB and the native one 382kB, both stripped. For further comparison, the bundled Python version is 4.9MB.

bovine3dom:
Eugh, I looked a bit more at `osxcross` and it looks like it needs some massive (~25GB) OSX SDK.

I'm thinking it might be easier to move the native messenger to a new repository and have GitHub actions build it natively on each OS and attach the artifacts to a release.

native/native_main.nim
Rummskartoffel: We should be able to gate it with something like this:
```nim
        of "ppid":
            when defined posix:
                reply.content = some($getppid())
            else:
                reply.cmd = some("error")
                reply.error = some("ppid is not available on this operating system")
```
Or we could just have it be handled by the else at the end and treat it as any other unknown command:
```nim
        of (when defined posix: "ppid" else: {}):
            reply.content = some($getppid())
```

bovine3dom: I think it might need to be `when defined getppid` because some POSIX stuff is available on Windows like `mkstemp`.

Rummskartoffel: `defined posix` doesn't check whether `posix` has been imported, it checks whether a compile-time "external symbol" called "posix" has been defined, which the compiler does automatically when targeting POSIX systems. What you're thinking of is called [`declared`](https://nim-lang.org/docs/system.html#declared%2Cuntyped) in Nim, but I've tried `when declared getppid` already, but that always returns true, regardless of whether it actually works on the platform you're targeting.

bovine3dom:
All further development of the binary component will happen here: https://github.com/tridactyl/native_messenger

This PR will just make compat changes where it's absolutely necessary.

Rummskartoffel (commented):
Terminological nitpick: *deprecated* implies it's still there, but use is discouraged. What we're doing is *removing* it from the Windows version entirely.

bovine3dom:
https://github.com/tridactyl/native_messenger/issues/3 probably needs to happen in this PR.

Also to do in this PR - update the help etc. to remove mentions of Python and provide a link to the repo so that people can inspect the code if they want.

bovine3dom:
I think what I'll do is merge this without changing the install scripts, so that people can test the Nim binaries by manually copying them over. If there's no problem we can then update the install scripts.

bovine3dom:
I've just remembered one reason why we wrote the messenger in Python before - it was so that it could be used in institutions where policies forbid running random binaries downloaded off from the internet, but random Python scripts are fair game.

I don't know in practice how many users that affected.

cmcaine:
Corporate security policies be like 🤡🤡🤡

bovine3dom:
Still do to in this PR: update documentation to remove mentions of Python, link to Nim repo / source code. Probably worth mentioning the old messenger for those behind security theatre.

src/lib/config.ts
```diff
@@ -844,7 +844,7 @@ export class default_config {
         excmd: "error",
         hinting: "warning",
         messaging: "warning",
-        native: "warning",
+        native: "debug",
```
bovine3dom, line 847:
```suggestion
        native: "warning",
```

src/lib/native.ts
```diff
@@ -435,12 +440,22 @@ export async function ff_cmdline(): Promise<string[]> {
                 '$pproc = Get-CimInstance Win32_Process | Where-Object -Property ProcessId -EQ $ppid}; Write-Output $pproc.CommandLine"',
         )
     } else {
-        output = await pyeval(
-            // Using ' and + rather than ` because we don't want newlines
-            'handleMessage({"cmd": "run", ' +
-                '"command": "ps -p " + str(os.getppid()) + " -oargs="})["content"]',
-        )
-        return output.content.trim().split(" ")
+
```
bovine3dom, line 443:
```suggestion
```

src/lib/native.ts
```diff
@@ -435,12 +440,22 @@ export async function ff_cmdline(): Promise<string[]> {
...
+        const actualVersion = await getNativeMessengerVersion()
+
```
bovine3dom, line 445:
```suggestion
```

src/lib/native.ts
```diff
@@ -346,6 +347,10 @@ export async function run(command: string, content = "") {
 
 /** Evaluates a string in the native messenger. This has to be python code. If
  *  you want to run shell strings, use run() instead.
+ *
+ *  __DEPRECATED!__
```
bovine3dom, line 351:
```suggestion
```

bovine3dom:
https://github.com/tridactyl/tridactyl/blob/1522f0900cabfd8d95dfa223231772fee61938b7/e2e_tests/e2e.test.ts#L156

1. Stop installing the native messenger on CI
2. Be surprised when E2E tests which include testing the native messenger all fail

sigh

PR #3170, opened 2021-01-03
Overhaul native.getBestEditor (#3150)

Rummskartoffel:
* Only check whether commands exist if they can exist to avoid unnecessary overhead
* add some more options, mostly for Windows
* set window class on some more terminals that support it.

I also tried to make it so on Windows a TUI editor just gets `start`ed when no third-party terminal emulator is found, but I haven't yet gotten that to work. Somehow, Python's `Popen` seems to be very thoroughly preventing that.

Also change formatting in a few unrelated places because Prettier said so.

### DIFF
--- a/src/lib/native.ts
+++ b/src/lib/native.ts
@@ -58,12 +58,17 @@ async function sendNativeMsg(
     }
 }
 
-export async function getrcpath(separator: "unix" | "auto" = "auto"): Promise<string> {
+export async function getrcpath(
+    separator: "unix" | "auto" = "auto",
+): Promise<string> {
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM - why is it a draft?

Rummskartoffel:
Because I when I submitted the PR I had another idea about I might solve the `start` thing, but I didn't have time to try it out immediately. So l decided to get the bulk of it submitted already, while keeping it as a draft so I could add it in case it worked – it didn't. Then I just forgot to mark the PR ready for review.

PR #3184, opened 2021-01-05
Add :tabaudio and `ga` binding.

### LINKED ISSUE #3183: Excmd for going to audio playing tab
It would be useful to have an excmd+keybind for going to the tab currently playing audio. Would you guys accept a PR for this? I'd be happy to take a crack at it.

Thanks for tridactyl! It's the main reason I switched back to Firefox :)

aduros:
Fixes #3183

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2163,6 +2163,15 @@ export async function tabpush(windowId: number) {
     return activeTabId().then(tabId => browser.tabs.move(tabId, { index: -1, windowId }))
 }
 
+/** Switch to the tab currently playing audio, if any. */
+//#background
+export async function tabaudio() {
+    const tabs = await browser.tabs.query({ currentWindow: true, hidden: false, audible: true })
+    if (tabs.length > 0) {
+        return browser.tabs.update(tabs[0].id, { active: true })
+    }
+}
+
 /**
  * Moves all of the targetted window's tabs to the current window. Only works for windows of the same type
  * (can't merge a non-private tab with a private window).
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2166,8 +2166,9 @@ export async function tabpush(windowId: number) {
 /** Switch to the tab currently playing audio, if any. */
 //#background
 export async function tabaudio() {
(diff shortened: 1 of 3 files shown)


src/excmds.ts
```diff
@@ -2163,6 +2163,15 @@ export async function tabpush(windowId: number) {
     return activeTabId().then(tabId => browser.tabs.move(tabId, { index: -1, windowId }))
 }
 
+/** Switch to the tab currently playing audio, if any. */
+//#background
+export async function tabaudio() {
+    const tabs = await browser.tabs.query({ currentWindow: true, hidden: false, audible: true })
```
bovine3dom, line 2169: I'm wondering if it should operate across all windows

aduros: I'm not sure. What should happen if the other windows aren't visible, such as when minimized or on a different desktop?

aduros: We could request focus on the audio window, which seems to work nicely in i3 at least. If the window is on the same desktop, it switches focus, otherwise it sets a needs-attention hint on the desktop it's on.

For minimized windows I guess they could be unminimized. I don't have a way to test it though.

bovine3dom: I think just setting the tab to active (in whichever window) and letting Firefox worry about the rest is fine : )

bovine3dom:
It now activates tabs regardless of where they are. Let me know if you think that's too confusing.

I'm wondering if `ga` should throw an error if no audible tabs are found. It could help to convince users that it isn't Firefox that's making the noise (and that Tridactyl isn't broken).

We could provide `tabaudio --quiet` to silence the warnings for non-interactive usage.

aduros:
Sounds good to me :) Thanks.

Are you sure about removing hidden too? What is a "hidden" tab anyways? I assumed it could mean a background page from another extension but not sure. Is it possible to make a hidden tab active?

bovine3dom:
Hidden tabs are real tabs that are hidden by extensions, for example in the "tab groups" style extensions.

I haven't checked that they can be activated but I assume they can. I'm not sure hidden tabs can play audio either, but if they can it seems like a great reason for `ga` to work on them : )

PR #3185, opened 2021-01-05
Display an icon for audible tabs.

aduros:
Refs #2811.

### DIFF
--- a/src/completions/Tab.ts
+++ b/src/completions/Tab.ts
@@ -43,13 +43,16 @@ class BufferCompletionOption
         const favIconUrl = tab.favIconUrl
             ? tab.favIconUrl
             : Completions.DEFAULT_FAVICON
+        const indicator = tab.audible
+            ? String.fromCodePoint(0x1F50A)
+            : ""
         this.html = html`<tr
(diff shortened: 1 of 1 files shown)


src/completions/Tab.ts
```diff
@@ -43,13 +43,16 @@ class BufferCompletionOption
         const favIconUrl = tab.favIconUrl
             ? tab.favIconUrl
             : Completions.DEFAULT_FAVICON
+        const indicator = tab.audible
+            ? String.fromCodePoint(0x1F50A)
+            : ""
         this.html = html`<tr
             class="BufferCompletionOption option container_${container.color} container_${container.icon} container_${container.name}"
         >
             <td class="prefix">${pre.padEnd(2)}</td>
```
bovine3dom, line 52: I was imagining the indicator going here, since it's quite important information and tab titles can be truncated on small screens

aduros: What do you think of putting it before the title but after the index?

bovine3dom: Oh, in the same field as the title? That could work. It would make future filtering harder but there's no reason why we couldn't just move it later if we ever got round to that.

PR #3189, opened 2021-01-06
Remove Windows Terminal from default Windows terminals again

Rummskartoffel:
Turns out WT, when run, apparently spawns another instance of itself and then immediately exits, with the newly spawned instance being the one that's actually displayed. Or something like that. That makes it not suited as a terminal for `:editor` after all, because once the first process exits, Tridactyl will think the user is finished writing and read the file, resulting in an empty text box. There doesn't seem to exist any option to pass to wt.exe to make it not do that, so we can't use it.

### DIFF
--- a/src/lib/native.ts
+++ b/src/lib/native.ts
@@ -156,7 +156,7 @@ export async function getBestEditor(): Promise<string> {
         }
         if (os === "win") {
             term_emulators.push(
-                ...["wt", "conemu -run", "mintty --class tridactyl_editor -e"],
+                ...["conemu -run", "mintty --class tridactyl_editor -e"],
             )
         }
         // These terminal emulators are cross-platform.


bovine3dom:
Is wt a new name for cmd?

doesn't `cmd /c start /wait` or something like that work? (NB: it has been a long time since I last had to care about this sort of thing : ))

Rummskartoffel:
Windows Terminal is Microsoft's new terminal emulator. It's pretty good – except of course this weird behaviour.

I've taken another look , and apparently `wt.exe` is just a launcher for the actual terminal which exits immediately. Invoking the *real* binary isn't possible because of missing access rights (while it's in theory possible to modify those, messing with system-owned files really isn't something a browser extension of all things should be doing), so there's no way to wait on the process we're actually interested in.
Oh well. Windows users would've gotten Notepad as their default `:editor` anyway in almost all cases, and WT enthusiasts will just have to make do without their favourite terminal.

bovine3dom:
Did we ever try `start /wait cmd /c vim`? Or is cmd.exe so bad that it's not worth trying?

Rummskartoffel:
I tried `start` with the old messenger and that didn't work, unfortunately. I could get it to start the process I wanted, sure, but it started in the background, without a window. And I wasn't able to find a way to force cmd to open a new window.
As for the new messenger, `start` is not (yet) available – but something along the lines of `cmd /c start /wait vim` or even `cmd /c start /wait cmd /c vim` doesn't do the trick, either.
I'm sure there must be some way of explicitly opening a new terminal, but I don't know it and haven't been able to find it. So, for now, it seems it's only possible to use `:editor` on Windows with a GUI app or a third-party terminal.

Rummskartoffel:
So there is [*a* way](https://docs.microsoft.com/en-us/windows/console/creation-of-a-console) to open a console, but that would require us to manually use the Windows API and write platform-specific code in the native messenger. Nim provides access to [`CreateProcessW`](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw), the function that should be able to accomplish that, [in the `winlean` module](https://nim-lang.org/docs/winlean.html#createProcessW,WideCString,WideCString,ptr.SECURITY_ATTRIBUTES,ptr.SECURITY_ATTRIBUTES,WINBOOL,int32,WideCString,WideCString,STARTUPINFO,PROCESS_INFORMATION), but it'd probably not be worth it.

bovine3dom:
Thanks for looking into it. I think you're right and that it probably isn't worth it : )

Issue #2677, opened 2020-07-29
Changelistjump array rather than one-off


bovine3dom:
I'm sure we have an issue for this already but I can't find it.

I think a fairly simple solution is this:

Stack of inputs + tabs used. When g; pressed pop from that stack. Whenever an input is entered, set a timer with e.g. 1 second to add that to the stack. If g; is pressed cancel the timer.

Need to add handling so that g; just skips over tabs which have closed or been navigated away from, but that might be tricky so leave it as something to big fixed later.

PR #3249, opened 2021-01-20
WIP: first go at change list jumping again

bovine3dom:
Related: #2677

Currently breaks `g;` - it just does `focusinput -l` every time

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2070,6 +2070,7 @@ export async function changelistjump() {
     const tail = state.prevInputs[state.prevInputs.length - 1]
     const jumppos = tail.jumppos ? tail.jumppos : state.prevInputs.length - 1
     const input = state.prevInputs[jumppos]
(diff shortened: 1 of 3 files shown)


src/excmds.ts
```diff
@@ -2070,6 +2070,7 @@ export async function changelistjump() {
     const tail = state.prevInputs[state.prevInputs.length - 1]
     const jumppos = tail.jumppos ? tail.jumppos : state.prevInputs.length - 1
     const input = state.prevInputs[jumppos]
+    state.storeNextInput = "false"
```
bovine3dom, line 2073: Should check if `input` is real before setting this

bovine3dom: Probably worth making a `setAsync` function so that we can wait for this to be successfully set.

https://github.com/tridactyl/tridactyl/blob/99941d796ea665c04accff61e2dea1e1bcb2ead0/src/state.ts#L100

PR #3262, opened 2021-01-23
Migrate taball functionality into tab and alias them

### LINKED ISSUE #3208: Merge `:tab` and `:taball` commands
I don't see any reason why these shouldn't be the same command underneath - they should just have different completions.

jgkamat:
Closes #3208

The case where a nonint string is passed dosen't give a very good message right now, but hopefully I will fix that later.

Not sure if I should try to remove tabIndexSetActive or try to call it from inside parseWinTabIndex, let me know what you prefer. 

Unfortunately the e2e tests don't seem to complete on my machine (on master) so I wasn't able to test those myself. 

Also not sure if I should be duplicating the docstring or doing something else.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3281,30 +3281,31 @@ export async function yankimage(url: string): Promise<void> {
 
 /** Change active tab.
 
-    @param index
-        Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+    @param id
+        A bare number means the current window is used. Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+
+        A string following the following format: "[0-9]+.[0-9]+" means the first number being the index of the window that should be selected and the second one being the index of the tab within that window. [[taball]] has completions for this format.
 
         "#" means the tab that was last accessed in this window
-
-    This is different from [[taball]] because `index` is the position of the tab in the current window.
  */
 //#background
-export async function tab(index: number | "#") {
-    return tabIndexSetActive(index)
-}
+export async function tab(id: string) {
+    if (Number.isInteger(Number(id)))
+        return tabIndexSetActive(Number(id));
+    if (id === "#")
+        return tabIndexSetActive(id);
 
-/** Change active tab.
-
-    @param id
-        A string following the following format: "[0-9]+.[0-9]+", the first number being the index of the window that should be selected and the second one being the index of the tab within that window.
-
- */
-//#background
-export async function taball(id: string) {
     const [winid, tabindex_number] = await parseWinTabIndex(id)
     const tabid = (await browser.tabs.query({ windowId: winid, index: tabindex_number }))[0].id
     await browser.windows.update(winid, { focused: true })
     return browser.tabs.update(tabid, { active: true })
+}
+
+/** Wrapper for [[tab]] with multi-window completions
+ */
+//#background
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -3281,30 +3281,38 @@ export async function yankimage(url: string): Promise<void> {
 
 /** Change active tab.
 
-    @param index
-        Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+    @param id
+        A string following the following format: "[0-9]+.[0-9]+", the first number being the index of the window that should be selected and the second one being the index of the tab within that window.
```
bovine3dom, line 3285: Somewhere in this I'd say that taball has completions for this so you don't need to guess window IDs etc

src/excmds.ts
```diff
@@ -3281,30 +3281,38 @@ export async function yankimage(url: string): Promise<void> {
...
+    if (id === "#")
+        return tabIndexSetActive(id);
+
+    const [winid, tabindex_number] = await parseWinTabIndex(id)
+    const tabid = (await browser.tabs.query({ windowId: winid, index: tabindex_number }))[0].id
+    await browser.windows.update(winid, { focused: true })
+    return browser.tabs.update(tabid, { active: true })
 }
 
 /** Change active tab.
```
bovine3dom, line 3304: I'd just have this docstring say that it is a wrapper for [[tab]] with multi-window completions

src/excmds.ts
```diff
@@ -3281,30 +3281,38 @@ export async function yankimage(url: string): Promise<void> {
...
 
-    @param index
-        Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+    @param id
+        A string following the following format: "[0-9]+.[0-9]+", the first number being the index of the window that should be selected and the second one being the index of the tab within that window.
 
-        "#" means the tab that was last accessed in this window
+        A bare number means the current window is used.
```
bovine3dom, line 3287: I think 99% of people will still be using it in this way, so this should come first. The x.y format is almost an aside by comparison

bovine3dom:
Yeah, I've never been able to get the E2E tests to work on Arch. They can take a very long time to run though (like, an hour), so it's probably good that you didn't manage it. It has worked fine on the CI : )

src/excmds.ts
```diff
@@ -3281,32 +3281,33 @@ export async function yankimage(url: string): Promise<void> {
 
 /** Change active tab.
 
-    @param index
-        Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+    @param id
+        A bare number means the current window is used.
```
bovine3dom, line 3285: We should still document what 0, -1 etc do

src/excmds.ts
```diff
@@ -3281,32 +3281,33 @@ export async function yankimage(url: string): Promise<void> {
...
 //#background
-export async function tab(index: number | "#") {
-    return tabIndexSetActive(index)
-}
-
-/** Change active tab.
-
-    @param id
-        A string following the following format: "[0-9]+.[0-9]+", the first number being the index of the window that should be selected and the second one being the index of the tab within that window.
+export async function tab(id: string) {
```
bovine3dom, line 3292: This should probably be `...id: string[]` so e.g. `:tab BBC news` works and doesn't silently drop the news

PR #3263, opened 2021-01-23
Add support for looking up tabs by url or title in tab excmd

jgkamat:
Depends on #3262

This is pretty simple but some implementation details could be done differently depending on what you prefer

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3321,11 +3321,62 @@ export async function yankimage(url: string): Promise<void> {
         A string following the following format: "[0-9]+.[0-9]+" means the first number being the index of the window that should be selected and the second one being the index of the tab within that window. [[taball]] has completions for this format.
 
         "#" means the tab that was last accessed in this window
+
+        A non integer string means to search the URL and title for matches, in this window if called from tab, all windows if called from anytab. Title matches can contain '*' as a wildcard.
  */
 //#background
-export async function tab(id: string) {
+export async function tab(...id: string[]) {
+    return tab_helper(true, false, ...id)
+}
+
+/** Wrapper for [[tab]] with multi-window completions
+ */
+//#background
+export async function taball(...id: string[]) {
+    return tab_helper(true, true, ...id);
+}
+
+/** Helper to change active tab. Used by [[tab]] and [[taball]].
+
+    @param interactive
+        Controls if we should prompt if multiple matches are found, or just pick the first match
+
+    @param anyWindow
+        True if we should search in all windows, or just the current one.
+
+    @param key
+        String or int tab search key, see [[tab]] for usage.
+ */
+//#background
+export async function tab_helper(interactive: boolean, anyWindow: boolean, ...key: string[]) {
+    const id = key.join(" ")
     if (Number.isInteger(Number(id))) return tabIndexSetActive(Number(id))
     if (id === "#") return tabIndexSetActive(id)
+
+    if (id !== null && id !== undefined && !/\d+\.\d+/.exec(id)) {
+        let defaultQuery = {}
+        if (!anyWindow)
+            defaultQuery = {windowId: (await activeTab()).windowId};
+
+        const results = new Map()
+        try {
+            (await browser.tabs.query({...defaultQuery, ...{url: id}})).forEach(tab => results.set(tab.id, tab))
+        } catch (e) { }
+        if (results.size < 2)
+            (await browser.tabs.query({...defaultQuery, ...{title: id.replace("*", "\\*")}})).forEach(tab => results.set(tab.id, tab))
+        if (results.size < 2)
+            (await browser.tabs.query(defaultQuery)).filter((tab => tab.url.includes(id))).forEach(tab => results.set(tab.id, tab))
+        if (results.size < 2)
+            (await browser.tabs.query({...defaultQuery, ...{title: "*" + id + "*"}})).forEach(tab => results.set(tab.id, tab))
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -3281,30 +3281,61 @@ export async function yankimage(url: string): Promise<void> {
...
+        } catch (e) { }
+        if (!results.length)
+            results = await browser.tabs.query({title: id.replace("*", "\\*")})
+        if (!results.length) {
+            const urlResult = (await browser.tabs.query({})).find((tab => tab.url.includes(id)))
+            if (urlResult)
+                results = [urlResult]
+        }
+        if (!results.length)
+            results = await browser.tabs.query({title: "*" + id + "*"})
```
jgkamat, line 3313: I thought it would be neat to let people specify their own wildcards so I didn't escape * here. I think that's the only special character for this function, the docs were not clear on what exactly this is.

src/excmds.ts
```diff
@@ -3281,30 +3281,61 @@ export async function yankimage(url: string): Promise<void> {
...
-        Starts at 1. 0 refers to last tab of the current window, -1 to penultimate tab, etc.
+    @param id
+        A string following the following format: "[0-9]+.[0-9]+", the first number being the index of the window that should be selected and the second one being the index of the tab within that window.
+
+        A bare number means the current window is used.
 
         "#" means the tab that was last accessed in this window
 
-    This is different from [[taball]] because `index` is the position of the tab in the current window.
+        A non integer string means to search the URL and title for matches
```
bovine3dom, line 3291: "Any other string searches for matches and switches to the first one."

I think I would expect `:tab` to only switch to matches in the current window and `:taball` to switch to matches in any window.

src/excmds.ts
```diff
@@ -3281,30 +3281,61 @@ export async function yankimage(url: string): Promise<void> {
...
+            results = await browser.tabs.query({title: id.replace("*", "\\*")})
+        if (!results.length) {
+            const urlResult = (await browser.tabs.query({})).find((tab => tab.url.includes(id)))
+            if (urlResult)
+                results = [urlResult]
+        }
+        if (!results.length)
+            results = await browser.tabs.query({title: "*" + id + "*"})
+        if (results.length) {
+            await browser.windows.update(results[0].windowId, { focused: true })
```
bovine3dom, line 3315: I'm wondering if the right thing to do in case of multiple matches is to run `fillcmdline_notrail("tabs", query)` to allow the user to pick?

Occasionally in situations like this we give commands an argument to say if they are interactive or not so scripts can still do something useful unattended

src/excmds.ts
bovine3dom: We should mention it in the docstring.

src/excmds.ts
```diff
@@ -3281,30 +3281,61 @@ export async function yankimage(url: string): Promise<void> {
...
 
         "#" means the tab that was last accessed in this window
 
-    This is different from [[taball]] because `index` is the position of the tab in the current window.
+        A non integer string means to search the URL and title for matches
  */
 //#background
-export async function tab(index: number | "#") {
-    return tabIndexSetActive(index)
+export async function tab(id: string) {
```
bovine3dom, line 3294: This should probably be `...id: string[]` so e.g. `:tab BBC news` works and doesn't silently drop the news

src/excmds.ts
```diff
@@ -3281,32 +3281,78 @@ export async function yankimage(url: string): Promise<void> {
...
+        try {
+            results = results.concat(await browser.tabs.query({...defaultQuery, ...{url: id}}))
+        } catch (e) { }
+        if (results.length < 2)
+            results = results.concat(await browser.tabs.query({...defaultQuery, ...{title: id.replace("*", "\\*")}}))
+        if (results.length < 2)
+            results = results.concat(await browser.tabs.query(defaultQuery)).filter((tab => tab.url.includes(id)))
+        if (results.length < 2)
+            results = results.concat(await browser.tabs.query({...defaultQuery, ...{title: "*" + id + "*"}}))
+        if (results.length) {
```
bovine3dom, line 3339: In practice this feels like the filtering might be a bit too loose:

```
yarn run build && yarn run run # starts Tridactyl with the new tab page and txti.es open
:set completions.Tab.autoselect false
:tab txti # :taball txti is prefilled. I expected to be taken to the txti.es tab
```

src/excmds.ts
```diff
@@ -3281,32 +3281,78 @@ export async function yankimage(url: string): Promise<void> {
...
+            if (interactive && results.length > 1)
+                return fillcmdline_notrail(interactive ? "taball" : "tab", id)
```
bovine3dom, line 3341:
```suggestion
                return fillcmdline_notrail(anyWindow ? "taball" : "tab", id)
```

: )

bovine3dom:
Note to self: this can be merged trivially one `pretty` has been run on it - no thinking required

src/excmds.ts
jgkamat: oops, I meant for this to be a set, hopefully this will fix it. I think `tab.id` is unique in the browser session, is that true?

bovine3dom: Yes. I think : )

bovine3dom:
Seems to work fine now, thanks!

As a very minor quibble - I'd prefer it if you didn't force-push changes to PRs that I'd already looked at once. It means I have to review the whole thing again rather than just the additional changes : )

PR #3274, opened 2021-01-24
Fix completion for bmark when -t is given

jgkamat:
When -t is given on the ex cli with bmarks, completions don't seem to
pop up in the same way as they do without -t.

Not sure if prefix+query is needed, as prefix is just thrown away in the next fn anyway

I also left the first set to lastExStr in case the fn exits early.

I'm not sure what -c is for in bmarks, I couldn't find any docs for that.

### DIFF
--- a/src/completions/Bmark.ts
+++ b/src/completions/Bmark.ts
@@ -76,6 +76,7 @@ export class BmarkCompletionSource extends Completions.CompletionSourceFuse {
             .slice(0, 10)
             .map(page => new BmarkCompletionOption(option + page.url, page))
 
+        this.lastExstr = prefix + query
         return this.updateChain()
     }
 


bovine3dom:
`-c` is so hard to use it's more like an easter egg. I certainly didn't know about it.

https://github.com/tridactyl/tridactyl/blob/af3f4b3937ea22ac932fbdacaeac7fd90887dcfe/src/excmds.ts#L1244

so `:bmarks -t -c [container] [bookmark]` should open that bookmark in the specified container. NB: `-tc`, `-c [container] -t` etc. won't work.

We _really_ need a proper argument parser

bovine3dom:
Oh, I let someone persuade me that space should select completions in `:bmarks` a while ago so you need to do `:set completions.Bmark.autoselect false` before you can even type that out.

bovine3dom:
I've improved the help a bit in 90dd599904266a3a35a24fe687a484574423c6c1

PR #3292, opened 2021-01-28
WIP: fix #1449 - stack themes

### LINKED ISSUE #1449: Import other theme/css files in theme

bovine3dom:
Usage: `:set themes ["themebase", "thememiddle", "themetop"]`

`:set themes ["dark", "quakelight"]` works well already.

Todo: documentation and user-friendliness (e.g. how do we want `:colours` to work? Need to add a "themes" change listener. Existence of :colours, :set theme and :set themes will 100% confuse users. Should document clearly that :set theme (and therefore :colours) will be ignored if `themes` exists)

@noctuid if you still use Tridactyl, how did you imagine this working?

### DIFF
--- a/src/content/styling.ts
+++ b/src/content/styling.ts
@@ -48,46 +48,49 @@ export async function theme(element) {
         insertedCSS = false
     }
 
-    const newTheme = await config.getAsync("theme")
+    const newThemes = await config.getAsync("themes")
+    if (newThemes.length == 0) newThemes.push(await config.getAsync("theme"))
 
-    /**
-     * DEPRECATED
-     *
-     * You don't need to add weird classnames to your themes any more, but you can if you want.
-     *
-     * Retained for backwards compatibility.
-     **/
-    if (newTheme !== "default") {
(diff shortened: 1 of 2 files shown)


noctuid:
Are `set theme` and `colorscheme theme` exactly the same? I think it would be preferable if there was just one setting and either a list or a single value could be specified. If that's not viable, I think it would be fine to document that `themes` overrides `theme`.

bovine3dom:
They're not the same - that's the source of confusion. `colourscheme` installs themes if it needs to - from disk or URL - and then runs `set` to set which installed theme should be used.

So if you want to use this with your own themes you'd need to do

```
colourscheme mytheme1
colourscheme mytheme2
set themes ["mytheme1", "mytheme2"]
```

Perhaps we should separate the relevant part out into a new themeinstall command.

PR #3303, opened 2021-01-31
Don't reinstall themes all the time

bovine3dom:
I was trying to make `:colours [theme you have already installed from the internet]` less confusing.

`:colours --refresh [theme from disk]` is now required if you wish to reload a theme from disk.

I think this might be too confusing to people making themes.

I'm wondering if it would be better instead to fall back to using a pre-existing `customthemes.[themename]` only if it fails to load the theme from disk.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -445,7 +445,9 @@ export function cssparse(...css: string[]) {
 
 /** @hidden */
 //#background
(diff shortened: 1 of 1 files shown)


bovine3dom:
Yeah, `--refresh` was too confusing. This way is better - it tries to load from internal storage if loading from disk fails.

PR #3306, opened 2021-01-31
WIP: silly first go at own reader mode

### LINKED ISSUE #102: Keybindings lost in Reader View
While reading https://www.redhat.com/archives/libvir-list/2017-November/msg00525.html , I switched over to the Reader View but lost my ability to use `j` and `k` to scroll. :(

bovine3dom:
Once finished will fix #102 (which I think would be the lowest numbered issue closed for a long time).

@hughwilliams94 - Do you fancy finishing this / polishing it up? I hope this is enough to get you started. Key flaws in this implementation: 1) doesn't appear to have any decent CSS. 2) there's no way to exit the mode (ideally the back button would work)

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -451,7 +451,7 @@ export async function loadtheme(themename: string) {
     // remove the "tridactylrc" bit so that we're left with the directory
     const path = (await Native.getrcpath()).split(separator).slice(0, -1).join(separator) + separator + "themes" + separator + themename + ".css"
     const file = await Native.read(path)
-    if (file.code !== 0)  {
+    if (file.code !== 0) {
         if (Object.keys(await config.get("customthemes")).includes(themename)) return
         throw new Error("Couldn't read theme " + path)
     }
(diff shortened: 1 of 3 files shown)


hughwilliams94:
Thanks for having a go, I'll try to take a look at this when I have a moment

src/excmds.ts
```diff
@@ -5096,6 +5096,15 @@ export async function extoptions(...optionNameArgs: string[]) {
     return winopen("-popup", selectedExtension.optionsUrl)
 }
 
+//#content_helper
+import { Readability } from "@mozilla/readability"
+
+//#content
+export async function reader2() {
+    const article = new Readability(document).parse()
+    document.body.innerHTML = article.content
```
bovine3dom, line 5105: Probably easiest to make a data uri - https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs - out of `article.content` and then use `open(` to open it. or maybe just do `document.location.href = [data URI]`. That way we'll get a useful back button.

PR #3309, opened 2021-02-01
Fix #3097 : add hintautoselect option

### LINKED ISSUE #3097: Add option to disable auto hint acceptance when only a single match
Tridactyl version: 1.20.3
Firefox version: Mozilla Firefox 83.0
Operating system: linux

Add a way to not auto accept a match in hint mode even if it is the only match, and always require to validate.

The idea is to forgive mistakes by always having the possibility to go back and re-type the hint chars.

fluem:
Fix #3097: This adds a boolean option `hintautoselect` (default:`true`) like @bovine3dom suggested. If set to `false`, hints won't be followed automatically even if there is only one link on the page or only one matched hint left.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -467,11 +467,12 @@ export function hintPage(
         )
 
     if (
-        modeState.hints.length == 1 ||
-        (firstTargetIsSelectable() && allTargetsAreEqual())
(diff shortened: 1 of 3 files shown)


edisondotme:
Awesome, thank you! One of the first settings I was searching for after installing this recently.

PR #3326, opened 2021-02-04
Add container support for winopen

### LINKED ISSUE #1126: winopen should support containers
I get sad every time I do `winopen -c emacs` and it doesn't work : (

fluem:
Fixes #1126. ```winopen``` now supports ```-c containername``` option. Any additional options like ```-private``` or ```-popup``` are incompatible and therefore ignored. Also catching error for ```tabopen``` if no container name is given.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2771,7 +2771,8 @@ export async function winopen(...args: string[]) {
         if (firefoxArgs === "--private-window") {
             throw new Error("Can't open a container in a private browsing window.")
         } else {
-            return composite(`tabopen -c ${address} ; sleep 100 ; tabdetach`)
+            args.unshift("-c")
+            return tabopen(...args).then(() => tabdetach())
         }
     }
 
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2291,6 +2291,7 @@ export async function tabopen(...addressarr: string[]): Promise<browser.tabs.Tab
             args.shift()
             argParse(args)
         } else if (args[0] === "-c") {
+            if (args.length < 2) throw new Error(`You must provide a container name!`)
             // Ignore the -c flag if incognito as containers are disabled.
             if (!win.incognito) {
                 if (args[1] === "firefox-default" || args[1].toLowerCase() === "none") {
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -2740,15 +2743,20 @@ export async function winopen(...args: string[]) {
...
             case "-popup":
                 createData.type = "popup"
-                args = args.slice(1, args.length)
+                args.shift()
                 break
 
+            case "-c":
+                if (args.length < 2) throw new Error(`You must provide a container name!`)
+                args.shift()
+                return composite(`tabopen -c ${args.join(" ")} | sleep 100 | tabdetach`)
```
bovine3dom, line 2758:
```suggestion
                return composite(`tabopen -c ${args.join(" ")}; sleep 100; tabdetach`)
```

`;` means run one after the other, `|` means pipe the results of each one into the next.

I don't like the `sleep 100` so if we can avoid it at all I'd like that.

src/excmds.ts
```diff
@@ -2740,15 +2743,20 @@ export async function winopen(...args: string[]) {
         switch (args[0]) {
             case "-private":
                 createData.incognito = true
-                args = args.slice(1, args.length)
+                args.shift()
                 firefoxArgs = "--private-window"
```
bovine3dom, line 2747: If `-private` and `-c` are given, we should throw an error. We don't want people to accidentally open things they think are private in a normal container. : )

bovine3dom (commented):
Thanks for this PR. I'd really like to see if we can avoid the `sleep`.

I think I opened an issue a while ago about making `tabopen()` returning a promise that would be fulfilled once the tab had loaded; that could be a good solution.

src/excmds.ts
fluem: Agree, I made it fail explicitly.

src/excmds.ts
fluem: @bovine3dom: Are referring to #2698?

I think something along the lines of 
`
return tabopen(args).then(() =>  tabdetach)
`
should already work, but I'll give it a try.

bovine3dom: I think I'm referring to the problem I was fixing in that PR - that is, that `openInTab` doesn't work very well because `await [make a new tab]` returns more quickly than Tridactyl loading in that tab.

We probably need to make Tridactyl fire an event to the background script that says it has been loaded in a tab. We can then listen for that event and run `openInTab` then?

That said, `openInTab` runs in the background so it shouldn't matter ...

Give it your way a go and see what happens : )

fluem: Sleepless in ```winopen```. It seems to work - was surprised myself.

bovine3dom:
Haha, @fleum, I just realised that I was the one who suggested the `sleep` in #1126. Sorry for changing my mind ... in my defense it was a long time ago.

Thanks for fixing it up - it works great : )

PR #3339, opened 2021-02-07
Avoid time paradox when undo tabs

fluem:
Issue #3302. Sorting the recently closed tabs takes care of @glacambre's case, but there are still cases with three tabs or multiple time jumps where the order is wrong. Have to look at this again.

### DIFF
--- a/src/completions/Sessions.ts
+++ b/src/completions/Sessions.ts
@@ -7,13 +7,13 @@ function computeDate(session) {
         ((new Date() as any) - session.lastModified) / 1000,
     )
     let qualifier = "s"
-    if (howLong > 60) {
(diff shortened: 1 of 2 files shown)


glacambre (commented):
LGTM, thanks for fixing this! :)

bovine3dom:
Happy to merge this now and you can think about the edge cases some other time, if you want. Either way is OK with me

fluem:
I sorted the wrong way round and it screws up the normal timeline. Have to figure it out a different way. Sorry, no ready to merge yet.

fluem:
Removed the incorrect sorting, but kept the nicer time formatting for negative times.

PR #3347, opened 2021-02-09
Add --overwrite and --cleanup options to 'saveas'

gsbabil:


src/excmds.ts
```diff
@@ -883,9 +883,23 @@ export async function restart() {
...
+
+        if (uniq_args_set.has("--overwrite")) {
+            overwrite = true
+            uniq_args_set.delete("--overwrite")
+        }
+
+        if (uniq_args_set.has("--cleanup")) {
+            cleanup = true
+            uniq_args_set.delete("--cleanup")
+        }
```
bovine3dom, line 900: These need `nativegate( [whatever version of the Nim messenger gets them, probably 0.3?] )` : )

src/lib/native.ts
```diff
@@ -343,8 +343,21 @@ export async function temp(content: string, prefix: string) {
...
-    return sendNativeMsg("move", { from, to }).catch(e => {
+export async function move(
+    from: string,
+    to: string,
+    overwrite: boolean,
+    cleanup: boolean,
+) {
+    const requireNativeMessengerVersion = "0.3.0"
+
+    if (!(await nativegate(requireNativeMessengerVersion, false))) {
```
bovine3dom, line 354:
```suggestion
    if (overwrite && cleanup && !(await nativegate(requireNativeMessengerVersion, false))) {
```

`move` works fine without these new parameters

src/lib/native.ts
```diff
@@ -343,8 +343,21 @@ export async function temp(content: string, prefix: string) {
...
+        throw new Error(
+            `'restart' on Windows needs native messenger version >= ${requireNativeMessengerVersion}.`,
```
bovine3dom, line 356: : )

src/lib/native.ts
```diff
@@ -343,8 +343,21 @@ export async function temp(content: string, prefix: string) {
...
 
-export async function move(from: string, to: string) {
-    return sendNativeMsg("move", { from, to }).catch(e => {
+export async function move(
+    from: string,
+    to: string,
+    overwrite: boolean,
+    cleanup: boolean,
+) {
+    const requireNativeMessengerVersion = "0.3.0"
```
bovine3dom, line 352: Note to self: don't merge this without setting this to the ultimately required version

src/lib/native.ts
gsbabil: You are welcome. 🙃

src/lib/native.ts
gsbabil: @bovine3dom `nativegate` for version `0.3.0` was added in https://github.com/gsbabil/tridactyl_native_messenger/commit/043488b3cff5943678ed62bbbddda7ca3bf90ad6. Thanks.

src/lib/native.ts
gsbabil: @bovine3dom 

- `move` with `--overwrite` and/or `--cleanup` should work as expected in https://github.com/gsbabil/tridactyl_native_messenger/commit/043488b3cff5943678ed62bbbddda7ca3bf90ad6. 

-  https://github.com/gsbabil/tridactyl_native_messenger/commit/043488b3cff5943678ed62bbbddda7ca3bf90ad6 also fixes the notification problem by using `messageActiveTab("commandline_frame", "fillcmdline" ...` So we have success/failure notifications for the user if the download/move succeeded or failed. Thanks.

src/excmds.ts
gsbabil: @bovine3dom Added in https://github.com/gsbabil/tridactyl_native_messenger/commit/043488b3cff5943678ed62bbbddda7ca3bf90ad6

src/lib/native.ts
bovine3dom: Are you saying we should stop users who can currently use `:saveas` from using `:saveas` because `:saveas` will continue to not return success messages unless they update?

I don't think that's a good enough reason to force people to update. Some people with bizarre corporate security ~~theatre~~ policies may not be able to use the Nim binary and will have to continue to use the Python script, so I want to keep that supported as far as possible.

Am I missing something?

gsbabil: @bovine3dom 

> Are you saying we should stop users who can currently use :saveas from using :saveas because :saveas will continue to not return success messages unless they update?

No. As mentioned in Matrix, my test-case was download attempts made on single files, and a notificaiton is useful from the user's perspective to know when/if the downloade completed successfully.

If this conflicts with multiple file download, I'm happy to migrate to a different notification approach where the user still gets to learn which file(s) downloaded property and which ones failed.

gsbabil: @bovine3dom 

> Am I missing something?

As long as we are talking about a single file download, I don't fully understand where the conflict actually is. 

As said, my only goal from user's point of view is to notify when/if the download completed. I'm saying let's show the user something on "success", as well as on "error. The code before was only throwing error, no notification on "success".

bovine3dom: I think we're talking past each other. Without this patch, say I have native 0.1.11:

`:saveas foo` # works

With this patch, native 0.1.11

`:saveas foo` # doesn't work: you need to update your native messenger.

This is bad, right?

gsbabil: @bovine3dom 

> I think we're talking past each other.

I thought you'd be pulling from my `master` where the latest currently points to https://github.com/gsbabil/tridactyl/commit/a508f432f20cdb875509402a8ee70d5aa3cdbaea

Can you you please confirm which commit you are reviewing? Thanks.

gsbabil: @bovine3dom 

> With this patch, native 0.1.11

> `:saveas foo` # doesn't work: you need to update your native messenger.`

This should work. 

Let me check `0.1.11` native messenger with the `--overwrite` and `--cleanup` patches at https://github.com/gsbabil/tridactyl/commit/a508f432f20cdb875509402a8ee70d5aa3cdbaea. Thanks.

bovine3dom: https://github.com/gsbabil/tridactyl/commit/ee7e339951120e43fe5f807a704a793b8da316c7 is what I'm looking at

src/lib/native.ts
gsbabil: @bovine3dom Fixed in https://github.com/gsbabil/tridactyl/commit/c17ac418266995de2f2540ee1178ee79eb41074a

src/lib/native.ts
gsbabil: @bovine3dom Better Python/Nim compatibility added in https://github.com/gsbabil/tridactyl/commit/c17ac418266995de2f2540ee1178ee79eb41074a

src/lib/native.ts
gsbabil: @bovine3dom 

> gsbabil@ee7e339 is what I'm looking at

Please review https://github.com/gsbabil/tridactyl/commit/c17ac418266995de2f2540ee1178ee79eb41074a it should support both old `saveas` behaviour with Python and new `saveas` with Nim. Thanks.

bovine3dom:
Closing in favour of #3362 : )

PR #3357, opened 2021-02-11
Re-word doc for gimode and make it display correctly

Rummskartoffel:
The angle brackets in `<Tab>` were being interpreted as HTML tags, so they need to be escaped.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -793,9 +793,9 @@ export class default_config {
     ttspitch = 1
 
     /**
-     * If nextinput, <Tab> after gi brings selects the next input
+     * When set to "nextinput", pressing &lt;Tab&gt; after gi selects the next input.
      *
-     * If firefox, <Tab> selects the next selectable element, e.g. a link
+     * When set to "firefox", &lt;Tab&gt; behaves like normal, focusing the next tab-indexed element regardless of type.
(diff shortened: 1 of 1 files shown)


src/lib/config.ts
```diff
@@ -793,9 +793,9 @@ export class default_config {
     ttspitch = 1
 
     /**
-     * If nextinput, <Tab> after gi brings selects the next input
+     * When set to "nextinput", pressing &lt;Tab&gt; after gi selects the next input.
```
bovine3dom, line 796:
```suggestion
     * When set to "nextinput", pressing `<Tab>` after gi selects the next input.
```

src/lib/config.ts
```diff
@@ -793,9 +793,9 @@ export class default_config {
...
      *
-     * If firefox, <Tab> selects the next selectable element, e.g. a link
+     * When set to "firefox", &lt;Tab&gt; behaves like normal, focusing the next tab-indexed element regardless of type.
```
bovine3dom, line 798:
```suggestion
     * When set to "firefox", `<Tab>` behaves like normal, focusing the next tab-indexed element regardless of type.
```

bovine3dom (approved):
Thanks : )

PR #3362, opened 2021-02-14
Add --overwrite and --cleanup options to 'saveas'

bovine3dom:
Supersedes #3347

### DIFF
--- a/src/background/download_background.ts
+++ b/src/background/download_background.ts
@@ -140,7 +140,7 @@ export async function downloadUrlAs(
                     if (operation.code != 0) {
                         reject(
                             new Error(
-                                `# 🔴 '${downloadItem.filename}' could not be moved to '${saveAs}' (${code2human(operation.code)}. Code: ${operation.code}) ...`,
+                                `${code2human(operation.code)}. '${downloadItem.filename}' could not be moved to '${saveAs}'. Error code: ${operation.code}`,
                             ),
                         )
                     } else {
--- a/src/background/download_background.ts
+++ b/src/background/download_background.ts
@@ -2,6 +2,7 @@
  * Background download-related functions
  */
 
+import * as Messaging from "@src/lib/messaging"
 import * as Native from "@src/lib/native"
 import * as config from "@src/lib/config"
 import { getDownloadFilenameForUrl } from "@src/lib/url_util"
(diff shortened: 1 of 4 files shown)


src/background/download_background.ts
```diff
@@ -78,14 +80,21 @@ export async function downloadUrl(url: string, saveAs: boolean) {
...
+ * @param URL the URL to download
  * @param saveAs If beginning with a slash, this is the absolute path the document should be moved to. If the first character of the string is a tilda, it will be expanded to an absolute path to the user's home directory. If saveAs begins with any other character, it will be considered a path relative to where the native messenger binary is located (e.g. "$HOME/.local/share/tridactyl" on linux).
- * If saveAs points to a directory, the name of the document will be inferred from the URL and the document will be placed inside the directory. If saveAs points to an already existing file, the document will be saved in the downloads directory but wont be moved to where it should be ; an error will be thrown. If any of the directories referred to in saveAs do not exist, the file will be kept in the downloads directory but won't be moved to where it should be.
+ * @param If true, overwrite the destination file, returns error code 1 otherwise if file exists
+ * @param If true, cleans up temporary downloaded source file e.g. in $HOME/Downlods/downloaded.doc when the move operation fails e.g. due to target destination exists, OS error etc.
```
bovine3dom, line 86: No user will ever find this documentation - we should move it to `:saveas` in excmds

src/excmds.ts
```diff
@@ -883,9 +883,28 @@ export async function restart() {
  * @param filename The name the file should be saved as.
  */
 //#content
-export async function saveas(...filename: string[]) {
-    if (filename.length > 0) {
-        return Messaging.message("download_background", "downloadUrlAs", window.location.href, filename.join(" "))
+export async function saveas(...args: string[]) {
+    let overwrite = false
+    let cleanup = false
+    const uniqueArgs = new Set(args)
```
bovine3dom, line 889: This is going to cause crazy bugs: `:saveas ~/Documents/this is a silly name for a folder/` won't work.

We should use the `while()` and `args.shift()` technique that we do elsewhere (in `hint()` perhaps?)

bovine3dom:
Removed "blocked" label as we can safely merge this before a compatible native messenger is released, providing that we set `requiredMessengerVersion` to a number we _know_ we aren't going to use until it is compatible

src/excmds.ts
gsbabil: @bovine3dom 

Will try to have a look as time permits. Thanks.

For the record, it seems that we have multiple problems here:

1. We can't pass more than one space in file/directory names
     - This is likely lingering from before with the `filename.join(" ")` pattern
     
2. We are losing repeating sub-strings in file/directory names e.g. `~/Documents/this is a silly name for a folder` becomes `~/Documents/this is a silly name for folder`

bovine3dom: I can look myself if you're sick of this code : )

I only left a comment because I spotted it and didn't want to forget. Thanks for all your work on this!

bovine3dom: > We can't pass more than one space in file/directory names
> This is likely lingering from before with the filename.join(" ") pattern

Oh god, why did you have to spot this, this is a fundamental issue with the command line

PR #3363, opened 2021-02-14
Add tabclosepinned option to config tabclose on pinned tabs

bew:
:warning: NOT TESTED

Solves #344 

---
NOTE: Another implementation possibility would be to change the function `idFromIndex` to `tabFromIndex` and return a Tab object, which would make it more generic.

And then in tabclose we'd return tabs from `maybeWinTabToTabId` (changing its name to `maybeWinTabToTab`) and make the pinned check earlier.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2519,14 +2519,22 @@ export async function tabclose(...indexes: string[]) {
         }
         return idFromIndex(id)
     }
+    let ids
     if (indexes.length > 0) {
-        const ids = await Promise.all(indexes.map(index => maybeWinTabToTabId(index)))
-        done = browser.tabs.remove(ids)
+        // Request to close multiple tabs
+        ids = await Promise.all(indexes.map(index => maybeWinTabToTabId(index)))
     } else {
-        // Close current tab
-        done = browser.tabs.remove(await activeTabId())
+        // Request to close the current tab
+        ids = [await activeTabId()]
     }
-    return done
+    const tabclosepinned = (await config.getAsync("tabclosepinned") === "true")
+    for (let tab_id of ids) {
+        const tab = (await browser.tabs.query({index: tab_id}))[0]
+        if (tab.pinned && !tabclosepinned) {
+            throw new Error(`Tab $tab_id is pinned and tabclosepinned is false, aborting tabclose`)
+        }
+    }
+    return browser.tabs.remove(ids)
 }
 
 /** Close all tabs to the side specified
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2511,7 +2511,6 @@ export async function fullscreen() {
 */
 //#background
 export async function tabclose(...indexes: string[]) {
-    let done
     async function maybeWinTabToTabId(id: string) {
         if (id.includes(".")) {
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -2519,14 +2519,22 @@ export async function tabclose(...indexes: string[]) {
...
-        done = browser.tabs.remove(ids)
+        // Request to close multiple tabs
+        ids = await Promise.all(indexes.map(index => maybeWinTabToTabId(index)))
     } else {
-        // Close current tab
-        done = browser.tabs.remove(await activeTabId())
+        // Request to close the current tab
+        ids = [await activeTabId()]
+    }
+    const tabclosepinned = (await config.getAsync("tabclosepinned") === "true")
```
bovine3dom, line 2530: You only need to use getAsync if your function is likely to be called within a few milliseconds of Tridactyl loading in the tab - get is fine otherwise

src/excmds.ts
```diff
@@ -2519,14 +2519,22 @@ export async function tabclose(...indexes: string[]) {
...
+    for (let tab_id of ids) {
```
bovine3dom, line 2531: We don't need to do this loop if tabclosepinned is true, so I'd move the check up to here

src/excmds.ts
bew: you never know :D sometimes I open and close tabs really fast

src/excmds.ts
bew: You mean, put the loop in a `if (!tabclosepinned) { ... }` ?

bew: Done

src/excmds.ts
```diff
@@ -2511,22 +2521,24 @@ export async function fullscreen() {
...
-    } else {
-        // Close current tab
-        done = browser.tabs.remove(await activeTabId())
+    const tabs = await Promise.all(indexes.length > 0 ? indexes.map(maybeWinTabToTabId) : [activeTab()])
+    const tabclosepinned = (await config.getAsync("tabclosepinned")) === "true"
+    if (!tabclosepinned) {
+        // Pinned tabs should not be closed, abort if one of the tabs is pinned
+        for (const tab of tabs) {
+            if (tab.pinned) {
+                throw new Error(`Tab ${tab.windowId}:${tab.index} is pinned and tabclosepinned is false, aborting tabclose`)
```
bovine3dom, line 2537:
```suggestion
                throw new Error(`Tab ${tab.windowId}:${tab.index + 1} is pinned and tabclosepinned is false, aborting tabclose`)
```

I'm not sure how to deal with `windowId` - I think it doesn't correspond to our window indexes. We could make `tabs` into a `winindex.tabindex: tabid` object but I don't think it's worth the effort

src/excmds.ts
```diff
@@ -2511,22 +2521,24 @@ export async function fullscreen() {
 */
 //#background
 export async function tabclose(...indexes: string[]) {
-    let done
     async function maybeWinTabToTabId(id: string) {
```
bew, line 2524: Since you're making these changes, the function name should be renamed to remove `Id` at the end

PR #3369, opened 2021-02-15
Make Windows use new native messenger

bovine3dom:
@Rummskartoffel does this look right to you? : )


Rummskartoffel:
Should be good

PR #3386, opened 2021-02-19
Fix a few trivial no-unsafe-call errors

Rummskartoffel:
See #2513. These are just some really trivial, mostly isolated ones.

### DIFF
--- a/src/background/commands.ts
+++ b/src/background/commands.ts
@@ -1,4 +1,4 @@
-import * as useractions from "@src/background/user_actions"
+import { useractions } from "@src/background/user_actions"
 import * as config from "@src/lib/config"
 import * as keyseq from "@src/lib/keyseq"
 import * as controller from "@src/lib/controller"
@@ -10,7 +10,7 @@ function makelistener(commands: Array<browser.commands.Command>) {
             "browsermaps",
             keyseq.mozMapToMinimalKey(command.shortcut).toMapstr(),
         )
-        if (exstring in useractions && typeof useractions[exstring] === "function") return (useractions[exstring] as () => void)()
+        if (exstring in useractions) useractions[exstring]()
         return controller.acceptExCmd(exstring)
     }
 }
--- a/src/background/commands.ts
+++ b/src/background/commands.ts
@@ -3,14 +3,14 @@ import * as config from "@src/lib/config"
 import * as keyseq from "@src/lib/keyseq"
 import * as controller from "@src/lib/controller"
 
-function makelistener(commands) {
+function makelistener(commands: Array<browser.commands.Command>) {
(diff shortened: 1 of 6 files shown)


src/background/commands.ts
```diff
@@ -3,14 +3,14 @@ import * as config from "@src/lib/config"
...
-function makelistener(commands) {
+function makelistener(commands: Array<browser.commands.Command>) {
     return (command_name: string) => {
         const command = commands.filter(c => c.name == command_name)[0]
         const exstring = config.get(
             "browsermaps",
             keyseq.mozMapToMinimalKey(command.shortcut).toMapstr(),
         )
-        if (exstring in useractions) return useractions[exstring]()
+        if (exstring in useractions && typeof useractions[exstring] === "function") return (useractions[exstring] as () => void)()
```
bovine3dom, line 13: This feels like the wrong place to be doing this. Can't we specify `useractions: Function[]` at source?

Rummskartoffel: Well, `useractions` is a module we're importing at the top of the file, so we'd have to do this:
```typescript
function escapehatch(){ /* ... */ }
export { escapehatch }
```
We'd still have to keep the type assertion to be able to call `useractions[exstring]`, but since `useractions` seems to only exist to implement `escapehatch`, anyway, we might as well just hard-code it and dodge both of these problems:
```suggestion
        if (exstring === "escapehatch") useractions.escapehatch()
```

bovine3dom:
```
export const useractions: Record<string, Function> {
escapehatch: () => { ... },
}
```

?

I hope it is only ever used for escapehatch but it's possible Mozilla will add more useful things that have to be user actions.

Rummskartoffel: [`@typescript-eslint/ban-types`](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md) doesn't like `Record<string, Function>`, but `Record<string, () => void>` does the trick.

src/background/commands.ts
```diff
@@ -1,16 +1,16 @@
...
-function makelistener(commands) {
+function makelistener(commands: Array<browser.commands.Command>) {
     return (command_name: string) => {
         const command = commands.filter(c => c.name == command_name)[0]
         const exstring = config.get(
             "browsermaps",
             keyseq.mozMapToMinimalKey(command.shortcut).toMapstr(),
         )
-        if (exstring in useractions) return useractions[exstring]()
+        if (exstring in useractions) useractions[exstring]()
```
bovine3dom, line 13: typo? If you don't return here `escapehatch` will execute twice. (It's also an ex-command)

PR #3390, opened 2021-02-20
shydactyl | make use of more variables and defaults

eeshugerman:
this does not change the appearance of the theme in any way (not intentionally, at least, and I compared side-by-side everything I could think to)


src/static/themes/shydactyl/shydactyl.css
```diff
@@ -28,22 +40,14 @@
...
 :root #completions {
-    --option-height: 1.4em;
-    color: black;
-    background: white;
-    display: inline-block;
-    font-size: unset;
     font-weight: 200;
-    overflow: hidden;
-    width: 100%;
-    border-top: unset;
```
eeshugerman, line 46: most of these are the same of the defaults

PR #3440, opened 2021-03-04
Tweak Matrix link on new tab

jezcope:
Hey folks, I love this extension! Hope you don't mind the tiny drive-by PR, but it's now recommended to use [matrix.to](https://matrix.to) for Matrix room links as this is flexible across a whole range of clients so I'm just proposing a change to the link on the new-tab page. :slightly_smiling_face:


jezcope:
Spotted that there are a few more links to the Matrix room, so updated those too.

PR #3443, opened 2021-03-04
Windows: fix infinite loop in native.ff_cmdline()

Rummskartoffel:
Because of backward compatibility, we still need to keep the old, slow method of looping up the process tree, but that respects 'browser' now.

Fixes tridactyl/native_messenger#40.


src/lib/native.ts
```diff
@@ -471,21 +471,33 @@ export async function clipboard(
...
+            /*
+            If browser binary is not "firefox.exe" but 'browser' is not set,
+            this will hang and use a lot of CPU.
+            See https://github.com/tridactyl/native_messenger/issues/40
+            */
+            output = await run(
+                `powershell -Command \
+"$ppid = Get-CimInstance -Property ProcessId,ParentProcessId Win32_Process | Where-Object -Property ProcessId -EQ $PID | Select-Object -ExpandProperty ParentProcessId;\
+$pproc = Get-CimInstance -Property ProcessId,ParentProcessId,Name,CommandLine Win32_Process | Where-Object -Property ProcessId -EQ $ppid;\
+while ($pproc.Name -notmatch '${await config.get(
```
bovine3dom, line 488: Can't we error out here if no process is called `browser`? That seems like it would fix 95% of cases.

bovine3dom: Woops, this comment should be one line higher

Rummskartoffel: Should be easy enough.

Rummskartoffel:
Added exit condition, improved formatting, added `-NoProfile`, made the code a bit more terse and idiomatic, and squashed.

src/lib/native.ts
```diff
@@ -471,21 +471,34 @@ export async function clipboard(
...
+if (-not ($processes | where { $_.Name -match '^${browser}' })) { exit 1; };\
+$ppid = ($processes | where { $_.ProcessId -EQ $PID }).ParentProcessId;\
+$pproc = $processes | where { $_.ProcessId -EQ $ppid };\
+while ($pproc.Name -notmatch '^${browser}') {\
+    $ppid = $pproc.ParentProcessId;\
+    $pproc = $processes | where { $_.ProcessId -EQ $ppid };\
+};\
+Write-Output $pproc.CommandLine;\
+"`,
+            )
```
bovine3dom, line 492:
```suggestion
            const browser_name = await config.get("browser")
            output = await run(
                `powershell -NoProfile -Command "\
$processes = Get-CimInstance -Property ProcessId,ParentProcessId,Name,CommandLine -ClassName Win32_Process;\
if (-not ($processes | where { $_.Name -match '^${browser_name}' })) { exit 1; };\
$ppid = ($processes | where { $_.ProcessId -EQ $PID }).ParentProcessId;\
$pproc = $processes | where { $_.ProcessId -EQ $ppid };\
while ($pproc.Name -notmatch '^${browser_name}') {\
    $ppid = $pproc.ParentProcessId;\
    $pproc = $processes | where { $_.ProcessId -EQ $ppid };\
};\
Write-Output $pproc.CommandLine;\
"`,
            )
```


Replacing the `browser` object seems like a trap

Rummskartoffel: I'd noticed that and considered it irrelevant, since I don't see us ever needing `browser` in that particular scope, but there's no reason not to not shadow it, either, so I don't care – fine with me.

PR #3479, opened 2021-03-15
Add browser compat linter

### LINKED ISSUE #2460: Check browser compatibility in tests

bovine3dom:
This currently seems to ignore the `browser` object, which is 95% of the reason we're trying to do this.

Further investigation needed.

Closes #2460 once finished, I hope.

(test with `yarn run eslint src/**.ts`)

Edit:

the package ast-metadata-inferer, upstream of eslint-plugin-compat, doesn't support webextensions. this PR uses our fork which attempts to add support

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ commands:
       - restore_cache:
           keys:
             - dependency-cache-{{ checksum "package.json" }}
-      - run: yarn install --frozen-lockfile
+      - run: yarn install
       - save_cache:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ commands:
       - restore_cache:
           keys:
             - dependency-cache-{{ checksum "package.json" }}
-      - run: yarn install
+      - run: yarn install --frozen-lockfile
       - save_cache:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
     "extends": [
         "plugin:@typescript-eslint/recommended",
         "plugin:@typescript-eslint/recommended-requiring-type-checking",
+        "plugin:compat/recommended",
         "prettier",
         "plugin:sonarjs/recommended"
     ],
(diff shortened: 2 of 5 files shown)


bovine3dom:
Not the error I was expecting to find

```
It's likely that the plugin isn't installed correctly. Try reinstalling by running the following:

    npm install eslint-plugin-compat@latest --save-dev

The plugin "eslint-plugin-compat" was referenced from the config file in ".eslintrc.js".
```

Guess I messed up somewhere in https://github.com/tridactyl/eslint-plugin-compat/tree/webext or https://github.com/bovine3dom/ast-metadata-inferer/tree/webext

bovine3dom:
https://github.com/bovine3dom/ast-metadata-inferer/runs/6575659968?check_suite_focus=true#step:7:79 failed but worked fine on my machine :/

NicolasWebDev:
@bovine3dom I am not able to find the package @tridactyl/eslint-plugin-compat anywhere on npm (see the error 404 on https://www.npmjs.com/package/@tridactyl/eslint-plugin-compat). I see that it was forked from https://github.com/amilajack/eslint-plugin-compat (by the way, is the fork really necessary? Knowing that there is no commit yet).

I would guess that you have those packages in your local machine, but they are not published to npm, or at least not publicly.

bovine3dom:
Thanks for looking at it :)

as far as I can tell if you put `foo/bar` in package.json rather than a version number, NPM and Yarn look for that package on GitHub rather than NPM.

There are commits on both forks - maybe you're looking at the wrong branches?

charlie39:
I ventured into it , the build runs okay my machine to but still the same problem in circleCI.
It doesn't install the GitHub package/repo.
I am also clueless but what's the problem but I am trying to take a different angle here. Why not have a separate run in the workflow in the jobs and attempt an explicit install?:

Something like 
```
...
run : yarn add https://github.com/tridactyl/eslint-plugin-compat#webext
...
```
Just a thought

bovine3dom:
Thanks, that would be a good thing to try :)

bovine3dom:
A friend pointed out that rather than fighting with the dependency management I should just vendor (i.e. copy paste) the code

bovine3dom:
Closing since glacambre's version of this works great and is merged

PR #3495, opened 2021-03-18
Add POC jsonviewer that works anywhere

bovine3dom:
1) doesn't work at the moment
2) should make this part of `tabopen` to handle data: URIs

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4183,12 +4183,10 @@ export function get(...keys: string[]) {
 
 /** Opens the current configuration in Firefox's native JSON viewer in the current tab.
  *
- * NB: only works well from a "normal" page, e.g not a Tridactyl help or new tab page.
- *
  * @param key - The specific key you wish to view (e.g, nmaps, autocmds.DocLoad). Also accepts the arguments `--default` or `--user` to view the default configuration, or your changes.
  *
  */
-//#content
+//#background
 export function viewconfig(...key: string[]) {
     // # and white space don't agree with FF's JSON viewer.
     // Probably other symbols too.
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -4206,11 +4204,13 @@ export function viewconfig(...key: string[]) {
...
- * View a JSON object in Firefox's JSON viewer. NB: only works well from a "normal" page, e.g not a Tridactyl help or new tab page.
+ * View a JSON object in Firefox's JSON viewer.
  */
-//#content
-export function jsonview(...json: string[]) {
-    window.location.href = "data:application/json," + json.join(" ").replace(/#/g, "%23").replace(/ /g, "%20")
+//#background
+export async function jsonview(...json: string[]) {
+    const tab = await tabopen("-w")
+    const url = "data:application/json," + json.join(" ").replace(/#/g, "%23").replace(/ /g, "%20")
```
Rummskartoffel, line 4212:
```suggestion
    const url = "data:application/json," + encodeURIComponent(json.join(" "))
```

src/excmds.ts
```diff
@@ -4206,11 +4207,13 @@ export function viewconfig(...key: string[]) {
...
 /**
- * View a JSON object in Firefox's JSON viewer. NB: only works well from a "normal" page, e.g not a Tridactyl help or new tab page.
+ * View a JSON object in Firefox's JSON viewer.
  */
-//#content
-export function jsonview(...json: string[]) {
-    window.location.href = "data:application/json," + json.join(" ").replace(/#/g, "%23").replace(/ /g, "%20")
+//#background
+export async function jsonview(...json: string[]) {
+    const tab = await tabopen("-w")
```
bovine3dom, line 4214: I just realised that this shouldn't work, right? `:viewconfig`'s whole problem was that it didn't work from the new tab page. We're now _always_ opening it from the new tab page.

I mean, I'm glad it does work, but I'm thinking that maybe `encodeURIComponent` was the real fix.

Rummskartoffel: Don't think it was. Try running `location = "data:application/json," + encodeURIComponent('{"foo": "bar"}')` in the new tab page's console. I'm still not getting the JSON viewer that way.

bovine3dom: so how on earth did we fix it? lol

Rummskartoffel: I'm guessing that the CSP bug that prevents the viewer from loading doesn't trigger through `tabs.executeScript()` for some reason. That's the only thing that's different, so it really can't be anything else. No idea *why* that is, though. Maybe we'll learn more once the bug I filed with Mozilla gets triaged.

Issue #822, opened 2018-07-17
Disable/Reenable toggle like uBlock and Privacy Badger?


utsekaj42:
Would it be possible to add a toolbar button to complete disable/reenable Tridactyl in a manner similar what is done with Privacy Badger or uBlockOrigin? Is the way these extensions work different than ignore mode? Both of these extensions remember which sites they are disabled for.

This would be helpful for trying to figure out if a site is rendering improperly due to Tridactyl or some other extension.

bovine3dom:
Related to #339.

bovine3dom:
Oh I guess #2004 wasn't actually an exact duplicate of this - this is browser wide rather than per URL

bovine3dom:
https://github.com/tridactyl/tridactyl/blob/df8a745f62b3c89b42998f496a2fd3240f42f664/src/content.ts#L36 per URL could be quite straightforward: wrap the whole thing in a config.getAsync("disable").then(...)

but that could break things that need to run very early in the lifetime of a page. Worth a go though

bovine3dom:
With #3497 this is now very easy - we just need to make a button and have it flip the superignore setting.

I guess we could make the button have an "enter ignore mode" thingy too.

avalonv:
Is this still not a thing? :(

I used back in 2020 but haven't in some time. What's the current best method to temporarily disable trydactyl?

bovine3dom:
@avalonv to disable it across the entire browser you can run `tri set superignore true` in the URL bar and then refresh the pages where Tridactyl was running to stop it. Run `tri unset superignore` in the address bar and refresh again to get it back.

I'll add a button one day, it wouldn't be hard to do :)

PR #3497, opened 2021-03-20
Super ignore mode

### LINKED ISSUE #2004: Feature/compatibility issue: Full ignore mode for URLs
Currently, when setting a site to be ignored, the Tridactyl iframe will still be loaded unto the page.
It would be great to have a way to blacklist/ignore a site completely as the current method can mess with frame detection.

To experience this issue:
1. `:autocmd DocStart vesalia.de mode ignore`
2. go to https://www.vesalia.de
3. the page will continuously reload

That site checks for `top.frames.length > 0` in order to break out of frames but having the Tridactyl frame there on all ocassions could potentially mess with other sites as well.

Tridactyl version: 1.17.1pre3329-ca15582

bovine3dom:
Fixes #2004. Related to #822

- Needs thorough testing to see if it breaks anything that needs to run quickly (e.g. DocStart autocmds)
- It looks like we can't prevent CSS from loading as we hardcode it in `manifest.json`. We could potentially switch to using `tabs.insertCSS` - https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/insertCSS - but that could be a pain. I suspect it isn't worth delaying this PR for it
- Need better documentation. `disable` is a bad name for a setting IMO

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -38,6 +38,8 @@ config.getAsync("superignore").then(async TRI_DISABLE => {
 // running from this entry point, which is to say, everything in the
 // content script, will use the excmds that we give to the module
 // here.
+
+if (TRI_DISABLE === "true") return
 const controller = await import("@src/lib/controller")
 const excmds_content = await import("@src/.excmds_content.generated")
 const hinting_content = await import("@src/content/hinting")
@@ -63,8 +65,6 @@ const scrolling = await import("@src/content/scrolling")
 const R = await import("ramda")
 const visual = await import("@src/lib/visual")
 const metadata = await import("@src/.metadata.generated")
-
-if (TRI_DISABLE === "true") return
 
 controller.setExCmds({
     "": excmds_content,
--- a/src/content.ts
+++ b/src/content.ts
@@ -33,6 +33,36 @@ import * as excmds_content from "@src/.excmds_content.generated"
 import { CmdlineCmds } from "@src/content/commandline_cmds"
 import { EditorCmds } from "@src/content/editor"
 import * as hinting_content from "@src/content/hinting"
+
+// Hook the keyboard up to the controller
+import * as ContentController from "@src/content/controller_content"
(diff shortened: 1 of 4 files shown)


src/content.ts
```diff
@@ -24,15 +24,48 @@ import {
...
+const excmds = await import("@src/.excmds_content.generated")
+const finding_content = await import("@src/content/finding")
+const itertools = await import("@src/lib/itertools")
+const messaging = await import("@src/lib/messaging")
+const State = await import("@src/state")
+const webext = await import("@src/lib/webext")
+const perf = await import("@src/perf")
+const keyseq = await import("@src/lib/keyseq")
+const native = await import("@src/lib/native")
+const styling = await import("@src/content/styling")
```
bovine3dom, line 59: I expected this not to have side-effects if it wasn't called, but that doesn't seem to be the case. (the root element of the page gets a TridactylThemeDefault class)

src/lib/config.ts
```diff
@@ -181,6 +181,15 @@ export class default_config {
...
     }
 
+    /*
+     * Disable Tridactyl almost completely within a page, e.g. `seturl ^https?://mail.google.com disable true`. Only takes affect on page reload.
+     *
+     * You are usually better off using `blacklistadd` and `seturl [url] noiframe true` as you can then still use some Tridactyl binds.
+     *
+     * NB: you should only use this with `seturl`. If you get trapped with Tridactyl disabled everywhere just run `tri reset disable` in the Firefox address bar.
+     */
+    disable: "true" | "false" = "false"
```
bovine3dom, line 191:
```suggestion
    superignore: "true" | "false" = "false"
```

src/content.ts
bovine3dom: 9a99c21b one of the sillier things I've done for a while

bovine3dom:
Seems to work OK now. If tests pass I'll merge it.

Issue #3047, opened 2020-12-01
hint whole page flag


bovine3dom:
There are some cases where it would be useful to hint an entire page (e.g. if the user has bound keys to scroll up/down in hint mode, or if the user has muscle memory for a specific hint they can't see).

We should add a flag that toggles this for all hint modes (default to off).

bovine3dom:
Added in #3534

Issue #2999, opened 2020-11-16
Add CSS selectors to more hint modes


bovine3dom:
There's very little reason, as far as I can tell, not to support CSS selectors in more hint modes using this line:

https://github.com/tridactyl/tridactyl/blob/9916d9f4334956e0fd533d13a6e83175a35f648a/src/excmds.ts#L4093

bovine3dom:
Added a while ago

Issue #2920, opened 2020-10-18
Do not pipe if hint is cancelled?


hejops:
*   Brief description of the problem:

I'd like to know if a pipe can be "terminated" if I decide to cancel a hint.

*   Steps to reproduce:

1. `composite hint -y | js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('chromium ' + url))` (or any appropriate executable)
2. Cancel the hint (Esc)
3. Chromium is still executed

*   Unix-like only: result of running `printf '%c\0\0\0{"cmd": "run", "command": "echo $PATH"}' 39 | ~/.local/share/tridactyl/native_main.py` in a terminal:

```ø{"cmd": "run", "content": "/home/joseph/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/sbin:/usr/lib/jvm/default/bin:/usr/bin/site_perl:/usr/bin/vendor_perl:/usr/bin/core_perl:/home/joseph/scripts:/home/joseph/.gem/ruby/2.7.0/bin\n", "code": 0}```

**Comment by bovine3dom** [comment]
No, we don't have any concept of success or failure, so once a pipe has started it can't be stopped. However, you can probably do what you want with `hint -W [ex command]` - e.g.

```
command openinchrome js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('chromium ' + url))
hint -W openinchrome
```

bovine3dom:
Another solution is to check that `url` isn't empty in your JS snippet

hejops:
`hint -W openinchrome` works quite well, thank you.

Can the -W flag be combined with -Jc? I tried something like `hint -Jc <elements> -W openinchrome` but the hint fails.

bovine3dom:
No, it can't. Hint mode is very messy really. You could add it or you could switch to callback hint mode - `:hint -JFc [selector with no spaces] elem => [some js]`. https://github.com/tridactyl/tridactyl/blob/eca1e1614b8d0e51dcb7369a3b4026fed72ad209/src/excmds.ts#L3998

[...]

Issue #1946, opened 2019-10-25
 Allow `hint -qy` to take arguments / turn off JS hints


bovine3dom:
I didn't really think before writing / merging #1944.

bovine3dom:
fixed in #3534

PR #3534, opened 2021-03-29
Refactor :hint

vtavernier:
## Summary
This commit refactors the `hint` ex command with the following:

### Internals
* Introduce an argument parser based on a state machine. We can now properly handle flags, combined short flags, and flags with optional arguments.
* Keeps backwards compatibility: all current flags are supported, but the ability to parse multiple options may introduce new behaviors (which would fail before silently).

### Added
* Add a `-!` flag: perform the default action on every hint
* Add a `-V` flag: consider elements out of the viewport for hinting

### Changed
* Changes `-f/-fr` to refine results from the default selectors
* Allow `-f/-fr` to read patterns containing spaces, if they are escaped with `\`

## Rationale
The hinting command already goes well beyond simply showing hints for clickable links. It makes sense that 'common' operations should have easier shortcuts than defaulting to JavaScript callbacks. This does make the code for `hint` slightly longer, but with the refactored control flow this doesn't seem significant.

## State
This is a first draft and comments are welcomed, as I am not familiar with the project. Most notably the following questions should be considered before merging:
* [x] Should users be able to refine selectors for non-default open modes? I could want to select images that have the .thumbnail class, so `:hint -S img.thumbnail` should work
* [x] `-;` option: I failed to understand the docs on what it does, maybe it should be made clearer? The implementation stayed the same but I might be wrong
* [ ] Some hintTabOpen calls had an async/await block, but others did not. Is this expected?
* [x] `-!` (new) immediate option: from early manual testing, this implementation seems to work, but I might be missing something?
* [ ] Error reporting: the argument parser can easily detect wrong argument combinations. How should they be reported to the user?
* [ ] Automated tests? What should be changed?

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -1135,33 +1135,6 @@ export function killables(includeInvisible = false) {
     ])
 }
 
-/** HintPage wrapper, accepts CSS selectors to build a list of elements
- * @hidden
- * */
-export function pipe(
-    selectors = DOM.HINTTAGS_selectors,
-    action: HintSelectedCallback = _ => _,
-    rapid = false,
-    jshints = true,
-): Promise<[Element, number]> {
-    return new Promise((resolve, reject) => {
-        hintPage(hintables(selectors, jshints), action, resolve, reject, rapid)
-    })
-}
-
-/** HintPage wrapper, accepts array of elements to hint
- * @hidden
- * */
-export function pipe_elements(
-    elements: Element[] | Hintables[] = DOM.elementsWithText(),
-    action: HintSelectedCallback = _ => _,
-    rapid = false,
-): Promise<[Element, number]> {
-    return new Promise((resolve, reject) => {
-        hintPage(toHintablesArray(elements), action, resolve, reject, rapid)
-    })
-}
-
 // Multiple dispatch? who needs it
 /** Returns an array of hintable objects from an array of elements
  * @hidden
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -1113,8 +1113,7 @@ export function hintByTextFilter(match: string | RegExp): HintSelectedCallback {
     return hint => {
         let text
         if (hint instanceof HTMLInputElement) {
-            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-            text = (hint as HTMLInputElement).value
+            text = hint.value
         } else {
             text = hint.textContent
         }
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -1001,16 +1001,21 @@ function pushSpace() {
 
     Elements are hintable if
         1. they can be meaningfully selected, clicked, etc
-        2. they're visible
+        2. they're visible (unless includeInvisible is true)
             1. Within viewport
             2. Not hidden by another element
 
     @hidden
 */
-export function hintables(selectors = DOM.HINTTAGS_selectors, withjs = false) {
+export function hintables(
+    selectors = DOM.HINTTAGS_selectors,
+    withjs = false,
+    includeInvisible = false,
+) {
+    const visibleFilter = DOM.isVisibleFilter(includeInvisible)
     const elems = R.pipe(
         DOM.getElemsBySelector,
-        R.filter(DOM.isVisible),
+        R.filter(visibleFilter),
         changeHintablesToLargestChild,
     )(selectors, [])
     const hintables: Hintables[] = [{ elements: elems }]
@@ -1020,7 +1025,7 @@ export function hintables(selectors = DOM.HINTTAGS_selectors, withjs = false) {
                 Array.from,
                 // Ramda gives an error here without the "any"
                 // Problem for a rainy day :)
-                R.filter(DOM.isVisible) as any,
+                R.filter(visibleFilter) as any,
                 R.without(elems),
                 changeHintablesToLargestChild,
             )(DOM.hintworthy_js_elems),
@@ -1075,15 +1080,19 @@ function isElementLargerThan(e1: Element, e2: Element): boolean {
 /** Returns elements that point to a saveable resource
  * @hidden
  */
-export function saveableElements() {
-    return DOM.getElemsBySelector(DOM.HINTTAGS_saveable, [DOM.isVisible])
+export function saveableElements(includeInvisible = false) {
+    return DOM.getElemsBySelector(DOM.HINTTAGS_saveable, [
+        DOM.isVisibleFilter(includeInvisible),
+    ])
 }
 
-/** Get array of images in the viewport
+/** Get array of images in the viewport, or all images if includeInvisible is true
  * @hidden
  */
-export function hintableImages() {
-    return DOM.getElemsBySelector(DOM.HINTTAGS_img_selectors, [DOM.isVisible])
+export function hintableImages(includeInvisible = false) {
+    return DOM.getElemsBySelector(DOM.HINTTAGS_img_selectors, [
+        DOM.isVisibleFilter(includeInvisible),
+    ])
 }
 
 /** Get array of selectable elements that display a text matching either plain
@@ -1093,29 +1102,36 @@ export function hintableImages() {
 export function hintByText(match: string | RegExp) {
     return DOM.getElemsBySelector(DOM.HINTTAGS_filter_by_text_selectors, [
         DOM.isVisible,
-        hint => {
-            let text
-            if (hint instanceof HTMLInputElement) {
-                // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
-                text = (hint as HTMLInputElement).value
-            } else {
-                text = hint.textContent
-            }
-            if (match instanceof RegExp) {
-                return text.match(match) !== null
-            } else {
-                return text.toUpperCase().includes(match.toUpperCase())
-            }
-        },
+        hintByTextFilter(match),
     ])
+}
+
+/** Return a predicate that checks whether an element matches a given text hinting filter
+ * @hidden
+ */
+export function hintByTextFilter(match: string | RegExp): HintSelectedCallback {
+    return hint => {
+        let text
+        if (hint instanceof HTMLInputElement) {
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            text = (hint as HTMLInputElement).value
+        } else {
+            text = hint.textContent
+        }
+        if (match instanceof RegExp) {
+            return text.match(match) !== null
+        } else {
+            return text.toUpperCase().includes(match.toUpperCase())
+        }
+    }
 }
 
 /** Array of items that can be killed with hint kill
 @hidden
  */
-export function killables() {
+export function killables(includeInvisible = false) {
     return DOM.getElemsBySelector(DOM.HINTTAGS_killable_selectors, [
-        DOM.isVisible,
+        DOM.isVisibleFilter(includeInvisible),
     ])
 }
 
@@ -1147,7 +1163,10 @@ export function pipe_elements(
 }
 
 // Multiple dispatch? who needs it
-function toHintablesArray(
+/** Returns an array of hintable objects from an array of elements
+ * @hidden
+ * */
+export function toHintablesArray(
     hintablesOrElements: Element[] | Hintables[],
 ): Hintables[] {
     if (!hintablesOrElements.length) return []
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -668,6 +668,10 @@ class Hint {
         this.hidden = false
     }
 
+    public static isHintable(target: Element): boolean {
+        return target.getClientRects().length > 0
+    }
+
     setName(n: string) {
         this.name = n
         this.flag.textContent = this.name
@@ -749,7 +753,7 @@ function buildHintsSimple(
     hintables: Hintables,
     onSelect: HintSelectedCallback,
 ) {
-    const els = hintables.elements
+    const els = hintables.elements.filter((el) => Hint.isHintable(el))
     const names = Array.from(
         hintnames(els.length + modeState.hints.length),
     ).slice(modeState.hints.length)
@@ -800,7 +804,7 @@ function buildHintsVimperator(
     hintables: Hintables,
     onSelect: HintSelectedCallback,
 ) {
-    const els = hintables.elements
+    const els = hintables.elements.filter((el) => Hint.isHintable(el))
     const names = Array.from(
         hintnames(els.length + modeState.hints.length),
     ).slice(modeState.hints.length)
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4361,6 +4361,8 @@ export async function hint(...args: string[]): Promise<any> {
         ExpectCallback,
         ExpectExcmd,
         ExpectSelector,
+        ExpectPipeSelector,
+        ExpectPipeAttribute,
     }
 
     // Open mode: how to act on the selected hintable element
@@ -4398,6 +4400,7 @@ export async function hint(...args: string[]): Promise<any> {
     let jshints = true
     let callback = null
     let excmd = null
+    let pipeAttribute = null
 
     // Parser state
     let state = State.Initial
@@ -4410,7 +4413,10 @@ export async function hint(...args: string[]): Promise<any> {
 
         switch (state) {
             case State.Initial:
-                if (arg.length >= 2 && arg[0] === "-" && arg[1] !== "-") {
+                if (arg == "-pipe") {
+                    // Special case for -pipe, which is not a |1,2]-letter argument
+                    state = State.ExpectPipeSelector
+                } else if (arg.length >= 2 && arg[0] === "-" && arg[1] !== "-") {
                     // Parse short arguments, i.e. - followed by (mostly) single-letter arguments,
                     // and some two-letter arguments.
 
@@ -4586,6 +4592,18 @@ export async function hint(...args: string[]): Promise<any> {
                 positionals.push(arg)
                 state = State.Initial
                 break
+            case State.ExpectPipeSelector:
+                // -pipe, first expect a selector
+                positionals.push(arg)
+                // Then, expect the attribute
+                state = State.ExpectPipeAttribute
+                break
+            case State.ExpectPipeAttribute:
+                // -pipe, second argument
+                pipeAttribute = arg
+                // Keep parsing options when we're done
+                state = State.Initial
+                break
         }
     }
 
@@ -4667,6 +4685,11 @@ export async function hint(...args: string[]): Promise<any> {
         const action = callback
             ? eval(callback)
             : (elem: any) => {
+                  if (pipeAttribute !== null) {
+                      // We have an attribute to pipe
+                      return elem[pipeAttribute]
+                  }
+
                   if (excmd) {
                       // We have an excmd to run. By spec, we append the element's href
                       if (elem.href) {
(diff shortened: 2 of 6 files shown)


src/excmds.ts
```diff
@@ -4345,35 +4347,231 @@ const KILL_STACK: Element[] = []
...
+        ImagesTab,
+        Kill,
+        KillTridactyl,
+        Scroll,
+        SaveResource,
+        SaveImage,
+        SaveAsResource,
+        SaveAsImage,
+        // TODO: What does this actually do? The doc is unclear
+        Semicolon,
```
bovine3dom, line 4378: `-;` calls `.focus()` on the selected element and explicitly tells Tridactyl that it is a child of the part of the page you want to scroll - https://github.com/tridactyl/tridactyl/blob/ced62d5fca6ea5aa986975cf5cf1765844498363/src/excmds.ts#L4603

It also has a little hack that lets you focus images.

The point of calling focus is mostly to let you then press the "right click button" on your keyboard (I think it's called context menu?) but on some sites you'll get some sort of hover pop-up.

src/excmds.ts
```diff
@@ -4302,8 +4300,12 @@ const KILL_STACK: Element[] = []
...
         - -J* disable javascript hints. Don't generate hints related to javascript events. This is particularly useful when used with the `-c` option when you want to generate only hints for the specified css selectors. Also useful on sites with plenty of useless javascript elements such as google.com
           - For example, use `bind ;jg hint -Jc .rc > .r > a` on google.com to generate hints only for clickable search results of a given query
-        - -br deprecated, use `-qb` instead
         - -F [callback] - run a custom callback on the selected hint, e.g. `hint -JF e => {tri.excmds.tabopen("-b",e.href); e.remove()}`.
+        - -V create hints for invisible elements. By default, elements outside the viewport when calling :hint are not hinted, this includes them anyways.
+        - -! perform action immediately. This effectively selects every hinted element in sequence
+          - For example, `hint -!bf Comments` opens in background tabs all visible links whose text matches `Comments`
+        - [selector] hint links that match the css selector
+          - `bind ;c hint [class*="expand"],[class="togg"]` works particularly well on reddit and HN
```
bovine3dom, line 4308: We should mention the deprecated -c bind because it's been around for so many years that people will bump into it in the wild and wonder what it did

src/excmds.ts
```diff
@@ -4345,35 +4347,231 @@ const KILL_STACK: Element[] = []
...
+                            case "#":
+                                newOpenMode = OpenMode.YankAnchor
+                                break
+                            case "y":
+                                newOpenMode = OpenMode.YankLink
+                                break
+                            case "p":
+                                newOpenMode = OpenMode.YankText
+                                break
+                            default:
```
bovine3dom, line 4508: `case "c":` will be so common that we would probably add dedicated (no-op) handling for it

src/excmds.ts
```diff
@@ -4389,289 +4587,216 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
...
-                hinting.hintableImages(),
-                async elem => {
-                    await hintTabOpen(new URL(elem.getAttribute("src"), window.location.href).href)
-                    return elem
-                },
-                rapid,
-            )
-            break
+            default:
+                hintables = hinting.hintables(positionals.length ? positionals.join(" ") : DOM.HINTTAGS_selectors, jshints, includeInvisible)
```
bovine3dom, line 4632: If selectors exist they should always override whatever the defaults are.

But how do we know that the positionals aren't e.g. callbacks?

src/excmds.ts
```diff
@@ -4345,35 +4347,231 @@ const KILL_STACK: Element[] = []
...
-    }
+                        if (newOpenMode !== undefined) {
+                            if (openMode !== OpenMode.Default) {
+                                // Notify that multiple open modes doesn't make sense
+                                logger.warning("multiple open mode flags specified, overriding the previous ones")
+                            }
 
-    // Hacky fix for #1374 - join rapid-hinted yanks
-    if (option === "-qy") {
-        run_exstr('composite hint -qpipe a href | js -p JS_ARG.join(" ") | yank')
```
bovine3dom, line 4362: I'm somewhat reluctant to lose this workaround since it was specifically requested. Can't we generalise it for all yank modes?

bovine3dom (commented):
Thanks so much for this PR!

I've just noted some things that jumped out at me while I skimmed it.

---

Do you think there's any possibility of you splitting this up into several smaller PRs? I really try to persuade people to keep PRs to <100 LOC so we have a good chance of reviewing them thoroughly and merging them before they go stale.

I'll have another look at this later today and try to answer some of your questions.

src/excmds.ts
vtavernier: Thinking back on it I actually didn't need to remove the -c flag, I just need to consider it's another argument with an option and push the corresponding value to `positionals`.

src/excmds.ts
vtavernier: See above

src/excmds.ts
vtavernier: Ok, so this was not the case with the current implementation, at least for the "non-default" modes, like ones that operate on killables, images, and saveables. It is relatively easy to add though.

From the argument parsing, `positionals` are the "left over" arguments which were not associated with a flag (e.g. `-f`, `-fr`, `-W`, or `-F`). Each flag that takes a value changes the current state to one that will process the remaining flags accordingly (specifically, `-F` and `-W` will consume all the remaining arguments and join them to form a callback, as the previous implementation was doing), which leaves only selectors in the `positionals` array.

So `bind a -F some js` will result in `positionals = ["a"]` and `callback = "some js"`, and trigger the JS callback with the `a` selector.

src/excmds.ts
vtavernier: I forgot I wanted to look at this issue before pushing. As a first step I'll replicate the behavior only for -y, but maybe this should be the case for all types of yanks?

vtavernier:
> Do you think there's any possibility of you splitting this up into several smaller PRs? I really try to persuade people to keep PRs to <100 LOC so we have a good chance of reviewing them thoroughly and merging them before they go stale.

I understand it's a big one and it makes review harder, but `hint` is a 326-lines function with argument parsing and actual command execution in one big switch/case that makes up most of these lines. Anything touching argument parsing will result in that switch changing and the important line count coming with it.

With my limited knowledge of the project's architecture I tried to split the behavior into the most important parts, i.e.
* Argument parsing
* Selecting elements
* Actions on hinted elements

Both element selection and actions are straight from the old code, so it shouldn't break much. I feel like trying to break this apart would require shoehorning the new argument parsing into the old code, which is definitely not designed to handle that.

Another added benefit is to have (IMO) saner argument parsing, as the current behavior doesn't allow combining options (which is where I started: I want to open links in the background whose text match some pattern, i.e. have `-f` and `-b` at the same time), although the docs do not mention that.

I could split away adding `-V` and `-!` to an extra PR but this'll barely shave off 10 lines, I don't think it is worth the extra work.

src/excmds.ts
vtavernier: I don't get what the "tells Tridactyl that it is a child of the part of the page you want to scroll". What part of the page are we talking about? Why does the element needs to be made a child of that part?

I guess if the goal is to make keyboard right-click doable... It's some kind of "FocusRightclick" action? Maybe it should be noted in the docs that it's the intended use, I didn't know about it, but I want to use it now that I do!

vtavernier:
Note to myself: the `-pipe` option seems to have been lost in the change, I'll add it by pushing again to this PR.

Summary of changes left to do:
* [ ] Add `-c` again
* [ ] Add `-pipe` again
* [ ] Rename Semicolon open mode

src/excmds.ts
bovine3dom: It's probably just easier to show you - e.g. https://doc.rust-lang.org/reference/types.html has two scrollable panes. `;;` lets you change which pane j,k etc. scroll. Just pick an element (any element) in each respective pane.

I don't think it _has_ to be a child of the element you want to scroll but it's usually easier to find a child than it is to find the exact element.

src/excmds.ts
bovine3dom: `hint a -F some js` instead of `hint -cF [selector] [callback]` will break a lot of user's code.

Including some of our own: https://github.com/tridactyl/tridactyl/blob/49c887222b72f7848957ba173a80037c49f29f57/src/lib/config.ts#L288

How would this handle `:hint a -F e => tri.excmds.tabopen("-b", e.href)`. Are we sure it wouldn't try to steal the `-b`?

---

I'm beginning to think that it's not worth trying to preserve backwards compatibility and instead call this function `hint2`. We can call the current one `hint_deprecated`, alias it to `hint` and use a config updater to try to change everyone's hint binds to `hint_deprecated`. In a few months we can do a similar dance and make `hint2` point to `hint`.

bovine3dom:
> I understand it's a big one and it makes review harder, but [...]

Yeah, I had a feeling this would be your answer : )

I'll skim it a few times to see if any obvious issues jump out so you can work on them soon. I'll try to carve out some time in a week or two to then look at it thoroughly (if we go for the `hint2` approach I can be much less thorough :) ).

I am likely to have no internet access from Thursday morning for around a week, perhaps a little longer.

src/excmds.ts
vtavernier: I based the initial implementation on what I could make out from the documentation and how "usual" argument parsing happens in CLIs, where in combined short-form flags, if one flag takes a value, that value has to be the next argument passed to the command. I was specifically looking for feedback on this by opening this PR, since I'm mostly using Tridactyl without bindings so I'm curious what users came up with to have some test cases.

I can change the parser to handle the `-cF` case (two short arguments taking values), it just requires an extra state and this will provide compatibility. I think we're pretty close to full backwards compatibility (especially after implementing this change), but I'll leave the choice of whether or not to feature-gate it up to you!

For the example you describe, `:hint a -F e => tri.excmds.tabopen("-b", e.href)` will go through the following parser steps (parser state in brackets):
* `[Initial] ['a', ...]`: lone argument, add to positionals (thus `a` is a selector part)
* `[Initial] [..., '-F', ...]`: single argument `-F`, transition to `ExpectCallback`
* `[ExpectCallback] [..., 'e', '=>', 'tri.excmds.tabopen("-b",', 'e.href)']`: in ExpectCallback, consume all the remaining arguments into a callback unconditionally and exit the parsing loop

Thus, `positionals = ['a']` and `callback = 'e => tri.excmds.tabopen("-b", e.href)'`.

vtavernier:
No problem for the delay reviewing this PR, I don't intend on switching away from Tridactyl anytime soon so I don't mind this staying open longer. As long as no major rewrite is underway that'd need rebasing everything of course :wink:

src/excmds.ts
vtavernier: Oh I see, by setting the focus to a particular (visible) element this indeeds switches the target for scroll events to its first parent with overflow enabled; aside from enabling keyboard right-clicking on it. `ScrollFocus` sounds nice enough (at least for the hint command internals!)

src/excmds.ts
vtavernier: Thanks for the link, I added some interesting examples I could find to the automated tests of the argument parser I added in https://github.com/tridactyl/tridactyl/pull/3534/commits/806760ca1c8f06b2cb1255c176dc0763a10c6631

src/excmds.ts
vtavernier: Fixed by https://github.com/tridactyl/tridactyl/pull/3534/commits/be5a66e046530f078c5bfe38bfa3ab09c7efe475. All yanks will put their results in the clipboard as a newline-delimited string. This works for rapid (-q) and immediate mode (-!). The key was that the hintPage function already resolves to the "onSelect" callback value, so it was just a matter of processing the result with `.then`

vtavernier:
Listing related issues I could find:
* `-V` would fix "hint whole page flag #3047"
* New behavior for parsing selectors/-c would fix:
  * Issue "Add CSS selectors to more hint modes #2999"
  * Proposed solution in "Do not pipe if hint is cancelled? #2920"
  * Issue "Allow hint -qy to take arguments / turn off JS hints #1946"
  * Issue "Make hint navigation with filter by "elem" like hint -; #1893"
* `-!` would fix "Ultra-rapid hint mode #935"
  
Related but not fixed directly:
* Replicate 'hint -p' with 'hint -pipe'? #1807

src/excmds.ts
```diff
@@ -4304,6 +4306,9 @@ const KILL_STACK: Element[] = []
           - For example, use `bind ;jg hint -Jc .rc > .r > a` on google.com to generate hints only for clickable search results of a given query
         - -br deprecated, use `-qb` instead
         - -F [callback] - run a custom callback on the selected hint, e.g. `hint -JF e => {tri.excmds.tabopen("-b",e.href); e.remove()}`.
+        - -V create hints for invisible elements. By default, elements outside the viewport when calling :hint are not hinted, this includes them anyways.
+        - -! perform action immediately. This effectively selects every hinted element in sequence
+          - For example, `hint -!bf Comments` opens in background tabs all visible links whose text matches `Comments`
 
     Excepting the custom selector mode and background hint mode, each of these hint modes is available by default as `;<option character>`, so e.g. `;y` to yank a link's target; `;g<option character>` starts rapid hint mode for all modes where it makes sense, and some others.
```
bovine3dom, line 4313:
```suggestion
    Excepting the custom selector mode, background hint mode and the "immediate" modifier, each of these hint modes is available by default as `;<option character>`, so e.g. `;y` to yank a link's target; `;g<option character>` starts rapid hint mode for all modes where it makes sense, and some others.
```

We also need to add a `;V` bind following this convention (just plonk it in src/lib/config.ts alongside the others - should be obvious).

I'd be tempted to add `;![x]` binds too for most modes where it made sense.

src/excmds.ts
```diff
@@ -4304,6 +4306,9 @@ const KILL_STACK: Element[] = []
           - For example, use `bind ;jg hint -Jc .rc > .r > a` on google.com to generate hints only for clickable search results of a given query
         - -br deprecated, use `-qb` instead
         - -F [callback] - run a custom callback on the selected hint, e.g. `hint -JF e => {tri.excmds.tabopen("-b",e.href); e.remove()}`.
+        - -V create hints for invisible elements. By default, elements outside the viewport when calling :hint are not hinted, this includes them anyways.
+        - -! perform action immediately. This effectively selects every hinted element in sequence
```
bovine3dom, line 4310:
```suggestion
        - -! execute all hints without waiting for a selection
```

I didn't understand what `!` did until I ran it. I think this is a bit better but not perfect

src/excmds.ts
```diff
@@ -4304,6 +4306,9 @@ const KILL_STACK: Element[] = []
           - For example, use `bind ;jg hint -Jc .rc > .r > a` on google.com to generate hints only for clickable search results of a given query
         - -br deprecated, use `-qb` instead
         - -F [callback] - run a custom callback on the selected hint, e.g. `hint -JF e => {tri.excmds.tabopen("-b",e.href); e.remove()}`.
+        - -V create hints for invisible elements. By default, elements outside the viewport when calling :hint are not hinted, this includes them anyways.
```
bovine3dom, line 4309: Bug:

1. visit http://txti.es/
2. `:hint -V`
3. Receive `# controller in excmd: TypeError: rect is undefined`. Hints are not numbered

src/excmds.ts
```diff
@@ -4270,7 +4271,7 @@ const KILL_STACK: Element[] = []
 
 /** Hint a page.
 
-    @param option
+    @param args
```
bovine3dom, line 4274: We should mention that most flags can be combined now, somewhere : )

src/content/hinting.ts
```diff
@@ -1093,61 +1102,44 @@ export function hintableImages() {
...
 
+/** Return a predicate that checks whether an element matches a given text hinting filter
+ * @hidden
+ */
+export function hintByTextFilter(match: string | RegExp): HintSelectedCallback {
+    return hint => {
+        let text
+        if (hint instanceof HTMLInputElement) {
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-assertion
+            text = (hint as HTMLInputElement).value
```
Rummskartoffel, line 1117:
```suggestion
            text = hint.value
```

vtavernier: Good catch, I didn't really look at this and just moved the filter's code into its own function!

src/excmds.ts
vtavernier: Your suggestion looks good to me, at least for now I don't have any other idea.

src/excmds.ts
vtavernier: Added `;V` in f4560d8f. For the `-!` bindings I think we have to consider the risks, for example `;!b` would open all visible links in background pages, which might add up to quite a lot (and melt the user's machine).

Maybe in a follow-up PR we could have `!` keyboard shortcut (in hint mode) that does the same as `-!` does upfront, but giving the user a chance to review the hinted targets?

src/excmds.ts
vtavernier: I reorganized the flags in the docs to mention this, as well as which flags are compatible/conflict with each other in 0a664c5d5cc10ae3e3b72f6f5726d1526bebf3c3

src/excmds.ts
vtavernier: It seems allowing "invisible" elements to be hinted breaks some invariants for the Hint class. I can get -V to "somewhat" work by checking if an element has client rects (efd05bf85bbdc4079664d7fc09416bbb1837d409), but the flags are offset for some reason.
image

src/excmds.ts
```diff
@@ -4389,289 +4383,182 @@ export async function hint(option?: string, selectors?: string, ...rest: string[
...
-                    return elem
-                },
-                rapid,
-            )
-            break
+                  if (config.excmd) {
+                      // We have an excmd to run. By spec, we append the element's href
+                      if (elem.href) {
+                          // /!\ RACY RACY RACY!
+                          run_exstr(config.excmd + " " + elem.href)
```
Rummskartoffel, line 4404: I think it should be possible to do this a bit more safely:
```suggestion
                          CTSELF["config.excmd"](elem.href)
```

vtavernier: I don't know about this, I just took the code for the various callbacks as-is and added it in the corresponding spot while refactoring. With the scope of this PR being already large enough I don't know changing even more things in the same batch?

Rummskartoffel: Yeah, you're right, it's kind of out of scope.

src/lib/hint_util.ts
```diff
@@ -0,0 +1,424 @@
...
+ */
+
+import Logger from "@src/lib/logging"
+import * as DOM from "@src/lib/dom"
+import * as hinting from "@src/content/hinting"
+
+/**
+ * Open mode: how to act on the selected hintable element
+ */
+export enum OpenMode {
```
bovine3dom, line 12: I've rebased this on master, where we added a new yarn command `yarn run rebuild` that uses `esbuild` and is much quicker (1 second versus 40 seconds).

It looks like `esbuild` can't deal with these `enum`s though - I get the error

`# controller in excmd: ReferenceError: can't access lexical declaration 'OpenMode' before initialization` whenever I try to run hint mode.

Do you think you might have time to look into it? We don't use `enum` anywhere else in the code base so if it looks like it's insurmountable could we rewrite the code to not use `enum`s?

We'll probably eventually switch to using `esbuild` for "proper" builds too as it really is much quicker.

Thanks : )

vtavernier: 8a059af fixes this. It seems esbuild miscompiles the `const` declaration I used to alias `hint_util.OpenMode` in `hint`, but enums are properly supported by esbuild!

PR #3538, opened 2021-03-31
mpvsafe: Allow mpv to keep running after Firefox quits

gnull:
Detach the mpv process after starting it. This prevents it from terminating with SIGPIPE after Firefox quits.

I think this should be the sensible behaviour. There's no use it terminating all the mpv instances after user closes Firefox.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -607,7 +607,7 @@ export class default_config {
         "mkt!": "mktridactylrc -f",
         "mktridactylrc!": "mktridactylrc -f",
         mpvsafe:
-            "js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('mpv ' + url))",
+            "js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('mpv --no-terminal ' + url))",
         exto: "extoptions",
         extpreferences: "extoptions",
         extp: "extpreferences",


src/lib/config.ts
```diff
@@ -607,7 +607,7 @@ export class default_config {
         "mkt!": "mktridactylrc -f",
         "mktridactylrc!": "mktridactylrc -f",
         mpvsafe:
-            "js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('mpv ' + url))",
+            "js -p tri.excmds.shellescape(JS_ARG).then(url => tri.excmds.exclaim_quiet('mpv ' + url + ' > /dev/null 2>&1 &'))",
```
bovine3dom, line 610: This presumably will work on OSX and Linux but not Windows.

Could you therefore add a Windows-specific config entry with the previous value? Circa line 1143.

gnull: Oh… I see that test fail. I will figure out what's wrong and push new changes.

src/lib/config.ts
```diff
@@ -1157,6 +1157,11 @@ const platform_defaults = {
 (New-Object System.Net.WebClient).DownloadFile('https://raw.githubusercontent.com/tridactyl/native_messenger/master/installers/windows.ps1', '%TEMP%/tridactyl_installnative.ps1');\
 & '%TEMP%/tridactyl_installnative.ps1' -Tag %TAG;\
 Remove-Item '%TEMP%/tridactyl_installnative.ps1'"`,
+
+        exaliases = {
```
bovine3dom, line 1161:
```suggestion
        exaliases: {
```

src/lib/config.ts
bovine3dom: I think it's just that colon Vs equals thing.

If you could add a comment above line 609 saying that Windows has an override it'll probably prevent some confusion in future developers. Thanks!

gnull:
I found a better way to solve the problem, now we don't need an exception for Windows.

I added the `--no-terminal` flag to mpv invocation. This way, `mpv` just doesn't use the stdout/err and can keep running after it is closed. I figured out that forking mpv with `&` was unnecessary, just fixing the stdout is enough to prevent crashes.

bovine3dom:
LGTM! I'll merge it once tests pass. Thanks!

Issue #3498, opened 2021-03-20
Consider switching to esbuild


bovine3dom:
It claims to be x times faster than webpack. Tridactyl being slow to build is one of my pet hates so this could improve my life if we can get it working.

https://github.com/evanw/esbuild

Rummskartoffel:
One related pet hate I have is that `yarn install` is slow – which indirectly also makes switching branches and pulling/merging slow –, which I believe is due in part or even mostly to webpack's dependencies. Maybe using esbuild would improve that as well.

bovine3dom:
If you want to better appreciate how quick yarn is, symlink it to npm and see what we used to put up with : D

bovine3dom:
Notes on moving to `esbuild` entirely:

```
yarn run esbuild src/content.ts --bundle --target=firefox68 --outfile=build/content.js
yarn run esbuild src/background.ts --bundle --target=firefox68 --outfile=build/background.js
```

is 90% of the way there but some packages we use need shims - https://github.com/tridactyl/tridactyl/blob/f5359f3370806fc503bdde70879be70a9ceb0c92/webpack.config.js#L33

Need to work out how to do those with `esbuild` or stop using the offending packages (`rss-parser` and `css`).

```
 > node_modules/css/lib/stringify/source-map-support.js:10:19: error: Could not resolve "path" (set platform to "node" when building for node)
    10 │ var path = require('path');
```

It looks like https://esbuild.github.io/api/#inject is the way to do it.

bovine3dom:
^ most of this is now resolved in #3498 (package.json "browser" field takes care of it).

I'll leave this issue open as it would be nice to drop `webpack` for the full-fat builds too.

bovine3dom:
> 
> 
> Struggling to work out how to replicate the old compilation / type checking - `tsc --project tsconfig.json --noEmit` gives me errors (on `master`) too.
> 
> Stuff like
> 
> ```
> src/completions/Sessions.ts:1:27 - error TS2691: An import path cannot end with a '.ts' extension. Consider importing '@src/lib/webext.js' instead.
> ```
> 
> which is clearly a lie as most of Tridactyl would be broken otherwise.

That error is correct, Webpack just doesn't follow "vanilla" TypeScript in that regard. Import paths ending in ".ts" are a long-standing WONTFIX for TS, as far as I know.

_Originally posted by @Rummskartoffel in https://github.com/tridactyl/tridactyl/pull/3554#issuecomment-826361192_

[...]

PR #3554, opened 2021-04-07
Switch to esbuild-loader

bovine3dom:
Progress towards #3498 (ideally we'd drop webpack entirely, but this is a good first step).

Reduces build times from ~40 seconds to ~25 seconds on my machine.

This skips type checking so we need to make sure that we still do that in CI and provide an easy way for people to run it when they need to. I imagine most editors will have type checking built-in via the language server protocol, so it shouldn't be a big deal, right?

Todo:
- [x] add `yarn run typecheck`
- [x] use `yarn run typecheck` during CI
- [x] investigate minification

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -49,13 +49,13 @@ else
   scripts/excmds_macros.py
 fi
 
-# .bracketexpr.generated.ts is needed for metadata generation
-"$(yarn bin)/nearleyc" src/grammars/bracketexpr.ne > \
-  src/grammars/.bracketexpr.generated.ts
-
 # You can use `--quick` to test out small changes without updating docs / metadata etc.
 # If you get weird behaviour just run a full build
 if [ "$QUICK_BUILD" != "1" ]; then
+
+    # .bracketexpr.generated.ts is needed for metadata generation
+    "$(yarn bin)/nearleyc" src/grammars/bracketexpr.ne > \
+      src/grammars/.bracketexpr.generated.ts
 
     # It's important to generate the metadata before the documentation because
     # missing imports might break documentation generation on clean builds
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -2,6 +2,20 @@
 
 set -e
 
+for arg in "$@"
+do
+    case $arg in
+        --quick)
+            QUICK_BUILD=1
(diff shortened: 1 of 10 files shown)


bovine3dom:
Struggling to work out how to replicate the old compilation / type checking - `tsc --project tsconfig.json --noEmit` gives me errors (on `master`) too.

Stuff like

```
src/completions/Sessions.ts:1:27 - error TS2691: An import path cannot end with a '.ts' extension. Consider importing '@src/lib/webext.js' instead.
```

which is clearly a lie as most of Tridactyl would be broken otherwise.

scripts/build.sh
```diff
@@ -2,6 +2,20 @@
 
 set -e
 
+for arg in "$@"
+do
+    case $arg in
+        --quick)
+            QUICK_BUILD=1
```
bovine3dom, line 9: Should make this emit a disclaimer saying that if it doesn't work run a proper build

bovine3dom:
I've been lazy and made normal builds use `tsc` again. I think we'll probably use `rebuild` most of the time (and let CI do type checking) so it's probably not a very big deal, right?

Otherwise we'd have to look at the `isolatedModules` option which we need for the metadata pass to make sense with `esbuild` (AFAICT).

So with this PR:

- :heavy_check_mark: rebuilds go from 40s -> 7s
- :red_circle: full builds unchanged (if we switch to esbuild we could get it down to 25s).

bovine3dom:
I binned `esbuild-loader`. `yarn run rebuild` now takes 1 second :rocket:

Rummskartoffel:
> 
> 
> Struggling to work out how to replicate the old compilation / type checking - `tsc --project tsconfig.json --noEmit` gives me errors (on `master`) too.
> 
> Stuff like
> 
> ```
> src/completions/Sessions.ts:1:27 - error TS2691: An import path cannot end with a '.ts' extension. Consider importing '@src/lib/webext.js' instead.
> ```
> 
> which is clearly a lie as most of Tridactyl would be broken otherwise.

That error is correct, Webpack just doesn't follow "vanilla" TypeScript in that regard. Import paths ending in ".ts" are a long-standing WONTFIX for TS, as far as I know.

PR #3561, opened 2021-04-10
Add "rebuild" package.json script that skips docs, new tab, and tutor

Rummskartoffel:
On my setup, "rebuild" takes ca. 23s, while "build" is around 45s.


Rummskartoffel:
Duplicate of #3554

PR #3564, opened 2021-04-10
Make Tridactyl reload pages on dev rebuilds

bovine3dom:
At the moment this is more confusing than anything else - Firefox reloads itself too early, before all the scripts have updated, so you end up with a page reload but running old code.

Have to go back to the CLI and press `R` to force a reload.

I'm thinking that we probably need to wait until all the files have compiled before we copy them to `build/`.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -5235,8 +5235,8 @@ browser.runtime.onInstalled.addListener(details => {
             updatenative(false)
         } else {
             // Temporary extension has been updated in place
-            // Reload all pages so Tridactyl can work in them
-            reloadall()
+            // Open a new tab where Tridactyl will work for convenience
+            tabopen()
         }
     }
 })
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -5230,8 +5230,15 @@ export async function keyfeed(mapstr: string) {
 //#background_helper
 browser.runtime.onInstalled.addListener(details => {
     if (details.reason === "install") tutor("newtab")
-    else if ((details as any).temporary !== true && details.reason === "update") updatenative(false)
-    // could add elif "update" and show a changelog. Hide it behind a setting to make it less annoying?
+    else if (details.reason === "update") {
+        if ((details as any).temporary !== true) {
+            updatenative(false)
+        } else {
+            // Temporary extension has been updated in place
+            // Reload all pages so Tridactyl can work in them
+            reloadall()
+        }
+    }
 })
 
 /** Opens optionsUrl for the selected extension in a popup window.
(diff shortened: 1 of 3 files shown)


Rummskartoffel:
That's strange. For me `web-ext` doesn't reload tabs at all, and never has – only the extension itself, which I prefer, because that way I can still inspect old console output and such.

src/excmds.ts
```diff
@@ -5230,8 +5230,15 @@ export async function keyfeed(mapstr: string) {
...
     if (details.reason === "install") tutor("newtab")
-    else if ((details as any).temporary !== true && details.reason === "update") updatenative(false)
-    // could add elif "update" and show a changelog. Hide it behind a setting to make it less annoying?
+    else if (details.reason === "update") {
+        if ((details as any).temporary !== true) {
+            updatenative(false)
+        } else {
+            // Temporary extension has been updated in place
+            // Reload all pages so Tridactyl can work in them
+            reloadall()
```
bovine3dom, line 5239: @Rummskartoffel

The idea of this PR is to make the pages reload (which I agree may be of debatable value - it might be better to run `tutor()` instead).

The problem is that `web-ext` reloads _the extension_ too soon and ends up with stale extension code. Presumably that is independent of this PR.

Rummskartoffel: I hadn't noticed that issue before, either. Do full builds have this problem as well?

bovine3dom: No, they don't. (I said this to you on Matrix but I should probably put it here for posterity)

src/excmds.ts
```diff
@@ -5230,8 +5230,15 @@ export async function keyfeed(mapstr: string) {
...
     if (details.reason === "install") tutor("newtab")
-    else if ((details as any).temporary !== true && details.reason === "update") updatenative(false)
-    // could add elif "update" and show a changelog. Hide it behind a setting to make it less annoying?
+    else if (details.reason === "update") {
+        if ((details as any).temporary !== true) {
+            updatenative(false)
+        } else {
+            // Temporary extension has been updated in place
+            // Open a new tab where Tridactyl will work for convenience
+            tabopen()
```
bovine3dom, line 5239: @Rummskartoffel new behaviour is to open a new tab once the extension has reloaded so that you have somewhere where Tridactyl works. Do you think that's still annoying? Do you have any better ideas?

(I have fixed the stale-extension code issue. I think)

Rummskartoffel: New tab's fine with me, I just wanted to make sure I don't lose stuff.

PR #3566, opened 2021-04-11
Various unseen changes to fuzzy completions

bovine3dom:
More changes to fuzzy completions by meep. Haven't looked at them yet


src/completions.ts
```diff
@@ -160,16 +160,18 @@ export abstract class CompletionSourceFuse extends CompletionSource {
...
+        // minMatchCharLength: 3,
     }
 
     // PERF: Could be expensive not to cache Fuse()
     // yeah, it was.
     fuse = undefined
 
     protected lastExstr: string
-    protected sortScoredOptions = false
+    protected sortScoredOptions = true
```
bovine3dom, line 174: I can't think of a good reason why this was ever `false`, so there could well be one

src/completions.ts
```diff
@@ -213,7 +215,10 @@ export abstract class CompletionSourceFuse extends CompletionSource {
 
         // Filter by query if query is not empty
         if (query) {
-            this.setStateFromScore(this.scoredOptions(query))
+            this.setStateFromScore(
+                this.scoredOptions(query, prefix === "taball " ? true : false),
+            )
```
bovine3dom, line 220: Surely this should be in `TabAll.ts`?

Need to look to see what it is actually doing. Why does `taball` need a special case that `tab` doesn't?

src/completions.ts
```diff
@@ -213,7 +215,10 @@ export abstract class CompletionSourceFuse extends CompletionSource {
...
+            // this.setStateFromScore(this.scoredOptions(query))
```
bovine3dom, line 221:
```suggestion
```

src/completions.ts
```diff
@@ -246,14 +251,15 @@ export abstract class CompletionSourceFuse extends CompletionSource {
     }
 
     /** Rtn sorted array of {option, score} */
-    scoredOptions(query: string): ScoredOption[] {
+    scoredOptions(query: string, extended = false): ScoredOption[] {
         const searchThis = this.options.map((elem, index) => ({
             index,
             fuseKeys: elem.fuseKeys,
         }))
+        this.fuseOptions.useExtendedSearch = extended
```
bovine3dom, line 259: Should probably document what this means

src/completions/Tab.ts
```diff
@@ -43,9 +43,7 @@ class BufferCompletionOption
         const favIconUrl = tab.favIconUrl
             ? tab.favIconUrl
             : Completions.DEFAULT_FAVICON
-        const indicator = tab.audible
-            ? String.fromCodePoint(0x1F50A)
-            : ""
+        // const indicator = tab.audible ? String.fromCodePoint(0x1f50a) : ""
```
bovine3dom, line 46: I don't see any good reason why this was removed, or why the code point became lowercase (if that even matters?)

src/completions/Tab.ts
```diff
@@ -76,8 +74,9 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
             "Tabs",
         )
         this.sortScoredOptions = true
-        this.shouldSetStateFromScore =
-            config.get("completions", "Tab", "autoselect") === "true"
+        this.shouldSetStateFromScore = false
+        // autoselect after space sucks for (fuzzy-) searching with multiple words
+        // config.get("completions", "Tab", "autoselect") === "true"
```
bovine3dom, line 79: As I said on Matrix we shouldn't do this as it will break muscle memory and annoy lots of users. The purpose of this change is to allow users to insert spaces into search queries.

I see three options:

- do nothing, or
- change the `<Space>` bind to `<S-Space>` and document it prominently on the new tab page
- make the `<Space>` bind configurable so it only works on completions where it makes sense (e.g. `:open` and not `:undo` nor `:tab`).

src/completions/TabAll.ts
```diff
@@ -3,7 +3,7 @@ import { browserBg } from "@src/lib/webext"
 import * as Containers from "@src/lib/containers"
 import * as Completions from "@src/completions"
 import * as Messaging from "@src/lib/messaging"
-import * as config from "@src/lib/config"
+// import * as config from "@src/lib/config"
```
bovine3dom, line 6: another spurious change?

src/completions/TabAll.ts
```diff
@@ -45,15 +45,15 @@ class TabAllCompletionOption
...
 
     constructor(private _parent) {
         super(["taball", "tabgrab"], "TabAllCompletionSource", "All Tabs")
 
         this.updateOptions()
         this._parent.appendChild(this.node)
-        this.shouldSetStateFromScore =
-            config.get("completions", "TabAll", "autoselect") === "true"
+        this.shouldSetStateFromScore = false
+        // config.get("completions", "TabAll", "autoselect") === "true"
```
bovine3dom, line 56: ditto as above on `<Space>` bind changes

src/excmds.ts
```diff
@@ -3458,7 +3458,30 @@ export async function yankimage(url: string): Promise<void> {
...
+        | Token     |  	Match type 	                  |    Description<br>|
+        |:-----------|:------------------------------:|--------------------------------------:|
+        | jscript 	|   fuzzy-match 	              |   Items that fuzzy match jscript|
+        | =scheme 	|   exact-match 	              |   Items that are scheme|
+        | 'python 	|   include-match 	              |   Items that include python|
+        | !ruby 	    |   inverse-exact-match 	      |   Items that do not include ruby|
+        | ^java 	    |   prefix-exact-match 	          |   Items that start with java|
+        | !^earlang 	|   inverse-prefix-exact-match 	  |   Items that do not start with earlang|
+        | .js$ 	    |   suffix-exact-match 	          |   Items that end with .js|
+        | !.go$ 	    |   inverse-suffix-exact-match 	  |   Items that do not end with .go|
```
bovine3dom, line 3478: this mixture of tabs and spaces hurts my brain

src/excmds.ts
```diff
@@ -3458,7 +3458,30 @@ export async function yankimage(url: string): Promise<void> {
...
+
+        The Fuzzyness Factor cant be set via the "fuzzy_threshold" (value between 0 and 1).<br>
+        0.1 "almost" only hits exact matches, <br>
+        0.3 "some" minor typos, missing chars also hit (default)<br>
+        Higher values if u have no clue what you are searching for...<br>
```
bovine3dom, line 3483: Need to reword this for grammar + to mention `completionfuzziness` name rather than `fuzzy_threshold`

src/lib/config.ts
```diff
@@ -271,7 +271,7 @@ export class default_config {
         // N: "findnext -1",
         // ",<Space>": "nohlsearch",
         M: "gobble 1 quickmark",
-        B: "fillcmdline taball",
+        B: "fillcmdline tab",
```
bovine3dom, line 274: I think this is my fault - I was testing something else. Remove

src/lib/config.ts
```diff
@@ -1133,11 +1133,10 @@ export class default_config {
...
 
     /**
-     * Threshold for fuzzy matching on completions. Lower => stricter matching. Range between 0 and 1: 0 corresponds to perfect matches only. 1 will match anything.
-     *
-     * https://fusejs.io/api/options.html#threshold
+     * Threshold for fuzzy matching.
+     * lower == more strict
      */
-    completionfuzziness = 0.3
+    fuzzy_threshold = 0.3
```
bovine3dom, line 1139: Revert all of this

src/completions.ts
```diff
@@ -160,16 +160,18 @@ export abstract class CompletionSourceFuse extends CompletionSource {
         includeScore: true,
         findAllMatches: true,
         ignoreLocation: true,
-        threshold: config.get("completionfuzziness"),
```
bovine3dom, line 163: Don't want to rename this setting

src/completions.ts
Rummskartoffel: Maybe some completions don't want to sort by best match for some reason?

bovine3dom: but why? :thinking:

Rummskartoffel: No idea

PR #3582, opened 2021-04-16
Hinting: improve hintables() performance by not using Ramda

Rummskartoffel:
I got annoyed at hinting being slow on YouTube, so I decided to look into it and found that Ramda's functions significantly slowed down `hintables()`. Doing things "by hand" makes it run a lot faster on complex pages – [this profile comparison](https://share.firefox.dev/3v1G0jD) I made reports about a 25% performance improvement. It definitely also *feels* faster to me, though of course that might just be because I'm biased in favour of my own code ;-).

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -1017,22 +1017,17 @@ export function hintables(
     includeInvisible = false,
 ) {
     const visibleFilter = DOM.isVisibleFilter(includeInvisible)
-    const elems = R.pipe(
-        DOM.getElemsBySelector,
-        R.filter(visibleFilter),
-        changeHintablesToLargestChild,
-    )(selectors, [])
+    const elems = changeHintablesToLargestChild(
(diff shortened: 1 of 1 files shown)


src/content/hinting.ts
```diff
@@ -1017,23 +1017,17 @@ export function hintables(
...
         hintables.push({
-            elements: R.pipe(
-                Array.from,
-                // Ramda gives an error here without the "any"
-                // Problem for a rainy day :)
-                R.filter(visibleFilter) as any,
-                R.without(elems),
-                changeHintablesToLargestChild,
-            )(DOM.hintworthy_js_elems),
-            hintclasses: ["TridactylJSHint"],
```
bovine3dom, line 1036: Should this line really have been removed?

Rummskartoffel: Whoops

PR #3589, opened 2021-04-21
Avoid updating DOM in-place

bovine3dom:
As per https://github.com/tridactyl/tridactyl/issues/1092#issuecomment-823734144

I haven't tested to see if performance is any better or worse.

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -296,28 +296,17 @@ export abstract class CompletionSourceFuse extends CompletionSource {
     }
 
     /** Call to replace the current display */
-    // TODO: optionContainer.replaceWith and optionContainer.remove don't work.
-    // I don't know why, but it means we can't replace the div in one go. Maybe
-    // an iframe thing.
     updateDisplay() {
-        /* const newContainer = html`<div>` */
-
-        while (this.optionContainer.hasChildNodes()) {
-            this.optionContainer.removeChild(this.optionContainer.lastChild)
-        }
+        const newContainer = document.createElement("div")
 
         for (const option of this.options) {
-            /* newContainer.appendChild(option.html) */
             if (option.state !== "hidden")
-                this.optionContainer.appendChild(option.html)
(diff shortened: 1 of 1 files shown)


src/completions.ts
```diff
@@ -296,28 +296,17 @@ export abstract class CompletionSourceFuse extends CompletionSource {
...
-    // TODO: optionContainer.replaceWith and optionContainer.remove don't work.
-    // I don't know why, but it means we can't replace the div in one go. Maybe
-    // an iframe thing.
     updateDisplay() {
-        /* const newContainer = html`<div>` */
-
-        while (this.optionContainer.hasChildNodes()) {
-            this.optionContainer.removeChild(this.optionContainer.lastChild)
-        }
+        const newContainer = this.optionContainer.cloneNode(false)
```
Rummskartoffel, line 300: `cloneNode()` always returns `Node`, so this needs a type assertion, otherwise we get an error on line 308:
```suggestion
        const newContainer = this.optionContainer.cloneNode(false) as HTMLElement
```

src/completions.ts
```diff
@@ -296,28 +296,17 @@ export abstract class CompletionSourceFuse extends CompletionSource {
...
 
         for (const option of this.options) {
-            /* newContainer.appendChild(option.html) */
             if (option.state !== "hidden")
-                this.optionContainer.appendChild(option.html)
+                // This is probably slow: `.html` means the HTML parser will be invoked
+                newContainer.appendChild(option.html)
```
Rummskartoffel, line 305: The slow part is the creation of the element in question, which happens in the constructor of `CompletionOptionHTML`'s descendants. This just accesses it.

bovine3dom: Yeah, you're right - https://github.com/tridactyl/tridactyl/blob/ced62d5fca6ea5aa986975cf5cf1765844498363/src/completions.ts#L102

I thought `.html` was a string that had to be parsed. Wasn't paying attention

PR #3604, opened 2021-04-23
Add drawable no mouse mode

### LINKED ISSUE #3546: [FR] Mouse drawing mode
Tridactyl version: 1.21.1
Firefox version: Waterfox Ltd. Waterfox 78.9.0
Operating system: win

# Making a feature request

**Add a annotation/drawing mode**
This mode should allow the user to draw on the page.
This can be implemented using a mechanism similar to the no_mouse_mode, where a canvas is added which reacts to drawing commands. Ideally the canvas should be as big as the `document`(not just as big as the `window`), and should scroll with the page.

While I guess that this feature is out of the scope of this project, I'm just shooting my shot I guess 😅

DhruvaSambrani:
Add drawable no mouse mode. Closes #3546


DhruvaSambrani:
Fixes #3546

src/excmds.ts
```diff
@@ -1634,7 +1634,20 @@ export function snow_mouse_mode() {
 export function pied_piper_mouse_mode() {
     toys.music()
 }
-
+/**
+ * Drawable variant of [[no_mouse_mode]]
```
bovine3dom, line 1638: I think this needs a little more explanation - it's not clear that this means "you can click the mouse to draw"

src/content/toys.ts
```diff
@@ -7,14 +7,22 @@
  *
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
+
+let erasor = false;
```
bovine3dom, line 11: "eraser" is the usual spelling : )

src/excmds.ts
```diff
@@ -1634,7 +1634,20 @@ export function snow_mouse_mode() {
...
+//#content
+export function drawable_mouse_mode() {
+    toys.drawable()
+}
+/**
+ * Toggle pen and erasor for drawable
+ */
+//#content
+export function toggle_pen() {
+    toys.toggle_pen()
```
bovine3dom, line 1649: I think it makes more sense to call this function `eraser_toggle` as I think it isn't obvious that `toggle_pen` swaps the pen for an eraser.

Also we generally try to follow the `[object][action]` naming convention

src/excmds.ts
bovine3dom: We should also mention toggle_pen in this help

src/excmds.ts
```diff
@@ -1634,7 +1634,20 @@ export function snow_mouse_mode() {
...
-
+/**
+ * Drawable variant of [[no_mouse_mode]]
+ */
+//#content
+export function drawable_mouse_mode() {
+    toys.drawable()
+}
+/**
+ * Toggle pen and erasor for drawable
```
bovine3dom, line 1645: Should link to drawable mouse mode here

bovine3dom (commented):
Just some bike shedding : )

DhruvaSambrani:
Hi, thanks for your review!

I only now got a build working, and I noticed that the previous implementation was not correct. I've made a lot of changes, and I'll commit it soon. I meant to change this to a WIP, but you reviewed before 😅

src/content/toys.ts
```diff
@@ -32,17 +32,88 @@ function makeBlock() {
...
+}
+
+function make_drawable(overlaydiv) {
+    overlaydiv.style.position = "absolute"
+    const c = document.createElement("canvas")
+    overlaydiv.appendChild(c)
+    const context = c.getContext("2d")
+    // making the canvas full screen
+    c.height = document.documentElement.scrollHeight
+    c.width = window.innerWidth*0.98
```
DhruvaSambrani, line 96: canvas overflows if the 0.98 is not added. I do not know how to fix this

DhruvaSambrani:
Hi @bovine3dom
I think the PR is now ready for your review.
3 potential things we could add/fix- 
- [ ] Add a function to set pen color
- [ ] Change cursor to a pen and an erasor
- [ ] Fix that pesky `*0.98` as in the review comment

bovine3dom:
> Add a function to set pen color

At that point it might be better to add an extra mode to Tridactyl with the bindings. E.g.

```
:bind \d mode graffiti
:bind --mode=graffiti <Space> drawing.clear
:bind --mode=graffiti <Escape> composite drawing.clear; mode normal
```

etc? (With the namespace done in a similar way to `text.` or `ex.`).

src/content/toys.ts
```diff
@@ -32,17 +32,88 @@ function makeBlock() {
...
-    overlaydiv.style.right = "0"
-    overlaydiv.style.bottom = "0"
-    overlaydiv.style.zIndex = "1000"
-    overlaydiv.style.opacity = "0.5"
+    overlaydiv.style.height = String(document.documentElement.scrollHeight)
+    overlaydiv.style.top = "0px"
+    overlaydiv.style.bottom = "0px"
+    overlaydiv.style.left = "0px"
+    overlaydiv.style.right = "0px"
+    overlaydiv.style.zIndex = "100"
```
bovine3dom, line 40: Why did you reduce the zIndex by a factor of 10? It used to be 1000

src/content/toys.ts
```diff
@@ -32,17 +32,88 @@ function makeBlock() {
...
+    overlaydiv.style.opacity = "0.8"
```
bovine3dom, line 41: I'd prefer that we made this configurable, e.g.

`function music(opts = {opacity: 0.5})`

as it makes `neo_mouse_mode` less usable. Which I realise is a niche concern...

src/content/toys.ts
bovine3dom: Should we care? It seems to work fine with the magic `*0.98`. I'm happy for weird workarounds to stay in the code - just add a comment saying what it does

src/excmds.ts
```diff
@@ -1634,11 +1634,28 @@ export function snow_mouse_mode() {
...
+ * Use [[mouse_mode]] to return, or refresh page.
+ * Suggested usage: `autocmd DocLoad .* drawable_mouse_mode`
+ */
+//#content
+export function drawable_mouse_mode() {
+    toys.drawable()
+}
+/**
+ * Toggle pen and eraser for [[drawable_mouse_mode]]
+ * Suggested usage: `bind e eraser_toggle` and map pen button to `e`
```
bovine3dom, line 1649:
```suggestion
 * Switch between pen and eraser for [[drawable_mouse_mode]]
 * Suggested usage: `bind e eraser_toggle`. If you have a digital pen, map the button to `e` to switch easily.
```

bovine3dom (commented):
I think it would also make sense to add the ability to save screenshots through Tridactyl via https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/captureVisibleTab

otherwise the doodling will always be ephemeral, which is a nice philosophical statement but perhaps less useful to users

src/content/toys.ts
bovine3dom: Configurable is not the right word - I don't mean `config.get`, I mean such that `drawable_mouse_mode` can set it to 0.8 when it is called

src/content/toys.ts
DhruvaSambrani: That looks like a mistake. Should I revert?

DhruvaSambrani:
> I think it would also make sense to add the ability to save screenshots through Tridactyl...

I agree! Where should I put it though?

src/content/toys.ts
bovine3dom: yes - `1000` is enough that stuff appears on top of most page elements. `100` might not be

bovine3dom:
Screenshots should be a different issue / PR, probably. And then just an excmd - probably `:screenshot [destination]` where destination can be a file path or something magic to put it in the clipboard. Maybe `-t[tab / wintab number]` to optionally specify a non-active tab via https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/captureTab

I think we have a data URI -> image blob thingy somewhere - it would have been written when we added "copy hinted image to clipboard"

src/content/toys.ts
DhruvaSambrani: If its only that, can we not just change the attribute in `make_drawable`? I've already done that for `style.position`.

DhruvaSambrani:
> Screenshots should be a different issue / PR, probably ...
Opened #3611

bovine3dom:
BTW - I've just noticed that you've made this PR from your master branch.

That's usually a bad idea because if you add any unrelated commits to your own master branch, they'll get added to this PR too. I've done it myself a few times and it's always embarrassing.

(I might have already told you this - I say it a lot - if you've balanced the risk/reward for yourself and decided to use master on purpose that's fine :) ).

I personally always set `master` on my fork branches to follow the upstream master branch, as it makes pulling in changes easier:

```
git remote add upstream git@github.com:tridactyl/tridactyl.git
git fetch --all
git branch master --set-upstream upstream/main

# then just `git checkout master` and `git pull` to keep up to date with master
```

We should probably add this to contributing.md :thinking: 

---

but don't do anything until we've merged this PR, obviously

src/content/toys.ts
bovine3dom: Sure, it doesn't make any difference to me : )

DhruvaSambrani:
> balanced the risk/reward for yourself

No I just realized way too late 😂
Thanks again!

DhruvaSambrani:
> > Add a function to set pen color
> 
> At that point it might be better to add an extra mode to Tridactyl with the bindings. E.g...

Are we doing this? If so, how?

bovine3dom:
> Are we doing this? If so, how?

The pseudo-namespacing is pretty gnarly. It's probably not worth it actually.

We could have them all as normal ex-commands, e.g. `drawingerasertoggle` `drawingpensize {+,-,,}[number]`, `drawingpencolour`, `drawingstart` which might make more sense. Adding a mode is as simple as adding a few lines to src/lib/config

bovine3dom:
As for whether we're doing it - that's up to you : )

DhruvaSambrani:
I'm putting this back into a WIP as it doesn't work for a pen tablet input. Probably a different event needs to be called. 😞

bovine3dom:
Probably need to use these events: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events#pointerevent_interface

bovine3dom:
It _could_ be cool to let people bind to mouse / pen buttons directly via `:bind`. https://github.com/tridactyl/tridactyl/pull/1814 did something similar for gamepads ... definitely something for a different PR / issue though.

DhruvaSambrani:
Opened #3613 accordingly

DhruvaSambrani:
@bovine3dom  I am now reasonably sure that everything works 😂

src/excmds.ts
```diff
@@ -1634,11 +1634,30 @@ export function snow_mouse_mode() {
 export function pied_piper_mouse_mode() {
     toys.music()
 }
-
+/**
+ * Drawable variant of [[no_mouse_mode]]
+ * In this mode, you can use the mouse or a pen table to draw. To erase any drawings, use [[eraser_toggle]]
```
bovine3dom, line 1639:
```suggestion
 * In this mode, you can use the mouse or a digital stylus to draw. To switch to an eraser, use [[eraser_toggle]]
```

I'm not 100% sure what you meant by "pen table" : )

src/content/toys.ts
```diff
@@ -32,17 +32,98 @@ function makeBlock() {
...
+    const c = document.createElement("canvas")
+    overlaydiv.appendChild(c)
+    const context = c.getContext("2d")
+    // making the canvas full screen
+    c.height = document.documentElement.scrollHeight
+    c.width = window.innerWidth*0.98  // workaround to fix canvas overflow
+    c.style.touchAction = "none" // for pen tablet to work
+    c.addEventListener("pointerdown", e=>handleDown(e,context))
+    c.addEventListener("pointerup", handleUp)
+    c.addEventListener("pointermove", e=>handleMove(e,context))
```
bovine3dom, line 124: When we do #3613 these event listeners will need to be removed

(no action needed now - this comment should appear in the issue and remind us)

bovine3dom:
Looks good. I need to spend a little more time looking at it before I merge it

I think I'll rename the commands a bit to try to fit our naming convention a bit better:

```
drawable_mouse_mode -> drawingstart
exalias: no_mouse_mode -> drawingstop
eraser_toggle -> drawingerasertoggle
```

DhruvaSambrani:
Cool!

Just to clarify, `mouse_mode` needs to be mapped to `drawingstop`, and the `mouse_mode` binding still needs to exist

DhruvaSambrani:
Also, 100 lines :party:

bovine3dom:
Yes, sorry, I meant that, not no_mouse_mode. https://github.com/tridactyl/tridactyl/blob/8d58ed953526f6c4253dcda54e8b17a3896c8068/src/lib/config.ts#L544 do leave the original names in place

PR #3628, opened 2021-04-29
Swap ramda for rambda

bovine3dom:
It supports TS a bit better and is a bit quicker.

But it has fewer functions.

```

 > src/completions/Tab.ts:243:23: warning: Import "differenceWith" will always be undefined because there is no matching export
    243 │         const diff = R.differenceWith(
        ╵                        ~~~~~~~~~~~~~~

 > src/.excmds_background.generated.ts:2354:6: warning: Import "forEachObjIndexed" will always be undefined because there is no matching export
    2354 │     R.forEachObjIndexed((tab, index) => {
         ╵       ~~~~~~~~~~~~~~~~~
```

In particular need replacing.

Related: #3582


Rummskartoffel:
`forEachObjIndexed()` seems to only be used once (excmds.ts:2763), where it should be replaceable with the following (untested):
```typescript
    Object.entries(windowTabs).forEach(([index, tab]) => {
        browser.tabs.move(tab.id, { index: parseInt(index, 10) })
    })
```

bovine3dom:
Thanks.

It was the difference algorithm I mostly wasn't looking forward to trying to replicate : )

Rummskartoffel:
Something like this should work (though I have no idea whether it's efficient, and if it isn't, how to do it more efficiently; also this doesn't dedupe like Ramda does, but for our use it shouldn't matter):
```typescript
const diff: BufferCompletionOption[] = [];
for (const prevOption of prevOptions) {
    if (!this.options.find(newOption => prevOption.tabId === newOption.tabId)) diff.push(prevOption)
}
```

tsconfig.json
```diff
@@ -6,15 +6,15 @@
     "noEmitOnError": true,
     "outDir": "build/tsc-out",
     "sourceMap": true,
-    "target": "es2017",
-    "lib": ["es2017","es2019.array", "es2018.promise", "dom", "dom.iterable"],
+    "target": "es2019",
+    "lib": ["es2019", "dom", "dom.iterable"],
```
Rummskartoffel, line 10: Note: [According to caniuse.com](https://caniuse.com/?feats=mdn-javascript_builtins_regexp_dotall,mdn-javascript_builtins_regexp_lookbehind_assertion,mdn-javascript_builtins_regexp_named_capture_groups,mdn-javascript_builtins_regexp_property_escapes,mdn-javascript_builtins_symbol_asynciterator,mdn-javascript_functions_method_definitions_async_generator_methods,mdn-javascript_grammar_template_literals_template_literal_revision,mdn-javascript_operators_destructuring_rest_in_objects,mdn-javascript_operators_spread_spread_in_destructuring,promise-finally), Firefox 68.0 doesn't support a tiny part of ES2018. I don't think it'll cause problems, but it might be good to keep in mind.

PR #3645, opened 2021-05-04
Webpack removal

### LINKED ISSUE #3498: Consider switching to esbuild

bovine3dom:
Just some stuff I sketched out that I didn't want to lose track of.

Need to fix the error mentioned in #3498 too.

Fixes #3498 once finished.


scripts/build.sh
```diff
@@ -69,17 +69,33 @@ if [ "$QUICK_BUILD" != "1" ]; then
...
+rmdir buildtemp
+
+# Copy extra static files across
+
+cp src/manifest.json build/
+
+# TODO: exclude psds and 1024px.png here
+# Try not to break builds on OSX / BSD in the process
+# //                        ignore: ["**/*.psd", "**/*1024px.png"],
+cp -r src/static build
```
bovine3dom, line 93: Blocked because of this - don't want to inflate the XPI size

Issue #3727, opened 2021-05-30
Remove content-security-policy weakening from exemplar RC


nathantypanski:
I understand this is "intended" behavior. However, I think it's important to have a bug report about this workaround: I was trying to troubleshoot something and was reading request/response headers, and when I saw Tridactyl had modified security headers on web pages I visited, I elected to stop using Tridactyl. I believe more users would do the same if they were aware of the implications.

I respect that this is a workaround for an upstream issue in Firefox, and not a problem rooted in Tridactyl. However, I think the right choice is to push for the changes necessary upstream vs adopting insecure settings on every webpage. In the meantime: if browser changes are necessary, then we should push for them, and if a more agreeable workaround exists, we should adopt it.

-   Brief description of the problem:

Tridactyl clobbers `content-security-policy` on login pages, e.g., accounts.google.com. This dramatically weakens browser protections against cross-site attacks.

Without Tridactyl:

```
content-security-policy: script-src  'nonce-rBNAvJVmxjdS6W10CyksyQ' 'unsafe-inline';object-src  'none';base-uri 'self';report-uri  /_/IdentityListAccountsHttp/cspreport;worker-src 'self'
```

With Tridactyl:

```
content-security-policy: script-src 'unsafe-eval';require-trusted-types-for 'script';object-src 'none'
```

-   Steps to reproduce:

    1. Open the developer console and inspect response headers.
    2. Reload any webpage that uses safe CSP settings.
    3. Note the `unsafe-inline`, `unsafe-eval`, etc. added to the CSP, replacing the site's settings.

-   Tridactyl version (`:version`): 1.21.1

-   Firefox version 89.0b15 (64-bit)

-   URL of the website the bug happens on:

   https://account.google.com

-   Contents of ~/.tridactylrc or ~/.config/tridactyl/tridactylrc (if they exist):

https://github.com/tridactyl/tridactyl/blob/c98aaf9e64a7ab0ed2471d4a6c958bdbd00a489c/.tridactylrc#L121-L140

bovine3dom:
Tridactyl doesn't touch the content security policy unless you tell it to.

If you remove that line from your RC file it will stop touching it.

If you have any recommendations for how to stop people from running that file blindly, I'd be happy to hear them.

In many ways everything in that file is a bad idea - if it was a good idea it would be default behaviour in Tridactyl. The point of it is to show what people can do with Tridactyl and with their RC files.

[...]

PR #3728, opened 2021-05-30
Ameliorate #3727: add even more warnings to RC file

bovine3dom:
It feels pretty daft to put all this stuff at the top of the RC file. There must be a better solution.

I'm toying with adding an undocumented `source --bovinestyle` which will strip leading `"`, then double commenting everything. People who want to edit the RC file and use commands can then still do so by uncommenting lines.


bovine3dom:
Merged

Issue #2485, opened 2020-06-14
Improve/Replace preventautofocusjackhammer


jgkamat:
*   Brief description of the problem:

(Originally) coming from wrappers/extensions like pentadactyl and qutebrowser, having the page grab focus is very annoying, as it invalidates all your (normal mode) bindings, which are frequently used to actually control the browser. Many sites grab focus both on page load and on various events through js, eg: through a timer. #326 attempted to fix this, but as documented in #393, this does not work reliably (and in my mind, that makes it useless). There is a "preventautofocusjackhammer" setting that (from my limited testing) seems to be much more reliable at blocking focus at the start of a page load, which is awesome. However, there are a couple issues preventing this from being practical.

1. This setting blocks some focus events via hinting (the github issue searchbox seems to be one for me). 
2. The setting disables itself after a "user interaction", which means a site that simply waits to focus elements after a user interaction will subvert this. (My personal use-case is custom bindings on chat websites focusing textboxes).

The solution surfingkeys found to this is (in my opinion) the best, which is:

1. [Listen for focus events, and blur them once you see one](https://github.com/brookhong/Surfingkeys/blob/936284a3c4abd4c859b5157a5dd8f945671ba724/content_scripts/normal.js#L422). I'm not sure if this is the best way, but any way of blocking all focus would work here. However preventautofocusjackhammer works at the moment seems to fullfill this.
2. [On a mouse-press or hint of an editable element, allow one focus-event through](https://github.com/brookhong/Surfingkeys/blob/936284a3c4abd4c859b5157a5dd8f945671ba724/content_scripts/normal.js#L416). This could be replaced by some lexical variable that is set on the hint function as well, if such a feature is available in js. This method would prevent a hint that does not focus anything but falsely detected as editable from allowing a single event through.
3. [Optionally on "passthrough/insert/ignore" mode, disable all of this blurring so the user can allow the page to grab focus as expected](https://github.com/brookhong/Surfingkeys/commit/21e9c5546a28647cc967bd56a5e4477a8f10445c#diff-7a5989442bcc6f7ea14c77ff6ddb9aa3).

With this setup I've been able to mirror the focus behavior of the more powerful plugins.

*   Tridactyl version (`:version`): 

1.19.1pre3928-53d0d18

*   Firefox version (Top right menu > Help > About Firefox):

 77.0.1 (64-bit)

*   URL of the website the bug happens on: 

[...]

PR #3740, opened 2021-06-01
Add setting to force page autofocus settings in certain modes

jgkamat:
This is a pretty simple change, just adding a setting so the behavior of "allowautofocus" can only be applied. This is point 3 in #2485 (if configured). By default, this setting preserves the current behavior.

Unfortunately the "prevent" vs "allow" wording is quite confusing - I tried to pick names the least confusing to me, but it's still pretty bad, let me know if you have better ideas :(. Maybe "preventallowautofocusmodes". I'm bad at names. I'm trying to avoid "allowautofocusmodes" as its unclear if a true value would enable the setting to change things on the page, or enable auto-focus  (which are opposite behaviors).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3796,6 +3796,8 @@ export function setmode(mode: string, key: string, ...values: string[]) {
     if (!mode || !key || !values.length) {
         throw new Error("seturl syntax: mode key value")
     }
+    if (key !== "allowautofocus")
+        throw new Error("Setting '" + key + "' not supported with setmode")
 
     return config.set("modesubconfigs", mode, ...validateSetArgs(key, values))
 }
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3778,6 +3778,28 @@ export function seturl(pattern: string, key: string, ...values: string[]) {
     return config.setURL(pattern, ...validateSetArgs(key, values))
 }
 
+/**
+ * Usage: `setmode mode key values`
+ *
+ * @param mode The Mode the setting should be set for, e.g. `insert` or `ignore`.
+ * @param key The name of the setting you want to set, e.g. `allowautofocus`
+ * @param values The value you wish for, e.g. `true`
+ *
+ * Currently this command is only supported for the following settings:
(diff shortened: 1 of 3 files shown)


src/excmds.ts
```diff
@@ -3778,6 +3778,28 @@ export function seturl(pattern: string, key: string, ...values: string[]) {
...
+ *
+ * Currently this command is only supported for the following settings:
+ * - [[allowautofocus]]
+ *
+ * Example:
+ * - `setmode ignore allowautofocus true`
+ */
+//#content
+export function setmode(mode: string, key: string, ...values: string[]) {
+    if (!mode || !key || !values.length) {
```
jgkamat, line 3796: It would be nice to do some check on mode being a valid ModeName, but I don't know how to do that, is there a way you can ask typescript for all the possible values for a type like that?

bovine3dom: Yeah it's something like `KeyOf TypeOf [mode object]` but it would get very messy quickly as users can make their own modes by just creating binds.

src/excmds.ts
```diff
@@ -3778,6 +3778,28 @@ export function seturl(pattern: string, key: string, ...values: string[]) {
...
+ * @param values The value you wish for, e.g. `true`
+ *
+ * Currently this command is only supported for the following settings:
+ * - [[allowautofocus]]
+ *
+ * Example:
+ * - `setmode ignore allowautofocus true`
+ */
+//#content
+export function setmode(mode: string, key: string, ...values: string[]) {
```
bovine3dom, line 3795: Might be wise to throw and error if key !== allowautofocus

src/excmds.ts
jgkamat: Ah, I didn't know that, I'll just leave things as is then for now

jgkamat:
Any chance I could get another round of review on this?

bovine3dom:
Ah, sorry, I completely forgot about this PR. Thanks for poking me - I'll look at it this afternoon : )

bovine3dom:
Works great, thanks!

Issue #3743, opened 2021-06-03
Urlmodify not working with composite


simonhughcom:
When using composite to run two `urlmodify` commands, the first one fails to run.
In my particular case I tried to run `composite urlmodify -g 2 search; urlmodify -s query searchterm`
The second `urlmodify -s query searchterm` correctly runs, but the first `urlmodify -g 2 search` fails to run.

bovine3dom:
Hi, thanks for filing this issue.

The trouble is that the instance of Tridactyl that runs composite is lost as soon as a new page is loaded. I'm not sure how we would go about fixing that.

You could try using `:jsb tri.excmds.urlmodify(...).then(()=>tri.excmds.urlmodify (...))` instead.

Or maybe even `:jsb tri.excmds.composite (...)`

simonhughcom:
No, neither of them work.
We could write a new argument for `urlmodify` that would make it accept an input as url and then pipe one into the other.
For example: `composite urlmodify_js -g 2 videos | urlmodify -i -s query searchterm URL_GOES_HERE`
Or even more simply `composite urlmodify_js -g 2 videos | urlmodify -s query searchterm -` where the dash means accept from stdin, to follow convention.

I can do it if you want me to, I just need to know how you would like it to be implemented.

simonhughcom:
> Or even more simply `composite urlmodify_js -g 2 videos | urlmodify -s query searchterm -` where the dash means accept from stdin, to follow convention.

Never mind, after looking at the composite function, I don't think this this version would work, it would have to be the first one with the `-i` flag.

bovine3dom:
`composite get_current_url | urlmodify_js ... | url_modify_js ... | open` seems very reasonable to me : )

simonhughcom:
So you want `urlmodify_js` to always accept an input rather than use the current URL?
That would break backwards compatibility, are you ok with that?

bovine3dom:
Ah, no, sorry. I didn't realise it did that - that was an odd choice to make.

I'd add a flag e.g. `u` that makes it accept a URL as the last argument, e.g. `urlmodify_js -tu www old www.reddit.com`. It shouldn't be too hard to implement, just remove the arg and the character once you've found them.

PR #3748, opened 2021-06-04
Add url input to urlmodify

simonhughcom:
Adds a new flag to allow `urlmodify` to accept a url as the last argument instead of using the current url as discussed in #3743.

To use simply add a `u` behind any flag, for example `urlmodify -su key value URL`

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1709,7 +1709,7 @@ export function urlparent(count = 1) {
  *  * -g <graftPoint> <newPathTail>
  */
 //#content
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -1725,11 +1725,19 @@ export function urlmodify(mode: "-t" | "-r" | "-s" | "-q" | "-Q" | "-g", ...args
  * `:composite urlmodify_js -t www. old. | tabopen `
  */
```
bovine3dom, line 1726: Could you mention the new `u` in the help? Thanks!

simonhughcom: I updated the help message.

PR #3835, opened 2021-07-16
Add a :tabrename command

### LINKED ISSUE #3577: Add a :tabrename command
Tridactyl version: 1.21.1
Firefox version: Mozilla Firefox 87.0
Operating system: linux

Hi, I really would appreciate the ability to rename the tab title.
Reason: I always have a bunch of the same url open in different tabs, with slightly different content, for comparison reasons.
To distinguish them, this feature would be really helpful.   

Thanks in advance,
Tobix

ivcz:
Closes #3577

### DIFF
--- a/src/completions/Tab.ts
+++ b/src/completions/Tab.ts
@@ -68,7 +68,14 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
 
     constructor(private _parent) {
         super(
-            ["tab", "tabclose", "tabdetach", "tabduplicate", "tabmove"],
+            [
+                "tab",
+                "tabclose",
+                "tabdetach",
+                "tabduplicate",
+                "tabmove",
+                "tabrename",
+            ],
             "BufferCompletionSource",
             "Tabs",
         )
--- a/src/completions/Tab.ts
+++ b/src/completions/Tab.ts
@@ -96,6 +96,8 @@ export class BufferCompletionSource extends Completions.CompletionSourceFuse {
 
     async filter(exstr) {
         this.lastExstr = exstr
+        const prefix = this.splitOnPrefix(exstr).shift()
+        if (prefix === "tabrename ") this.shouldSetStateFromScore = false
         return this.onInput(exstr)
     }
 
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3524,6 +3524,31 @@ export async function taball(...id: string[]) {
     return tab_helper(true, true, ...id)
 }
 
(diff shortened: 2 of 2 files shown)


bovine3dom:
https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/Tab.ts#L84 is slightly annoying here - if you want to rename a tab to something close enough to the original name so that the tab is still highlighted, when you press `Enter` it will erase the title you typed and give it a blank title

I think the decent solution probably is something like https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/History.ts#L61 where we ensure that `tabrename` only has completions for the first argument. The best solution is probably to change https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/Tab.ts#L84 to instead be set per excmd rather than per completion, but that's quite a bit messier.

What do you think?

ivcz:
Only having completions for the first argument seems to actually be really important. After reading this i tried typing the existing title of a tab and it resets after each space if the word was already in there.

I don't think changing 
https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/Tab.ts#L84
to per cmd conditions is the best idea since rename seems to be the only odd one out

ivcz:
I think i might have misunderstood you so just to be clear
Possible solutions are either a more complicated condition here
https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/Tab.ts#L84
Or implementing a filter like in src/completions/History.ts here
https://github.com/tridactyl/tridactyl/blob/c1132307c318a60b7a523f505629576d61c78ddc/src/completions/Tab.ts#L97
Am i correct?

ivcz:
Would this be an ok or is it too hacky?

``` 
   async filter(exstr) {
        this.lastExstr = exstr
        let [prefix, query] = this.splitOnPrefix(exstr)
        if (prefix === "tabrename ") {
            this.shouldSetStateFromScore = false
        }
        return this.onInput(exstr)
    }
```

bovine3dom:
That looks great, thanks!

I really don't like how completions work at all in Tridactyl - it's a slow mess of hacks with loads of copy-pasted code. One day I'm going to tear it all out and start from scratch so it's more configurable at runtime (we really should let people define their own completions, for example). But not today :)

I'll merge it once it passes CI.

PR #3876, opened 2021-07-31
Update Find mode section of readme with notes how to use Find in page

slmb3:
Update Find mode section of readme with notes how to use Find in page.
This is very useful for controlling find settings such as highlighting matches, case-sensitive matching, etc., and can be done without unbinding Ctrl+F.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -111,10 +111,12 @@ You can try `:help key` to know more about `key`. If it is an existing binding, 
 
 Find mode is still incomplete and uses the Firefox feature "Quick Find". This will be improved eventually.
 
--   `/` — open the find search box
+-   `/` — open the Quick Find search box
+-   `/` then `<C-f>` — open the Find in page search box
 -   `<C-g>`/`<C-G>` — find the next/previous instance of the last find operation (note: these are the standard Firefox shortcuts)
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -109,12 +109,14 @@ You can try `:help key` to know more about `key`. If it is an existing binding,
 
 #### Find mode
 
-Find mode is still incomplete and uses the Firefox feature "Quick Find". This will be improved eventually.
+Find mode is still incomplete and uses the Firefox feature "Quick find". This will be improved eventually.
```
bovine3dom, line 112:
```suggestion
Find mode is still incomplete and uses the Firefox feature "Quick Find". This will be improved eventually.
```

"Quick Find" is what Mozilla calls it so we should stick to that : )

bovine3dom (changes requested):
This looks good but I'm fairly sure Firefox always styles it as Quick Find - I'd rather we stuck with that for consistency.

Thanks for making this PR : )

readme.md
bovine3dom: Ah, I see that the Quick Find bar itself says `Quick find` in the search box, but I think that's just a normal imperative sentence, i.e. it has a capital "Q" just because it's the start of the sentence. So I still think we should call the bar the Quick Find bar :)

slmb3:
Fixed, thanks. Also re-worded a bit as I realized the find in page works when any input field is active, such as when typing this comment

bovine3dom:
Looks great, thanks!

Issue #3894, opened 2021-08-20
Investigate html-tagged-template.ts


bovine3dom:
1) it comes from here https://github.com/straker/html-tagged-template but is uncredited
2) no-one else seems to use it; surely there's a more mainstream way of doing what we're doing

Looks like it was added at about the same time completions were added, i.e. 2017.

PR #3895, opened 2021-08-21
Preact completions playground

bovine3dom:
Related to #3894 

Inferno doesn't play nicely with esbuild apparently so let's try Preact

---

Preact doesn't like you reaching in and changing HTML once you've ``` html`` ```d it so I'm going to bin everything and start from scratch without using object-oriented ~~anti-patterns~~ highly-respected corporate-grade development practices rather than try to fix it.

That does mean that I might never do anything


bovine3dom:
We should fix #3917 at the same time as this, it should be less painful

PR #3903, opened 2021-09-05
Make blacklistadd command add TabEnter

tribela:
## Pitch
`blacklistadd` command is good for some websites using keyboard shortcuts like gmail.
But it is get back to `normal` mode when switched tab.
Let `blacklistadd` command also adding `autocmd <url> TabEnter mode ignore`

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4058,7 +4058,7 @@ export function autocmddelete(event: string, url: string) {
  */
 //#background
 export function blacklistadd(url: string) {
-    return autocmd("DocStart", url, "mode ignore")
+    return autocmd("DocStart", url, "mode ignore") && autocmd("TabEnter", url, "mode ignore")
 }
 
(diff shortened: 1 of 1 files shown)


bovine3dom:
Hi, thanks for this PR.

I can't reproduce the issue you're trying to fix - tabs stay in ignore mode when I switch tab.

Are you on a very old version of Tridactyl? `1.21.1` is the current version.

You don't need to leave ignore mode to change tabs, by the way - if you press `ctrl+o` (`<C-o>` in Tridactyl-speak), the next key sequence will be accepted as if you were in normal mode. So, for example, `<C-o>b` brings up the list of tabs, `<C-o>J` switches tabs, etc. :)

---

If we do add a `TabEnter` autocommand, Tridactyl would fight anyone who manually changed a tab on a blacklisted page to normal mode whenever they changed tabs. So it's something I think we should avoid.

Issue #3882, opened 2021-08-06
Switch tridactyl theme depending on the system theme


azmeuk:
Some OS have dark and light variants of their themes. For instance GNOME have [an extension](https://extensions.gnome.org/extension/2236/night-theme-switcher/) that switches automatically the themes depending on the time or the luminosity.

Other Firefox extensions like [darkreader](https://addons.mozilla.org/fr/firefox/addon/darkreader/) can automatically detects the system theme, and adapt accordingly.

I think this would be awesome if tridactyl could let the users choose a dark theme and a light theme, and use the one or the other depending on the main system theme.

What do you think?

bovine3dom:
I think the easiest way to do this would be to make a new theme, e.g. "auto", and use the newish media directive https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme to change the relevant Tridactyl CSS variables. Since the default theme is always applied you would just need to add a dark mode rule.

aaronrancsik:
I guess this can be closed :)

bovine3dom:
If you had said "fixes #3882" in your PR it would have been closed automatically when the PR was merged :)

aaronrancsik:
Ops, next time I will keep in mind.

PR #3923, opened 2021-10-01
Add a new "auto" theme.

aaronrancsik:
This theme can adapt to system's light or dark theme variants. 
Solve: #3882


aaronrancsik:
I'm just copied the whole "default" and "dark" theme  CSS into the corresponding media queries.. 
I bet there is more elegant solution because this way every time the "default" or "dark" theme is changing this theme is require an update.
Let me know how horrible my PR :D I will fix it.

bovine3dom:
In theory you should be able to use `import` statements with relative URLs, conditionally loaded using media queries as shown here under examples: https://developer.mozilla.org/en-US/docs/Web/CSS/@import

I've never worked out how to do it but I'm not sure I have ever tried.

aaronrancsik:
@bovine3dom All right I will give it a try thanks.

bovine3dom:
Works great, thanks!

PR #3926, opened 2021-10-02
Remove most searchurls

### LINKED ISSUE #3925: Get rid of most/all searchurls
Tridactyl version: 1.21.1pre6005-83b1f7c7
Firefox version: Mozilla Firefox 88.0
Operating system: linux

I think the only ones I ever use on purpose are `google`, `bing`, `scholar` and `duckduckgo`.

Would anyone object to us removing most of them? That'd be:

```
    googlelucky: "https://www.google.com/search?btnI=I'm+Feeling+Lucky&q=",
    googleuk: "https://www.google.co.uk/search?q=",
    yahoo: "https://search.yahoo.com/search?p=",
    twitter: "https://twitter.com/search?q=",
    wikipedia: "https://en.wikipedia.org/wiki/Special:Search/",
    youtube: "https://www.youtube.com/results?search_query=",
    amazon: "https://www.amazon.com/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=",
    amazonuk:
        "https://www.amazon.co.uk/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=",
    startpage:
        "https://startpage.com/do/search?language=english&cat=web&query=",
    github: "https://github.com/search?utf8=✓&q=",
    searx: "https://searx.me/?category_general=on&q=",
    cnrtl: "http://www.cnrtl.fr/lexicographie/",
    osm: "https://www.openstreetmap.org/search?query=",
    mdn: "https://developer.mozilla.org/en-US/search?q=",
    gentoo_wiki:
        "https://wiki.gentoo.org/index.php?title=Special%3ASearch&profile=default&fulltext=Search&search=",
    qwant: "https://www.qwant.com/?q=",
```

(you can obviously re-add them with e.g. `:set searchurls.mdn https://developer.mozilla.org/en-US/search?q=`)

bovine3dom:
I kept using them by accident.

You can add them back with `:set searchurls.xyz url`

Closes #3925 - happy to hear arguments for specific ones which should be saved

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -741,31 +741,32 @@ export class default_config {
      * - When running `tabopen translate en ja Tridactyl`, with a `translate` searchurl defined with `set searchurls.translate https://translate.google.com/#view=home&op=translate&sl=%s1&tl=%s2&text=%s3`, tridactyl will navigate to `https://translate.google.com/#view=home&op=translate&sl=en&tl=ja&text=Tridactyl`.
      *
      * [[setnull]] can be used to "delete" the default search engines. E.g. `setnull searchurls.google`.
+     *
+     * NB: lots of searchurls were removed in May 2024. You can restore them with:
+     *
+     * ```
+     *     set searchurls.googleuk https//www.google.co.uk/search?q=
+     *     set searchurls.yahoo https//search.yahoo.com/search?p=
+     *     set searchurls.twitter https//twitter.com/search?q=
+     *     set searchurls.wikipedia https//en.wikipedia.org/wiki/SpecialSearch/
+     *     set searchurls.amazon https//www.amazon.com/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=
+     *     set searchurls.amazonuk https//www.amazon.co.uk/s/ref=nb_sb_noss?url=search-alias%3Daps&field-keywords=
+     *     set searchurls.startpage https//startpage.com/do/search?language=english&cat=web&query=
+     *     set searchurls.github https//github.com/search?utf8=✓&q=
+     *     set searchurls.searx https//searx.me/?category_general=on&q=
+     *     set searchurls.cnrtl http//www.cnrtl.fr/lexicographie/
+     *     set searchurls.osm https//www.openstreetmap.org/search?query=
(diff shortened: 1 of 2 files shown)


glacambre (approved):
I think you will get a lot of new issues about this change but I'm not against it :)

src/lib/config.ts
```diff
@@ -658,31 +658,33 @@ export class default_config {
...
         google: "https://www.google.com/search?q=",
         googlelucky: "https://www.google.com/search?btnI=I'm+Feeling+Lucky&q=",
         scholar: "https://scholar.google.com/scholar?q=",
-        googleuk: "https://www.google.co.uk/search?q=",
         bing: "https://www.bing.com/search?q=",
         duckduckgo: "https://duckduckgo.com/?q=",
-        yahoo: "https://search.yahoo.com/search?p=",
-        twitter: "https://twitter.com/search?q=",
-        wikipedia: "https://en.wikipedia.org/wiki/Special:Search/",
-        youtube: "https://www.youtube.com/results?search_query=",
```
bovine3dom, line 672: I actually use this one so I should add it back

davidscotson:
Adding a character to make it unlikely to be duplicated by a normal search term seems to be a standard workaround for this, e.g. in the duckduckgo bang searches:

https://duckduckgo.com/bang?c=News&q=bbc

This also opens the door to including a more comprehensive set. The duck duck go ones are accessible (https://duckduckgo.com/bang.v255.js), but presumably there's a curated open source list somewhere that could be auto-converted to work with tridactyl settings.

(I was looking into automatically grabbing the search from sites that advertise it with opensearch but a human created list seems like it's probably more widely useful)

bovine3dom:
Including the bangs is a neat idea. Do you know what licence they're released under? We might not be able to include them.

p-victor:
Is this still planned? I use bangs from ddg and would like all of them removed, I manually remove them in my config for now but it seems clunky

bovine3dom:
Kinda. Personally I only ever use `google`, `scholar` and `youtube` on purpose, everything else is by accident and it's really annoying.

I think I'll remove them in the betas and see if anyone complains

PR #3938, opened 2021-10-03
Easy reader

### LINKED ISSUE #102: Keybindings lost in Reader View

bovine3dom:
Fix #102

Todo:

- ~~open in a new page/tab which will both make the back button work and stop the reader getting infected by the page's css~~
    - should probably add a triopen command that opens data/ish urls in a tridactyl environment? _is that safe?_
~~- add sane css~~ done


bovine3dom:
made some decent progress on this.

todo:

- santise html with https://github.com/mozilla/readability#user-content-security
- open in a clean tridactyl page so that we get rid of page scripts/css etc and can use the back button
- work out where the "hero" image on bbc new articles is going

bovine3dom:
Probably better to use `document.write` to replace the page since  other extensions should continue to work, which they wouldn't do on Tridactyl pages.

Also would reduces security risks by meaning there is no access to the browser API etc.

bovine3dom:
`moz-extension` links are stripped out here https://github.com/leizongmin/js-xss/blob/8884b21308315915ca5456a8552ddc23366c8325/lib/default.js#L182

it is probably a niche enough bug that it doesn't matter that they don't work. we could override the filter but we would then miss out on any updates to that filter

PR #3969, opened 2021-10-19
Add WIP cheaper URI change event

bovine3dom:
Broken at the moment

I think the problem is that I need to wrap my head around `this`, which I've been putting off for about 4 years now

Related: #3905 - if working, would allow us to work cheaply on SPAs too

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -51,14 +51,11 @@ const realwindow = (window as any).wrappedJSObject ?? window // wrappedJSObject 
 const triPushState = (hist => (
     (...args) => {
         const ret = hist(...args)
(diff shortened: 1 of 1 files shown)


bovine3dom:
That patch seems to work from my testing, thanks!

 I'll try it out with #3905 to see if it works on a real SPA soon :)

GHolk:
Hey, the `popstate` event handler defined in content script is not triggered when webpages use History API to manipulate history, so the autocmd HistoryState will not be triggered when go backfoward in history in SPA.

Maybe we should use WebExtension's [webNavigation.onHistoryStateUpdated](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onHistoryStateUpdated) event?

PR #3986, opened 2021-11-03
Fix the `f` and `F` hints on DDG in the sample `.tridactylrc`

eunikolsky:
The fix is to use the [`~=` attribute selector][selector] to match the
result class instead of `=`; it works on DuckDuckGo now.

[selector]: https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#syntax

### DIFF
--- a/.tridactylrc
+++ b/.tridactylrc
@@ -71,9 +71,8 @@
 " bindurl www.google.com F hint -Jbc #search div:not(.action-menu) > a
 " 
 " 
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks for this! The failed checks aren't your fault obviously but I'd like to let the others run to see if they're real.

Will merge once the CI has finished :)

PR #4022, opened 2021-12-11
Fix tabopen on about:newtab

### LINKED ISSUE #3656: `:set newtab about:newtab` opens about:newtab if you have native installed
Tridactyl version: 1.21.1pre5624-ee7c2588
Firefox version: Mozilla Firefox 88.0
Operating system: linux

We should ignore `about:newtab` and just open our normal new tab page, which we do do, but the native messenger somehow kicks in and opens a real about:newtab too.

Steps to reproduce:

1. Install native
2. `:set newtab about:newtab`
3. `:tabopen`

Expected behaviour: our fake new tab page opens.

Actual behaviour: our fake new tab page opens followed by a real about:newtab opened by the native messenger

pvs:
Closes #3656

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -205,7 +205,7 @@ if (
     window.location.pathname === "/static/newtab.html"
 ) {
     config.getAsync("newtab").then(newtab => {
-        if (newtab !== "about:blank") {
+        if (!["about:blank", "about:newtab"].includes(newtab)) {
             if (newtab) {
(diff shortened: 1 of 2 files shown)


bovine3dom:
Thanks for this! I'll try to check it works and get it merged soon. I'll have to take your word that it works for Windows because I don't have easy access to one at the moment.

Rummskartoffel:
Seems to work on Windows.

PR #4024, opened 2021-12-13
Fix #3915

### LINKED ISSUE #3915: Prevent navigating from a page while editing
Tridactyl version: 1.21.1pre6005-83b1f7c7
Firefox version: Mozilla Firefox 88.0.1
Operating system: linux

When editing using `editorcmd` and accidentally navigating away from the page, there is no way to restore the link between the file being edited and the input field.

It's quite common on the Web for webpages to ask the user to confirm whether to leave or stay on the page while things are being edited. Would it be possible to put such a safety guard in place while `editorcmd` is running?

pvs:
Closes #3915

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -301,16 +301,13 @@ export function getInputSelector() {
 /** @hidden */
 //#content
 export function addTridactylEditorClass(selector: string) {
-    const elem = document.querySelector(selector)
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks for this, works great! Sorry it took so long to merge :)

PR #4038, opened 2021-12-29
Allow mode change to restore mode indicator.

vintnes:
Fixes issue #2690
Rather than delete the statusIndicator span, apply the TridactylInvisible class. This class includes pointer-events:none, which prevents conditional overlap with the statusIndicator EventListener "mouseenter".


bovine3dom:
Works great, thanks, and happy new year!

PR #4073, opened 2022-01-17
Add native install note for FF installed from Snap

gerardbosch:
It seems that containerised installation of Firefox cannot access `native`: https://github.com/tridactyl/tridactyl/issues/2406


bovine3dom:
Thanks! Hope you didn't lose too much time before you found out it wasn't possible :)

PR #4114, opened 2022-02-19
Exclude current window from tabpush list

mhijazi1:
Exclude the current window when generating the list of windows for `tabpush`


bovine3dom:
Works great, thanks!

PR #4115, opened 2022-02-19
Fix tridactyl#2510 use auto instead of default.css for content pages

davidscotson:
Just started looking at the code to generate my own theme, but this seems like a small improvement that might be welcome more generally.

Basically, the pages that previously loaded default.css, then had a short delay while the user's tridactyl theme kicked in, will now load auto.css, getting the default or dark theme depending on Firefox level user preference for the short time before their theme choice takes over, reducing or eliminating the brief flash for people using dark themes for both Firefox and Tridactyl.

I needed to change auto.css to always load default, then optionally load dark as it seems like dark on it's own is missing some styles (e.g. the commandline has no background set). That's probably more of a change than I'd want to make, and maybe there's a better way to do that. I'm possibly looking at making all the themes be auto switchable (web extensions seem the perfect excuse to use the latest CSS features) so happy to revisit that if it has negative impacts.

Only tested on Mac, but seems to work with the built in dark and light theme and the one that switches based on the mac system preference.

I'm possibly looking at making all the themes be auto switchable and loading them directly, so even the minor changes from e.g. dark to midnight shouldn't happen, but that's a bigger change.

### DIFF
--- a/src/static/css/authors.css
+++ b/src/static/css/authors.css
@@ -1,4 +1,4 @@
-@import url("../themes/default/default.css");
+@import url("../themes/auto/auto.css");
 
 /* Adapted from https://css-tricks.com/snippets/css/star-wars-crawl-text/ */
 body {
--- a/src/static/css/commandline.css
+++ b/src/static/css/commandline.css
@@ -1,4 +1,4 @@
-@import url("../themes/default/default.css");
+@import url("../themes/auto/auto.css");
 
 body {
     overflow: hidden;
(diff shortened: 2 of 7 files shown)


bovine3dom:
Thanks, this works great provided that the user has their Firefox / system setting set to dark.

Ideally we'd tell Firefox to be in dark mode with the `theme` API for themes that were dark so people could have the system setting different to the Tridactyl setting if they so desired

PR #4126, opened 2022-02-27
Add illegal character sanitisation in filename

gsbabil:
Firefox's download API [0] currently does not support invalid characters
in the "saveAs" value, and the following error is thrown if download
filename contains illegal character(s). The issue has been reported to
the Firefox team, and currently it is in "WONTFIX" status [1].

  "filename must not contain illegal characters"

The issue prevents Tridactyl from downloading files from URLs with
certain "illegal" characters, e.g. ":" (colon), that are valid for URLs,
but are rejected in the filenames by the underlying OS's file system.

The default behaviour for operating systems like Mac is to replace the
invalid character(s) with "space" when saving the file to the disk.

The patches here implement basic filename sanitisation support by
introducing the following three configuration parameters in config.ts:

  - illegalfilenamechars
  - illegalwindowsfilenames
  - illegalfilenamereplacement

Essentially, if the "illegalfilenamechars" are found in the downloaded
filename, these characters are replaced by the
"illegalfilenamereplacement" value. The same logic is applied to each of
the comma-separated "illegalwindowsfilenames" values; if any of these
names matches the "saveAs" file-name, the file-name is suffixed with the
"illegalfilenamereplacement" value.

[0] https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download
[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1390473

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -703,26 +703,6 @@ export class default_config {
     homepages: string[] = []
 
     /**
-     * Default set of characters that are to be considered illegal
-     * as download file-names.
-     */
-    illegalfilenamechars = "#%&{}\\<>*?/$!'\":@+`|="
-    /**
-     * Default value that will be used to replace the illegal
-     * character(s), if found, in the download file-name.
-     */
-    illegalfilenamereplacement = "_"
-
-    /**
-     * Comma-separated list of Windows file-names which, if matches
-     * with the download file-name, will be suffixed with the
-     * "illegalfilenamereplacement" value.
-     */
-    illegalwindowsfilenames = "CON, PRN, AUX, NUL, COM1, COM2,"
-        + "COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1,"
-        + "LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, LPT9,"
-
-    /**
      * Characters to use in hint mode.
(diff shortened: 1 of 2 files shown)


src/lib/config.ts
```diff
@@ -1191,6 +1208,10 @@ const platform_defaults = {
 (New-Object System.Net.WebClient).DownloadFile('https://raw.githubusercontent.com/tridactyl/native_messenger/master/installers/windows.ps1', '%TEMP%/tridactyl_installnative.ps1');\
 & '%TEMP%/tridactyl_installnative.ps1' -Tag %TAG;\
 Remove-Item '%TEMP%/tridactyl_installnative.ps1'"`,
+        downloadforbiddenchars: "#%&{}\\<>*?/$!'\":@+`|=",
```
bovine3dom, line 1211: Hey @gsbabil, thanks for this PR, I finally got around to looking at it :)

It looks like Linux/OSX permit many more characters in filenames than Windows does, so I have moved the Windows settings to the Windows-only defaults

I've also renamed the settings to match up better with the `downloadsskiphistory` setting in the hope that people will find them easier to find.

Let me know if you have any thoughts - otherwise I think this is good to merge.

bovine3dom:
Merged, thanks!

PR #4127, opened 2022-02-27
fix #4005 newtab changelog background color tweak

### LINKED ISSUE #4005: midnight theme doesn't colour changelog panel correctly on new tab page
Tridactyl version: 1.21.1pre6087-7e160a11
Firefox version: Mozilla Firefox 94.0.1
Operating system: linux

davidscotson:
Uses a transparent black that looks like #eee on white backgrounds, but works on dark backgrounds as well.

Some themes already override this, but this helps for the midnight and quake themes which don't, and lets new themes work by default, while still allowing them to override if they want to.

### DIFF
--- a/src/static/themes/default/default.css
+++ b/src/static/themes/default/default.css
@@ -101,7 +101,7 @@
     --tridactyl-of-bg: #ffec8b;
 
     /*new tab spoiler box*/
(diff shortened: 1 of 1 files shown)


bovine3dom:
ah, nice fix, thanks!

PR #4188, opened 2022-04-10
Fix #4158: bookmarklet execution bug

### LINKED ISSUE #4158: Bookmarklet in url-encoding does not execute, and the script tag is left in html after execution
-   Brief description of the problem:

    Bookmarklet in url-encoding does not execute,
    and the script tag is left in html after execution.

-   Steps to reproduce:

    1. `<a href="javascript:alert(%22hello%20world!%22)">hello</a>` does work in firefox, but does not work in tridactyl's `:open`
    2. Bookmarklet like [download html](https://github.com/GHolk/bookmarklet/blob/master/download-html.js) will download a broken html.

-   Tridactyl version (`:version`): 1.22.1

-   Firefox version (Top right menu > Help > About Firefox): Mozilla Firefox 91.7.0esr

-   URL of the website the bug happens on: all

-   Config (in a new tab, run `:viewconfig --user`, copy the url and paste it somewhere like gist.github.com): all

## solution
1. Maybe try `bookmarklet.replace(/(%[0-9A-Fa-f]{2})+/g, code => decodeURIComponent(code))`. `decodeURIComponent` will throw error in malform url, but firefox does not, so filter it first.
2. About second issue, maybe just `window.eval(bookmarklet)`? I am not familiar with execution permission in firefox addon. Are there any reason to use script tag instead of eval?

GHolk:
In original implementation, a script tag will leave in html,
and the bookmarklet in URL Encoding will not work.
Moreover, if bookmarklet return a string after evaluation,
the string should replace current page's html content.

One can execute bookmarklet by assign it to `window.location.href`,
but it does not work in add-on's context,
so I assign it with `window.eval()`.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -1326,7 +1326,7 @@ export async function open(...urlarr: string[]) {
         // Open URLs that firefox won't let us by running `firefox <URL>` on the command line
         return nativeopen(url)
     } else if (/^javascript:/.exec(url)) {
-        const escapeUrl = url.replace(/"/g, "\\\"")
+        const escapeUrl = url.replace(/[\\"]/g, "\\$&")
         window.eval(`window.location.href = "${escapeUrl}"`)
     } else {
         const tab = await ownTab()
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM, thanks!

PR #4215, opened 2022-04-28
Fix #822: add button to toggle Tridactyl

### LINKED ISSUE #822: Disable/Reenable toggle like uBlock and Privacy Badger?

bovine3dom:
Todo:

- toggle should work through a pop-up rather than from the button to make it less surprising (and to allow us to add more options such as toggling ignore mode)
- add feedback when Tridactyl is re-enabled, e.g. through the pop-up
- improve icon, maybe put a big red cross through it to make it more obvious that it is very disabled


src/background.ts
```diff
@@ -84,6 +84,31 @@ browser.tabs.onActivated.addListener(ev => {
...
+            "100": "static/logo/Tridactyl_red_100px.png",
+            "150": "static/logo/Tridactyl_red_150px.png",
+        }} :
+        { path: {
+            "64": "static/logo/Tridactyl_64px.png",
+            "100": "static/logo/Tridactyl_100px.png",
+            "150": "static/logo/Tridactyl_150px.png",
+        }}
+    if (toIgnore == "true") {
+        excmds_background.fillcmdline("Tridactyl disabled, refresh pages to take effect")
```
bovine3dom, line 104:
```suggestion
        excmds_background.fillcmdline_notrail("Tridactyl disabled, refresh pages to take effect")
```

PR #4289, opened 2022-06-25
Fix #4285: mktridactyl output wrong seturl command

### LINKED ISSUE #4285: Wrong seturl/bindurl output from mktridactylrc
mktridactylrc outputs wrong seturl/bindurl command if there are more than one settings in the same url.

## Steps to reproduce:
```
seturl my-domain config1 123
seturl my-domain config2 456
mktridactylrc // or with --clipboard
```

### expect output
```
seturl my-domain config1 123
seturl my-domain config2 456
```

### current output
```
seturl my-domain config1 123
set config2 456
```

## system
```
Tridactyl version: 1.22.1
Firefox version: Mozilla Firefox 101.0.1
Operating system: win
```

## how to fix
Don't use `.shift()` in https://github.com/tridactyl/tridactyl/blob/a8b30e8/src/lib/config.ts#L1884
prefix will be empty array in next key, so the output will not have url parameter.

GHolk:
If there are more than one seturl/bindurl config
in one url, `mktridactylrc` will outputs wrong
seturl/bindurl commands, which become global settings.

Because `prefix.shift()` will mutate the array.
Then in the next iteration, the subconfig and url in prefix
will miss.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1882,8 +1882,9 @@ const parseConfigHelper = (pconf, parseobj, prefix = []) => {
         if (typeof pconf[i] !== "object") {
             if (prefix[0] === "subconfigs") {
                 const pattern = prefix[1]
+                const subconf = [...prefix.slice(2), i].join(".")
                 parseobj.subconfigs.push(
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM, thanks!

PR #4293, opened 2022-06-30
Fix the hints on DuckDuckGo in the sample `.tridactylrc`

eunikolsky:
The webpage was updated and that broke the `f` and `F` hints in the
sample rc file (since commit 3bf7e4ab). This change updates the hints so
that they match the search result links again.


bovine3dom:
LGTM, thanks!

Issue #4290, opened 2022-06-25
New autocmd in webRequest event type will not overwrite the old one


GHolk:
When setting autocmd in webRequest event type, if there is already a registered autocmd,
the old one will not be overwritten, but both the old one and the new one will keep working.
This will cause unexpected behavior.

E.g: some one want to change the rule to cancel some http request,
and set `autocmds.*.BeforeRequest` to a new value.
In fact, the old one is still working, so there are two event listener;
if one of them return `{cancel: true}`, the request will be canceled.

## Expected behavior
* throw an error if there already exists a event listener
* remove the old listener before adding the new one
* name or number the event listeners so they can be removed specificly

I prefer to *throw an error*.
I can submit a PR if we reach a consensus.

GHolk:
The other unexpected behavior is that the old event listener will not be able to be removed.
Because autocmddelete will just delete the last event listener.

bovine3dom:
I think my alternative suggestion was to allow the naming of autocmds so that each RegEx could have multiple autocmds. Happy to merge #4309 in principle once I have time to check that it works :)

GHolk:
I think naming the autocmds will be a little complex to me. We have to implement add/list/delete operations and change the behavior of `autocmd*`.

Currently one can add multiple hook with regexp trick: `https://github.com/tridactyl[1]*` `https://github.com/tridactyl[2]*`,
for url pattern:  `https://github.com/tridactyl*`  `https://github.com/tridactyl**`, althought there are some limits.

bovine3dom:
I was just going to do e.g. `autocmd DocStart [RegEx] --name=[name] ...` with name being optional. I suppose we should document the RegEx trick but it's not exactly a well polished solution

PR #4309, opened 2022-07-15
Fix the new webRequest autocmd overwriting the old one

GHolk:
There are two commits:

 1. Fix the bug that the new web request autocmd overwrites the web request autocmd in the other url patterns.
 2. Throw an error when adding a new web request autocmd will overwrite the old one in config, which will cause problems.

Feel free to skip the second commit if you do not like it.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4471,10 +4471,8 @@ export function autocmd(event: string, url: string, ...excmd: string[]) {
     // rudimentary run time type checking
     // TODO: Decide on autocmd event names
     if (!AUCMDS.includes(event)) throw new Error(event + " is not a supported event.")
-    if (webrequests.requestEvents.includes(event)) {
-        if (config.get("autocmds", event, url)) {
-            throw new Error(`Autocmd ${event} for ${url} already exist`)
-        }
+    if (webrequests.requestEvents.includes(event) && config.get("autocmds", event, url)) {
+        throw new Error(`Autocmd ${event} for ${url} already exist`)
     }
     return config.set("autocmds", event, url, excmd.join(" "))
 }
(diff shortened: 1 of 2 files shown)


GHolk:
Just notice I did not link this pr to the issue #4290 .

bovine3dom:
```
yarn run v1.22.15
$ bash hooks/pre-commit
Running pre-commit hook...
Done in 0.07s.

/home/circleci/project/src/excmds.ts
  4474:5  error  Merge this if statement with the nested one  sonarjs/no-collapsible-if

✖ 1 problem (1 error, 0 warnings)


Exited with code exit status 1
CircleCI received exit code 1
```

Could you fix the linting error? :)

Sorry for the delay in getting to this, it's one of the (many) issues I missed when I didn't get emails for a few weeks. The outlook for the next few weeks isn't great either, I'll have extremely limited time until the last week of August. Sorry!

GHolk:
I will find a way that acceptable to the lint. I didn't use lint in my projects......

src/excmds.ts
```diff
@@ -4471,6 +4471,9 @@ export function autocmd(event: string, url: string, ...excmd: string[]) {
     // rudimentary run time type checking
     // TODO: Decide on autocmd event names
     if (!AUCMDS.includes(event)) throw new Error(event + " is not a supported event.")
+    if (webrequests.requestEvents.includes(event) && config.get("autocmds", event, url)) {
+        throw new Error(`Autocmd ${event} for ${url} already exist`)
+    }
```
bovine3dom, line 4476: I can't merge this because it will break RC files that innocently set exactly the same autocmd every time, so I'll merge without it :)

bovine3dom:
Merged in eed0f18f312b1c6ecd4ac1ddf355c54a5967d710, thanks!

PR #4328, opened 2022-08-03
Fix timestamp in completion of back command for shydactyl

GHolk:
The completion for back/forward can show the history list.
The item contain title, url and timestamp,
but timestamp does not show up in shydactyl theme,
because the `.url` and `.title` both takes `width: 50%;`,
so no space is left for the timestamp.

This commit remove the `.title`'s width contrain,
so the title will adjust its width automatically.
Also style `.time` to appropriate width and align right.

### DIFF
--- a/src/static/css/commandline.css
+++ b/src/static/css/commandline.css
@@ -84,6 +84,14 @@ input:focus {
     width: 50%;
 }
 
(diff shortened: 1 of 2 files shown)


bovine3dom:
Thanks!

PR #4339, opened 2022-08-10
Fix wrong empty selection detect

GHolk:
The anchorOffset and the focusOffset are counted from the anchorNode and the focusNode.
They may be different node, so a same offset does not mean empty selection.
Just isCollapsed is enough.


bovine3dom:
This bug annoys me so much, I'll be so happy if you've really fixed it! I'll try to review this as soon as possible :)

PR #4340, opened 2022-08-10
WIP: zoom-invariant UI scaling

dit7ya:
I am opening this WIP PR for feedback though there are some things which need to be done. Please let me know if you have suggestions on how to implement the following

- [ ] There needs to be a UI scaling setting which would determine the UI scaling. Or should this be left aside to themes and CSS totally?
- [ ] UI gets cut off when zoomed out to something like 70% or less.

I am not familiar with the codebase so I am suspecting there are elegant ways of achieving these.

### DIFF
--- a/src/content/commandline_content.ts
+++ b/src/content/commandline_content.ts
@@ -39,6 +39,24 @@ async function init() {
         hide()
         document.documentElement.appendChild(cmdline_iframe)
         enabled = true
+
+        // Determine the initial page zoom level when the it first loads
+        // So that we can use this value later
+        const zoomLevelOnPageLoad = Math.round(window.devicePixelRatio * 100)
+
+        window.addEventListener("resize", () => {
+            // Get the current zoom level once a resize event has been triggered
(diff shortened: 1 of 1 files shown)


src/content/commandline_content.ts
```diff
@@ -39,6 +39,24 @@ async function init() {
         hide()
         document.documentElement.appendChild(cmdline_iframe)
         enabled = true
+
+        // Determine the initial page zoom level when the it first loads
+        // So that we can use this value later
+        const zoomLevelOnPageLoad = Math.round(window.devicePixelRatio * 100)
```
bovine3dom, line 45: This introduces a bug:

1. zoom in on a page
2. reload / visit in a new tab
3. tridactyl is now zoomed in

bovine3dom (changes requested):
I think we can let CSS deal with UI scaling; no-one has ever asked for it.

I'm more concerned about the way bits of the UI disappear when you zoom out. Unimportant stuff disappears at 90% but at about 70% it starts to become unusable with completions getting hidden. Do you think this is something we can fix by changing the CSS in the default theme?

PR #4355, opened 2022-08-16
Add which-key like dynamic cheat sheet for keybinds

maradude:
# Rough first draft

Currently, it shows a somewhat similar cheat sheet as [which-key.nvim](https://github.com/folke/which-key.nvim) for a non-terminator key is pressed. Cannot yet be disabled.

Quite possible broken in many places as I only tested the happy path in `yarn run run` for basic commands without doing much else yet. Existing tests pass.

## Concerns

Most of the code is copied from cmdline, so I'm unsure about the additional performance overhead created and what
potential security concerns might exist in the current implementation.

## To-dos

Some to-dos that I thought should still be done:

- [ ] Add toggle to enable/disable cheat sheet e.g. `:whichkey enable/disable`
- [ ] Test for bugs. 
- [ ] Add tests.
- [ ] Improve presentation code (most of the presentation code stolen from command line, including creating a new iframe).
- [ ] Add command for showing binds without needing to press a non-terminal bind, e.g. `:whichkey-show normal`.
- [ ] Change default style (current style is kinda opinionated and non-standard)
- [ ] Add user styling

### DIFF
--- a/scripts/esbuild.js
+++ b/scripts/esbuild.js
@@ -1,11 +1,20 @@
-const esbuild = require('esbuild')
+const esbuild = require("esbuild")
 
-for (let f of ["content", "background", "help", "newtab", "commandline_frame"]) {
-        esbuild.build({
-        entryPoints: [`src/${f}.ts`],
-        bundle: true,
-        sourcemap: true,
-        target: "firefox68",
-        outfile: `buildtemp/${f}.js`,
-    }).catch(() => process.exit(1))
+for (let f of [
+    "content",
+    "background",
+    "help",
+    "newtab",
+    "commandline_frame",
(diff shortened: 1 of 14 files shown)


runar-rkmedia:
I would love for this feature to land. Is there something I can help with? Or is there perhaps some other plan to implement a similar functionality?

I could perhaps have a swing at some of the Todos lined out above this weekend, and see if the current code still works, seeing as it has been almost a year without activity.

bovine3dom:
Hi @runar-rkmedia - all of your suggestions for things to do sound sensible. Usually once a PR gets as old as this rebasing it on master gets annoying, so let me know if you need a hand.

PR #4391, opened 2022-09-21
Add --url option to define command in certain url and fix a recursion detection bug

GHolk:
Accept url option in command and output them in mktridactylrc with --url option. The original code can not output alias in subconfigs correctly (if they does exist). The doc about it and a example is added too.

The original code does not detect the recursive alias correctly. Thought it does detect, it detect before it define the alias; it should detect it after defining.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4092,10 +4092,15 @@ export async function tab_helper(interactive: boolean, anyWindow: boolean, ...ke
  * added in a future release will be SILENTLY overridden. Aliases are
  * expanded recursively.
  *
+ * If the `--url` option is given, the command will only be defined in the
+ * specified url pattern (just like `bindurl`). It is useful to overwrite
+ * command like urlparent or urlroot in some websites like github repositories.
+ *
  * Examples:
  *  - `command t tabopen`
  *  - `command tn tabnext_gt`
  *  - `command hello t` This will expand recursively into 'hello'->'tabopen'
+ *  - `command --url https://github.com/.+/.+/.+ urlroot urlmodify -g 2 .
  *
  * Commands/aliases are expanded as in a shell, so, given the commands above,
  * entering `:tn 43` will expand to `:tabnext_gt 43`. You can use this to create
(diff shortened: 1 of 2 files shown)


src/excmds.ts
```diff
@@ -4108,14 +4113,21 @@ export async function tab_helper(interactive: boolean, anyWindow: boolean, ...ke
...
+    const def = args.slice(1).join(" ")
     try {
-        const def = definition.join(" ")
+        if (url) await config.setURL(url, "exaliases", name, def)
+        else await config.set("exaliases", name, def)
         aliases.expandExstr(name)
-        return config.set("exaliases", name, def)
     } catch (e) {
-        config.unset("exaliases", name)
+        await config.unset("exaliases", name)
```
bovine3dom, line 4130: as i understand it in this code we: 
1. set the alias in the config
2. check that there are no loops (throw an error)
3. catch that error and unset the alias

so surely here there needs to be a matching unsetURL?

src/excmds.ts
```diff
@@ -4092,10 +4092,15 @@ export async function tab_helper(interactive: boolean, anyWindow: boolean, ...ke
  * added in a future release will be SILENTLY overridden. Aliases are
  * expanded recursively.
  *
+ * If the `--url` option is given, the command will only be defined in the
+ * specified url pattern (just like `bindurl`). It is useful to overwrite
+ * command like urlparent or urlroot in some websites like github repositories.
```
bovine3dom, line 4097: I had never thought of doing this, this is a brilliant idea :)

src/lib/config.ts
```diff
@@ -1925,7 +1925,12 @@ const parseConfigHelper = (pconf, parseobj, prefix = []) => {
                     parseobj.nulls.push(`setnull ${i}.${e}`)
                 } else if (i === "exaliases") {
                     // Only really useful if mapping the entire config and not just pconf.
-                    if (e === "alias") {
+                    if (prefix[0] === "subconfigs") {
+                        const url = prefix[1]
+                        const command = `alias --url ${url} ${e} ${pconf[i][e]}`
+                        parseobj.aliases.push(command)
+                    }
+                    else if (e === "alias") {
```
bovine3dom, line 1933:
```suggestion
                    } else if (e === "alias") {
```

Fix the linting error :)

>   1932:21  error  Closing curly brace does not appear on the same line as the subsequent block  brace-style

src/excmds.ts
GHolk: In fact, there are some problems for `urlparent`. If we `alias urlparent urlparent 2`, it will be expand recursively and infinitely. I'm not sure whether should we really add this feature.

bovine3dom: We could disallow loops in aliases and instead persuade people to do e.g. `:alias urlparent js tri.excmds.urlparent(2)`

PR #4414, opened 2022-10-10
Reduce flickering when opening new about:blank tab

richard96292:
This patch updates the dark theme background color to reduce flickering when opening a new about:blank tab. 

The videos below show the state before and after applying the patch. 

Before:
https://user-images.githubusercontent.com/68248740/194871356-0a2d53ab-9a30-4efb-852d-acb2c9182bb2.mp4

Notice how the background of the new about:blank tab is flickering.

After:
https://user-images.githubusercontent.com/68248740/194871154-056ba34a-73c0-43be-bf89-fa89c98d9a8d.mp4

Now the background color doesn't change when the extension has finished loading.

### DIFF
--- a/src/static/themes/dark/dark.css
+++ b/src/static/themes/dark/dark.css
@@ -229,8 +229,12 @@
 
     /* }}} */
 
+    /* 2021 firefox redesign changed the background color of a new tab.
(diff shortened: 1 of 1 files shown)


bovine3dom:
Looks great, thanks!

PR #4430, opened 2022-10-20
improve the fallbacks for opening tabs with hint

ThunderKey:
Firefox does not allow browser.tabs.create to create new tabs with file:// urls (see [MSDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/create#url)).

The current fallback just clicked the links and therefore replaced the current page.

The new workaround adds the appropriate modifier keys to the click. This allows the links to still be opened in tabs, either in the background or foreground.

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -696,7 +696,7 @@ export function simulateClick(
         (target as HTMLAnchorElement).target === "_blank" ||
         (target as HTMLAnchorElement).target === "_new"
     const href = (target as HTMLAnchorElement).href
-    if (href !== undefined && href.startsWith("file:")) {
+    if (href?.startsWith("file:")) {
         // file URLS cannot be opend with browser.tabs.create
         // see https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/create#url
         // still create a new tab
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -668,26 +668,53 @@ export function anchors(includeInvisible = false) {
     ])
 }
 
+export const enum TabTarget {
+    CurrentTab,
+    NewTab,
+    NewBackgroundTab,
+    NewWindow,
+}
+
+const tabTargetToModifierKey = {
+    [TabTarget.CurrentTab]: {},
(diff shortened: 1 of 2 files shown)


bovine3dom:
Hey, thanks for this PR. I think it would be better to add the desired target (current tab, new tab, new window) to the simulateClick function here https://github.com/tridactyl/tridactyl/blob/2f1f6520be5f93c855d0250b1f19e0733af5bf2f/src/lib/dom.ts#L672

That way we could preserve the workarounds for _blank and _new links which would otherwise be blocked by the popup blocker.

ThunderKey:
Thanks for the feedback. Adding options to the `simulateClick` is a good idea.

The issue with the "_blank" and "_new" workarounds is, that it uses `openInNewTab` which will fail in the case of file: urls.

I could therefore add it to the `simulateClick` function, but skip the "_blank" and "_new" workarounds in those cases and pass additional options to the `mouseEvent`, but I'm not sure if this makes the code very clear.

bovine3dom:
Yeah just add a case `&& !( [file url])` here https://github.com/tridactyl/tridactyl/blob/2f1f6520be5f93c855d0250b1f19e0733af5bf2f/src/lib/dom.ts#L680 maybe with a comment that says file URLs aren't supported by openInNewTab

src/lib/dom.ts
```diff
@@ -668,34 +668,61 @@ export function anchors(includeInvisible = false) {
...
     // magic 'short lived event handler' context.
     //
     // OTOH, hardly anyone uses that functionality any more.
-    if (
+    let usePopupBlockerWorkaround =
         (target as HTMLAnchorElement).target === "_blank" ||
         (target as HTMLAnchorElement).target === "_new"
-    ) {
+    const href = (target as HTMLAnchorElement).href
+    if (href !== undefined && href.startsWith("file:")) {
```
bovine3dom, line 699: You can simplify this to

```suggestion
    if (href?.startsWith("file:")) {
```

ThunderKey: Thanks for the tip :)

bovine3dom:
LGTM, thanks!

Issue #3635, opened 2021-04-30
Mode-specific settings


bovine3dom:
Seems like it could be useful in situations where you want to change settings within a page, e.g. for hiding the mode indicator when something has entered fullscreen:

```
setmode fullscreen modeindicator false
bind --mode=fullscreen 🕷🕷INHERITS🕷🕷 ignore
# should probably add a `:bindinheritance child parent` command to make the above simpler
autocmd FullscreenEnter mode fullscreen
autocmd FullscreenLeft mode normal
```

bovine3dom:
closing in favour of #5476

PR #4439, opened 2022-10-30
Add config option to use key codes instead of characters

treapster:
This is a patch intended to solve #983. While i was working on it i discovered two strange things: 
 
A [comment](https://github.com/tridactyl/tridactyl/blob/dbd5bd42990260c38ce0b482133a1e292891c76e/src/lib/keyseq.ts#L483) at `translateKeysUsingKeyTranslateMap` saying that  we can't add anything to `MinimalKey` because everything will break, and [another one](https://github.com/tridactyl/tridactyl/blob/dbd5bd42990260c38ce0b482133a1e292891c76e/src/lib/keyseq.ts#L493) just below it saying that we in fact use `MinimalKey` everywhere through the pipeline . So i tried to add a flag to MinimalKey to check if it was translated and didn't see any breakage, and then eliminated `KeyEventLike` from all functions beside top-level generator, since it actually needs to accept both types. But otherwise, we can just convert KeyboardEvnet to MinimalKey inside generator, optionally using event.key/event.code, and use it for everything else. Maybe it's too radical change or i don't understand something, so comments are appreciated.  

~~I'm also not sure if you like turning the option on by default, but i don't see much reasons not to. It's cool when things work out of the box, and it's highly relevant for people using any layout beside en/us.~~ But anyway it's somewhat drafty patch so i'm waiting for  your feedback.


src/lib/config.ts
```diff
@@ -536,6 +536,13 @@ export class default_config {
         hintmaps: "false",
     }
 
+    /**
+     * Instead of fetching actual character which depends on layout,
+     * use machine code of a key.
+     * Recommended for non-latin or non-qwerty users.
+     */
+    usecodesinsteadofkeys: "true" | "false" = "true"
```
bovine3dom, line 544: We should work hard to make this setting easy to find - probably mention internationalization, Dvorak, bépo, Russian in the title so it is easy to find.

In theory I'm happy to make it be the default for new users but we should use a config updater (and make sure it works!) to make sure it's false for current users. Otherwise we will break Tridactyl for ~hundreds of people overnight.

bovine3dom:
This looks great, thanks!

We should edit `bind` so that if the use key code setting is true (or maybe if a flag is given) it binds to the key code equivalent sequence rather than the characters given. E.g. if I'm using AZERTY with the keycode interpretation on and I do `:bind a`, we need to actually bind to `q` so that when I press `a` my bind is triggered.

We also need to make sure that it's impossible to feed fake keys to Tridactyl by websites - `.isTrusted` needs to be absolutely trustworthy otherwise malicious websites could hijack Tridactyl and we could get our second RCE CVE...

bovine3dom:
Having played with this for a bit it seems to work great (although it looks like it might have broken our automated tests).

I'm not 100% sure that we should be translating the keys in hint mode though - it's a little bit mind-bending to see a hint labelled "J" and then have to think about where J would be on QWERTY rather than just pressing the J key on your keyboard. Maybe we should allow people to change the setting per mode with `:setmode`

treapster:
@bovine3dom wow, i didn't think of bind. It's gets complicated because we receive raw string input and it seems like the only way to make it work is to make bind to listen for keyboard events. However, shortcuts are usually implemented this way in ton of native apps, so not impossible. But probably requires new command like :bindcode?

Regarding isTrusted, i didn't touch the checks and they all happen on actual KeyboardEvent before getting inside parsing infrastructure, so i expect it to work as it was.

As for hint mode, we should probably provide an option to disable it like you say, but overall i think the whole idea of the feature is that it always works like QWERTY regardless.

Speaking of tests, i tried to run them locally but they were running for eternity and seemed hung up, so i thought it's probably my setup(probably need beta firefox?). Looks like i was wrong. But is the runtime of 80+ minutes normal for the tests, or it's a consequence of my changes?

bovine3dom:
I've never actually tried to get the tests running locally. They're retried if they fail because selenium is fairly temperamental. I'll look at them but I'm assuming they are genuinely broken. They seem to be failing on `:rssexec` completions here https://github.com/tridactyl/tridactyl/blob/7d819fa1401d5192dd99feb2942c35e743589f60/e2e_tests/e2e.test.ts#L138 which is weird, maybe you have broken the sendKeys function here: https://github.com/tridactyl/tridactyl/blob/7d819fa1401d5192dd99feb2942c35e743589f60/e2e_tests/utils.ts#L56

Could you make a small PR fixing some random tiny thing in Tridactyl that I can merge? Once you do that I won't have to manually click to enable the tests each time you push something new because GitHub will decide that you're a trusted collaborator. 

---

For `:bind`, we probably want to adapt `:gobble` or `:nmode` to accept a key to terminate the sequence with as well as accepting a number of characters, and then return the collected key events to a callback rather than just appending the characters to an ex-string. We're already ~80% of the way there with `:gobble 3 fillcmdline bind` :)

E.g. `:alias bindkeycodewizard gobble --js <CR> keys -> tri.excmds.fillcmdline("bind", tri.util.[somefunctionyou'llwrite](keys))`

bovine3dom:
I should add that I'm happy for that bind/gobble mode feature to be in a separate PR ... in fact I'd probably prefer it to keep the size of this PR manageable :)

treapster:
Seems like i broke sendKeys as you say, because when i type :rssexec manually completions are shown properly.
Also thanks for mentioning gobble, i didn't know it was a thing. Great idea with using it + termination key, will make PR. I guess gobble should accept number | string or just string, and we assume no one uses numbers as terminators?

bovine3dom:
> we assume no one uses numbers as terminators?

I hadn't thought about that but I think it's a very reasonable thing to assume :) I don't think it matters whether you say number | string or just string.

src/lib/config.ts
treapster: How can we distinguish between an update and new installation?

treapster:
If i run `yarn run jest`, i am able to reproduce test failure locally. However, it reproduces even if i make `alwaysqwerty` false by default, and i get failure even if i check out master which is obviously not broken. This is very weird. And i have no idea how to debug these tests or at least make them produce some output/logs.

treapster:
> We should edit bind so that if the use key code setting is true (or maybe if a flag is given) it binds to the key code equivalent sequence rather than the characters given. E.g. if I'm using AZERTY with the keycode interpretation on and I do :bind a, we need to actually bind to q so that when I press a my bind is triggered.

I think it can be done simply by doing `keymap q a`.
And now after i implemented :bindkey, i'm starting to think that it is probably not that good an idea, because if you do bindkey and press q, you've bound to character q, and if you later do keymap q a, previous bind will stop working.  
I think to reduce chances of weird things happening, we should probably throw away bindkey and force everyone to use translatemaps, because bindkey ignores translation maps and thus can cause problems when you try to bind while having anything in translation map. And i don't think it's a good idea to use tranalation in bindkey either.

bovine3dom:
> I think it can be done simply by doing keymap q a.

For me the point of this PR is to stop people from having to do stuff like `keymap q a`. Why would someone run that later? I don't think we need to make sure that the features work well together.

The nice thing about the key binding wizard is that the binds will still work if you keep changing keyboard layout because your binds will always be in the canonical QWERTY form. `keymap` by contrast is really annoying if you change layouts regularly :)

treapster:
> > I think it can be done simply by doing keymap q a.
> 
> For me the point of this PR is to stop people from having to do stuff like `keymap q a`. Why would someone run that later? I don't think we need to make sure that the features work well together.
> 
> The nice thing about the key binding wizard is that the binds will still work if you keep changing keyboard layout because your binds will always be in the canonical QWERTY form. `keymap` by contrast is really annoying if you change layouts regularly :)

So we just leave it as it is and don't translate keypresses captured by bindkey?
If you do keymap with code interpretation, it will work the same regardless of layout, it'll just convert one canonical form to another. And you still need keymap for hint mode(if you're not used to qwerty), so it looks like if you do keymap q a and keymap a q for hinting and then try to use bindkey with any of mapped keys, you will bind to the other key, because the actual command will translate keys. Seems like we better translate keys in bindkey.

treapster:
As of tests, is turned out they use package in web-ext-artifacts that was only updated after signing. However, even with updated extension, they hang even before the callback to test() is called and as of now i don't know why.
<details>

```
yarn run v1.22.19
$ jest --bail --runInBand
 FAIL  e2e_tests/e2e.test.ts (100.753s)
  ● webdriver › `:rssexec` works

    : Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Error:

      124 | 
      125 | 
    > 126 |     test("`:rssexec` works", async () => {
          |     ^
      127 |         const driver = await getDriver()
      128 |         try {
      129 |             await sendKeys(driver, ":set rsscmd js "

      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (e2e_tests/e2e.test.ts:126:5)
      at Env.<anonymous> (node_modules/jasmine-fail-fast/dist/jasmine-fail-fast.js:57:26)
      at apply (node_modules/lodash/lodash.js:490:27)
      at Env.wrapper [as describe] (node_modules/lodash/lodash.js:5379:16)
      at Object.<anonymous> (e2e_tests/e2e.test.ts:21:1)

Test Suites: 1 failed, 1 of 6 total
Tests:       1 failed, 7 skipped, 8 total
Snapshots:   0 total
Time:        101.233s, estimated 201s

```  
</details>
This is not the same error as in CI, and it happens both on this branch and on master.  driver.sleep() didn't help, but i figured out it is possible to debug them in vs code with js debug terminal and jest --watch. But as of now, i couldn't find the issue.

treapster:
Its super weird, seems like the driver is completely not working on my setup, because when i jun jest an empty window opens, and it doesn't have tridactyl loaded. Ad opposed to `yarn run run` which properly opens window with tridactyl, but doesn't run any tests.

treapster:
Here's a full log just in case
<details>

```

~/randomshit/tridactyl$ scripts/build.sh --old-native

Using TypeScript 3.9.10 from /home/denis/randomshit/tridactyl/node_modules/typescript/lib
Rendering [========================================] 100%

Documentation generated at /home/denis/randomshit/tridactyl/generated/static/docs

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/lib/webext.ts:308:15:
      308 │         return eval(js)(rest)
          ╵                ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/lib/webext.ts:308:15:
      308 │         return eval(js)(rest)
          ╵                ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/lib/webext.ts:308:15:
      308 │         return eval(js)(rest)
          ╵                ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/background/webrequests.ts:24:21:
      24 │     const listener = eval(func) as Function
         ╵                      ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/lib/webext.ts:308:15:
      308 │         return eval(js)(rest)
          ╵                ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/.excmds_background.generated.ts:2584:218:
      2584 │ ...eId : argument == "--title" ? (l, r) => l.title < r.title : argument == "--url" || argument == "" ? (l, r) => l.url < r.url : eval(argument)
           ╵                                                                                                                                  ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/.excmds_background.generated.ts:4979:11:
      4979 │     return eval(jsContent)
           ╵            ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/.excmds_content.generated.ts:4274:14:
      4274 │             ? eval(config.callback)
           ╵               ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

▲ [WARNING] Using direct eval with a bundler is not recommended and may cause problems [direct-eval]

    src/.excmds_content.generated.ts:4800:11:
      4800 │     return eval(jsContent)
           ╵            ~~~~

  You can read more about direct eval and bundling here: https://esbuild.github.io/link/direct-eval

Installing manifest here: /home/denis/.mozilla/native-messaging-hosts/
Installing script here: XDG_DATA_HOME: /home/denis/.local/share/tridactyl

Successfully installed Tridactyl native messenger!
Run ':native' in Firefox to check.
'node_modules/cleanslate/docs/files/cleanslate.css' -> 'build/static/css/cleanslate.css'
~/randomshit/tridactyl$ web-ext build --source-dir build --overwrite-dest                                    ✔  22s    
Applying config file: ./package.json
Building web extension from /home/denis/randomshit/tridactyl/build
Destination exists, overwriting: /home/denis/randomshit/tridactyl/web-ext-artifacts/tridactyl-1.22.1.zip
Your web extension is ready: /home/denis/randomshit/tridactyl/web-ext-artifacts/tridactyl-1.22.1.zip
~/randomshit/tridactyl$ yarn run jest
yarn run v1.22.19
$ jest --bail --runInBand
 FAIL  e2e_tests/e2e.test.ts (100.745s)
  ● webdriver › `:rssexec` works

    : Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 100000ms timeout specified by jest.setTimeout.Error:

      124 | 
      125 | 
    > 126 |     test("`:rssexec` works", async () => {
          |     ^
      127 |         const driver = await getDriver()
      128 |         try {
      129 |             await sendKeys(driver, ":set rsscmd js "

      at new Spec (node_modules/jest-jasmine2/build/jasmine/Spec.js:116:22)
      at Suite.<anonymous> (e2e_tests/e2e.test.ts:126:5)
      at Env.<anonymous> (node_modules/jasmine-fail-fast/dist/jasmine-fail-fast.js:57:26)
      at apply (node_modules/lodash/lodash.js:490:27)
      at Env.wrapper [as describe] (node_modules/lodash/lodash.js:5379:16)
      at Object.<anonymous> (e2e_tests/e2e.test.ts:21:1)

Test Suites: 1 failed, 1 of 6 total
Tests:       1 failed, 7 skipped, 8 total
Snapshots:   0 total
Time:        101.265s, estimated 201s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
~/randomshit/tridactyl$
```
</details>

treapster:
Turns out driver.addExtensions() tried to install extensions fully, which is only possible on beta. By using firefox.Driver and driver.installAddon(extensionPath, /*temporaty*/true) i managed to at least make the test launch and get the same error as in CI.

treapster:
But even with patched tests they fail even on master branch with the following
<details>

```
yarn run v1.22.19
$ jest --bail --runInBand
 FAIL  e2e_tests/e2e.test.ts (165.945s)
  ● webdriver › `:guiset` works

    expect(received).toEqual(expected) // deep equality

    Expected: "userChrome.css written. Please restart Firefox to see the changes."
    Received: ""

      195 |                 await driver.sleep(2000)
      196 |                 expect(await driver.executeScript(`return document.getElementById("tridactyl-input").value`))
    > 197 |                 .toEqual("userChrome.css written. Please restart Firefox to see the changes.")
          |                  ^
      198 |                 expect(newProfiles.find(p => fs
      199 |                                         .readdirSync(path.join(p, "chrome"))
      200 |                                         .find(files => files.match("userChrome.css$")))

      at Object.<anonymous> (e2e_tests/e2e.test.ts:197:18)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

Test Suites: 1 failed, 1 of 6 total
Tests:       1 failed, 5 skipped, 2 passed, 8 total
Snapshots:   0 total
Time:        166.435s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
</details>

src/lib/config.ts
bovine3dom: In theory new installations of Tridactyl will have `:set configversion 0.0` - https://github.com/tridactyl/tridactyl/blob/65586aa61c549f83f3d2c335cd1ecd12d093f392/src/lib/config.ts#L77 - and existing installations will have a higher number.

So we should be able to add a line here https://github.com/tridactyl/tridactyl/blob/65586aa61c549f83f3d2c335cd1ecd12d093f392/src/lib/config.ts#L1591 that turns on the keycode mode.

Really though I think we should make it so the default config version is whatever the current config version is, then we could instead have an updater that set the keycode mode for the legacy installations and have the real default be easily found in the configuration. I'm not sure why it doesn't work like that currently.

I'm not very sure that the config updater works reliably - maybe we could replace https://github.com/tridactyl/tridactyl/blob/65586aa61c549f83f3d2c335cd1ecd12d093f392/src/background.ts#L287 with a `config.getAsync(...).then`

bovine3dom:
> But even with patched tests they fail even on master branch with the following

That's not very concerning, it probably means that you don't have `:native` working properly :)

bovine3dom:
```patch
diff --git a/e2e_tests/e2e.test.ts b/e2e_tests/e2e.test.ts
index 3994e059..da4f20bd 100644
--- a/e2e_tests/e2e.test.ts
+++ b/e2e_tests/e2e.test.ts
@@ -33,6 +33,7 @@ describe("webdriver", () => {
         const options = (new Options())
                 .setPreference("xpinstall.signatures.required", false)
                 .addExtensions(extensionPath)
+        options.setBinary("/usr/bin/firefox-developer-edition")
         if (env["HEADLESS"]) {
             options.headless();
         }
```

is enough for me to get the tests to run on master

bovine3dom:
The problem seems to be that on your branch the driver can't enter colons correctly.

Rather than `:set rsscmd ....` for the first test I get `:search scmd`, with hint mode flashing up briefly. To me that suggests that it is entering the `;s` hint mode - it must be triggering a semi-colon rather than a colon. It works fine for me using it as a human, though.

<details>
```
 FAIL  e2e_tests/e2e.test.ts (85.015s)
  ● webdriver › `:rssexec` works

    expect(received).toBeGreaterThan(expected)

    Expected: > 3
    Received:   0

      140 |             await driver.switchTo().frame(iframe)
      141 |             const elements = await driver.findElements(By.className("RssCompletionOption"))
    > 142 |             expect(elements.length).toBeGreaterThan(3)
          |                                     ^
      143 |             const url = await elements[0].getAttribute("innerText")
      144 |
      145 |             // Then, make sure rsscmd is executed and has the right arguments

      at Object.<anonymous> (e2e_tests/e2e.test.ts:142:37)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  ● webdriver › `:editor` works

    NoSuchWindowError: Browsing context has been discarded

      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:524:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:587:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:515:28)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Driver.execute (node_modules/selenium-webdriver/lib/webdriver.js:741:17)
      at Actions.perform (node_modules/selenium-webdriver/lib/input.js:1006:5)
      at Object.<anonymous> (e2e_tests/e2e.test.ts:165:17)

  ● webdriver › `:guiset` works

    WebDriverError: TypeError: can't access dead object

      at Object.throwDecodedError (node_modules/selenium-webdriver/lib/error.js:524:15)
      at parseHttpResponse (node_modules/selenium-webdriver/lib/http.js:587:13)
      at Executor.execute (node_modules/selenium-webdriver/lib/http.js:515:28)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)
      at Driver.execute (node_modules/selenium-webdriver/lib/webdriver.js:741:17)
      at Object.<anonymous> (e2e_tests/e2e.test.ts:193:17)

  ● webdriver › `:colourscheme` works

    expect(received).toMatch(expected)

    Expected substring: "TridactylOwnNamespace TridactylThemeDark"
    Received string:    ""

      219 |             await driver.sleep(100)
      220 |             expect(await driver.executeScript(`return document.documentElement.className`))
    > 221 |                 .toMatch("TridactylOwnNamespace TridactylThemeDark")
          |                  ^
      222 |         } catch (e) {
      223 |             fail(e)
      224 |         } finally {

      at Object.<anonymous> (e2e_tests/e2e.test.ts:221:18)
      at processTicksAndRejections (node:internal/process/task_queues:95:5)

  ● webdriver › `:setpref` works

    expect(received).toMatch(expected)

    Expected pattern: /user_pref\("a.b.c", "d"\);/
    Received string:  "user_pref(\"app.normandy.api_url\", \"\");
    user_pref(\"app.update.checkInstallTime\", false);
    user_pref(\"app.update.disabledForTesting\", true);
    user_pref(\"browser.dom.window.dump.enabled\", true);
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.cfr\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.cfr-fxa\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.message-groups\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.messaging-experiments\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.snippets\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.asrouter.providers.whats-new-panel\", \"null\");
    user_pref(\"browser.newtabpage.activity-stream.discoverystream.config\", \"[]\");
    user_pref(\"browser.newtabpage.activity-stream.feeds.snippets\", false);
    user_pref(\"browser.newtabpage.activity-stream.feeds.system.topstories\", false);
    user_pref(\"browser.newtabpage.activity-stream.fxaccounts.endpoint\", \"\");
    user_pref(\"browser.newtabpage.activity-stream.tippyTop.service.endpoint\", \"\");
    user_pref(\"browser.safebrowsing.blockedURIs.enabled\", false);
    user_pref(\"browser.safebrowsing.downloads.enabled\", false);
    user_pref(\"browser.safebrowsing.malware.enabled\", false);
    user_pref(\"browser.safebrowsing.passwords.enabled\", false);
    user_pref(\"browser.safebrowsing.phishing.enabled\", false);
    user_pref(\"browser.sessionstore.resume_from_crash\", false);
    user_pref(\"browser.shell.checkDefaultBrowser\", false);
    user_pref(\"browser.startup.homepage_override.mstone\", \"ignore\");
    user_pref(\"browser.startup.page\", 0);
    user_pref(\"browser.uitour.enabled\", false);
    user_pref(\"browser.warnOnQuit\", false);
    user_pref(\"browser.webapps.checkForUpdates\", 0);
    user_pref(\"datareporting.healthreport.documentServerURI\", \"http://%(server)s/dummy/healthreport/\");
    user_pref(\"datareporting.healthreport.logging.consoleEnabled\", false);
    user_pref(\"datareporting.healthreport.service.enabled\", false);
    user_pref(\"datareporting.healthreport.service.firstRun\", false);
    user_pref(\"datareporting.healthreport.uploadEnabled\", false);
    user_pref(\"datareporting.policy.dataSubmissionEnabled\", false);
    user_pref(\"datareporting.policy.dataSubmissionPolicyBypassNotification\", true);
    user_pref(\"devtools.console.stdout.chrome\", true);
    user_pref(\"dom.ipc.reportProcessHangs\", false);
    user_pref(\"extensions.autoDisableScopes\", 0);
    user_pref(\"extensions.enabledScopes\", 5);
    user_pref(\"extensions.installDistroAddons\", false);
    user_pref(\"extensions.update.enabled\", false);
    user_pref(\"extensions.update.notifyUser\", false);
    user_pref(\"focusmanager.testmode\", true);
    user_pref(\"general.useragent.updates.enabled\", false);
    user_pref(\"geo.provider.testing\", true);
    user_pref(\"geo.wifi.scan\", false);
    user_pref(\"hangmonitor.timeout\", 0);
    user_pref(\"idle.lastDailyNotification\", -1);
    user_pref(\"marionette.port\", 0);
    user_pref(\"media.gmp-manager.updateEnabled\", false);
    user_pref(\"media.sanity-test.disabled\", true);
    user_pref(\"network.manage-offline-status\", false);
    user_pref(\"network.sntp.pools\", \"%(server)s\");
    user_pref(\"plugin.state.flash\", 0);
    user_pref(\"remote.log.level\", \"Info\");
    user_pref(\"security.certerrors.mitm.priming.enabled\", false);
    user_pref(\"services.settings.server\", \"\");
    user_pref(\"startup.homepage_welcome_url\", \"about:blank\");
    user_pref(\"startup.homepage_welcome_url.additional\", \"\");
    user_pref(\"toolkit.startup.max_resumed_crashes\", -1);
    user_pref(\"xpinstall.signatures.required\", false);
    "

      233 |             await driver.sleep(2000)
      234 |             const file = fs.readFileSync(path.join(newProfiles[0], "user.js"), { encoding: "utf-8" })
    > 235 |             expect(file).toMatch(/user_pref\("a.b.c", "d"\);/)
          |                          ^
      236 |         } catch (e) {
      237 |             fail(e)
      238 |         } finally {

      at Object.<anonymous> (e2e_tests/e2e.test.ts:235:26)

  ● webdriver › `:tabopen<CR>` opens the newtab page.

    TimeoutError: Timed out waiting for promise to resolve after 10001ms

      at node_modules/selenium-webdriver/lib/webdriver.js:839:15
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:643:19)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)

  ● webdriver › `:tabopen https://example.org<CR>` opens example.org.

    TimeoutError: Timed out waiting for promise to resolve after 10000ms

      at node_modules/selenium-webdriver/lib/webdriver.js:839:15
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:643:19)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)

  ● webdriver › `:tabopen qwant https://example.org<CR>` opens qwant.

    TimeoutError: Timed out waiting for promise to resolve after 10001ms

      at node_modules/selenium-webdriver/lib/webdriver.js:839:15
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:643:19)
      at listOnTimeout (node:internal/timers:564:17)
      at processTimers (node:internal/timers:507:7)

Test Suites: 1 failed, 5 passed, 6 total
Tests:       8 failed, 376 passed, 384 total
Snapshots:   0 total
Time:        88.237s
Ran all test suites.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

</details>

treapster:
In my case i get :tabopen instad of search, but it's sure something with semicolons.

treapster:
Ok, the issue is simple - selenuim sends keyboardEvent which has `code: "Semicolon"` and `key: ":"`, and does not have Shift modifier. Which results in semicolon in `minimalKeyFromKeyboardEvent`.

treapster:
Here is a stupid workaround which assumes alwaysqwerty is true, just to check whether it will pass

treapster:
Also if firefox.Driver will work, we should probably use it, because it has temporary mode which works without betas and dev editions

treapster:
> Also if firefox.Driver will work, we should probably use it, because it has temporary mode which works without betas and dev editions, so anyone can run this tests

But CI seems to dislike it, moving back

treapster:
Wowzers, that works! Now we have to decide whether it's a good idea to test with alwaysqwerty=true and properly update config.

treapster:
And also, what do you thinka about that?

> So we just leave it as it is and don't translate keypresses captured by bindkey? If you do keymap with code interpretation, it will work the same regardless of layout, it'll just convert one canonical form to another. And you still need keymap for hint mode(if you're not used to qwerty), so it looks like if you do keymap q a and keymap a q for hinting and then try to use bindkey with any of mapped keys, you will bind to the other key, because the actual command will translate keys. Seems like we better translate keys in bindkey, or even in whole gobble

bovine3dom:
Oh, sorry, I had thought about that but I apparently forgot to tell you.

I think really the alwaysqwerty settings needs to be per mode, defaulting to off in hint mode. We don't support per-mode settings very well but you can see our current workaround in `:setmode`.

I think that would be easier than trying to support keymap and alwaysqwerty at the same time.

treapster:
I think disabling alwaysqwerty in hint or any other mode defeats the purpose and introduces inconsistency, and I wouldn't say calling translatekeysinplace or something makes it much more complex

treapster:
If we use alwaysqwerty + keymap, we can achieve behaviour of any layout user wants without being affected by actual layout chosen. But such thing is only possible if we always use keymap, including gobble.

treapster:
But the complexity of keytranslatemap is keytranslatemodes. Is it useful to have one mappings in one mode and other in another? From my naive view it will work better to either map everything in minimalKeyFromKeyboardEvent, or not map at all. What are use cases for mapping-in-one-mode-but-not-in-another behavior?

bovine3dom:
> What are use cases for mapping-in-one-mode-but-not-in-another behavior?

Personally I would want to use "alwaysqwerty" in modes where I'm predominantly using keys to trigger commands which I have engrained in my muscle memory, like shortcuts, so really just normal mode and whatever custom modes I add. In modes where I am typing out words/characters I would like to use my chosen keyboard layout, so in insert/input mode, ex-mode and hint mode.

I don't think you are suggesting that we should have alwaysqwerty apply to ex or insert mode. I think our only difference is in hint mode.

Perhaps part of the problem is that I am using vimperator style hint filtering so when I select a hint I have to type out a fragment of a word, which is much more like writing text in an input box than it is pressing the equivalent to the `j` key to scroll down.

bovine3dom:
> If we use alwaysqwerty + keymap, we can achieve behaviour of any layout user wants without being affected by actual layout chosen. But such thing is only possible if we always use keymap, including gobble.

I don't understand what you're saying here. Are you suggesting we should have `alwaysqwerty` on in all modes and if the user doesn't want `alwaysqwerty` to apply in one mode they should use `keymap` to undo it? That seems like we're asking them to do a lot of labour.

treapster:
> > What are use cases for mapping-in-one-mode-but-not-in-another behavior?
> 
> Personally I would want to use "alwaysqwerty" in modes where I'm predominantly using keys to trigger commands which I have engrained in my muscle memory, like shortcuts, so really just normal mode and whatever custom modes I add. In modes where I am typing out words/characters I would like to use my chosen keyboard layout, so in insert/input mode, ex-mode and hint mode.
> 
> I don't think you are suggesting that we should have alwaysqwerty apply to ex or insert mode. I think our only difference is in hint mode.
> 
> Perhaps part of the problem is that I am using vimperator style hint filtering so when I select a hint I have to type out a fragment of a word, which is much more like writing text in an input box than it is pressing the equivalent to the `j` key to scroll down.

AFAIK there is no such thing as ex mode, when you're typing a command in commandline, you're directly typing into html input field. So these keypresses are not treated by us and the options we discuss here have no effect. If i'm wrong, then of course any maps or keys shouldn't apply there. What i'm saying is that in any shortcut modes such as normal, hint, gobble and whatever others there is no reason to have separate switches for each. So, when we're dealing with such shortcut modes(which i believe are all modes in which we intercept individual keypresses), it's reasonable to have just two boolean switches in config - whether to use alwasqwerty, and whether to apply keytranslatemap. So i'm basically suggesting to ditch keytranslatemodes map to simplify things. And if we ditch it, we can use alwasqwerty=true and usetranslatemap=false as defaults. for new users

treapster:
So we will have just one more if block in minimalKeyFromKeyboadEvent which checks whether we want to use translation map and apply it, that's all. What do you think about that?

treapster:
> > > What are use cases for mapping-in-one-mode-but-not-in-another behavior?
> > 
> > 
> > Personally I would want to use "alwaysqwerty" in modes where I'm predominantly using keys to trigger commands which I have engrained in my muscle memory, like shortcuts, so really just normal mode and whatever custom modes I add. In modes where I am typing out words/characters I would like to use my chosen keyboard layout, so in insert/input mode, ex-mode and hint mode.
> > I don't think you are suggesting that we should have alwaysqwerty apply to ex or insert mode. I think our only difference is in hint mode.
> > Perhaps part of the problem is that I am using vimperator style hint filtering so when I select a hint I have to type out a fragment of a word, which is much more like writing text in an input box than it is pressing the equivalent to the `j` key to scroll down.
> 
> AFAIK there is no such thing as ex mode, when you're typing a command in commandline, you're directly typing into html input field. So these keypresses are not treated by us and the options we discuss here have no effect. If i'm wrong, then of course any maps or alwaysqwerty shouldn't apply there. What i'm saying is that in any shortcut modes such as normal, hint, gobble and whatever others there is no reason to have separate switches for each. So, when we're dealing with such shortcut modes(which i believe are all modes in which we intercept individual keypresses), it's reasonable to have just two boolean switches in config - whether to use alwasqwerty, and whether to apply keytranslatemap. So i'm basically suggesting to ditch keytranslatemodes map to simplify things. And if we ditch it, we can use alwasqwerty=true and usetranslatemap=false as defaults. for new users

Okay, i was wrong about direct input and now i see that minimalKeyFromKeyboardEvent is called even when typing. Then my idea is to manually apply alwaysqwerty and keymap in shortcut modes and don't apply them in typing modes, instead of exposing this in config through keytranslatemodes. Or we can just create another function that translates keyboard events literally without applying maps or using keycodes, and call this function from commandline_state handler and other places where we want literal keys.

treapster:
I wrote a lot of text, so to summaize:
1. We ditch keytranslatemodes and decide by ourserves whether to treat key literally or apply translations and use key codes.
2. When we're in 'shortcut' modes such as normal, hint gobble and similar, we always translate and use codes if config says so.
3. When we're in 'typing' modes such as commandline and input, there are two possibilities - either user enters characters, or uses shortcut with modifier such as Ctrl or Alt. These cases are easily distinguishible by a simple `if (key.hasNonShiftModifiers())`, and if it's true we apply map/use codes depending on config, while pass simple keys unmodified in case the key is simple

treapster:
How about discussing it in discord or something?

bovine3dom:
Input mode is just insert mode with some extra binds added for cycling between inputs.

I really would prefer we made it configurable by the end user as for which modes used `alwaysqwerty`. If you and I disagree that `hint` mode should use it then I'm sure lots of other people will disagree too. Do you think it would complicate the code a lot?

I'm not keen on treating modifier+key commands specially compared to all the other commands. I think some people use obscure keys like `¹` as leaders instead of using modifiers to execute commands.

---

You did write a lot of text :) I'm a little busy today but I might be able to take some time to read it thoroughly and think about it tomorrow. Happy to communicate on Discord if that's easier for you.

treapster:
> I really would prefer we made it configurable by the end user as for which modes used `alwaysqwerty`. If you and I disagree that `hint` mode should use it then I'm sure lots of other people will disagree too. Do you think it would complicate the code a lot?
> 
> I'm not keen on treating modifier+key commands specially compared to all the other commands. I think some people use obscure keys like `¹` as leaders instead of using modifiers to execute commands.

I agree that in cases when we can't reliably distinguish shortcuts from characters it makes sense to make it configurable, but in hint/gobble/normal it's pretty definite, isn't it? My point is to maximize the number of shortcuts which can work independent of layout, so aside from insert and ex mode, what stops us from using keycodes and keymap regardless of mode? I'd like more context on vimperator style hints, it tried them and didn't see much difference from simple. As of discord, what time is convenient for you? I think discussing by voice may be more productive:)

treapster:
Interesting: despite the fact that keypresses in ex mode are caught in minimalKeyFromKeyboardEvent and translated, the ex string is still obtained unmodified from input field like i said, so the current implementation does not affect command line even though it uses key codes unconditionally. I guess it will work the same for insert mode, so the main concern is hint, gobble and other shortcut modes.

treapster:
I only now noticed that exmode uses two listeners, one for shortcuts on keydown, and the other for exstring on input. Well, then there is no reason to distinguish between shortcuts and input, and thus we can safely apply translations and use keycodes to catch shortcuts in ex mode without fear of disturbing user input. The only problem is i can't make ex mode shortcuts such as `<C-k>` and `<A-k>` work even on current beta.

treapster:
> I only now noticed that exmode uses two listeners, one for shortcuts on keydown, and the other for exstring on input. Well, then there is no reason to distinguish between shortcuts and input, and thus we can safely apply translations and use keycodes to catch shortcuts in ex mode without fear of disturbing user input. The only problem is i can't make ex mode shortcuts such as `<C-k>` and `<A-k>` work even on current beta.

But things such as <A-f> and <A-b> work perfectly with alwaysqwerty! It allows you to do something like `:command *any text in any layout*<Shortcut>` without switching layouts, so i now don't see any reason to not use keycodes in exmode when alwaysqwerty = true, and the same goes for translation maps. Why complicate config and code with switches for every possible mode?

treapster:
So, here is the patch:
- Every key is translated once in the top of the pipeline
- Less code
- No config bloat

Let me express my understanding of the matter: if you use keymap, your goal is to make one keyboard layout appear as another to tridactyl in all shortcut-related aspects. Thus it's logical to make translation and interpretation in all shortcut modes. As for modes which use user input, as i said above the translation does not affect it, so we don't lose anything here, but allow proper shortcuts in all modes.

src/lib/config.ts
```diff
@@ -525,16 +536,20 @@ export class default_config {
...
+    /**
+     * Instead of fetching actual character which depends on selected layout,
+     * use machine code of a key. Makes any keyboard work as if it is qwerty.
+     * Recommended for everyone, especially multilingual.
+     *
+     * If you don't want qwerty, you can use keytranslatemap to map it to something else,
+     * and this setting is still useful because it makes all events tied to the physical
+     * keys.
+     */
+    alwaysqwerty: "true" | "false" = "true"
```
bovine3dom, line 552: I'm really not sure this should be true by default, what do other applications do?

If you've never seen a QWERTY keyboard in your life it must be pretty weird to be told to press the `w` key to open a new window and then find that your `w` key doesn't work

src/lib/keyseq.ts
```diff
@@ -484,25 +478,111 @@ function numericPrefixToExstrSuffix(numericPrefix: MinimalKey[]) {
...
+    Comma: ",",
+    Semicolon: ";",
+    Slash: "/",
+}
+
 /**
- * Translates the given set of keyEvents (in place) as specified by
- * the given key translation map. All keys *and* values in the key
- * translation map must be length-1 strings.
+ * Convert key to key as if shift is pressed
```
bovine3dom, line 578:
```suggestion
 * Convert key to key as if shift is pressed on a US QWERTY layout
```

src/lib/keyseq.ts
```diff
@@ -484,25 +478,111 @@ function numericPrefixToExstrSuffix(numericPrefix: MinimalKey[]) {
...
+    "]": "}",
+    "\\": "|",
+    ";": ":",
+    "'": '"',
+    ",": "<",
+    ".": ">",
+    "/": "?",
+}
+
+const keycodetranslatemap = {
```
bovine3dom, line 531: We should link to the page they came from (I imagine https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values#code_values_on_linux_x11 ?) and triple check to make sure all pressable keys are present.

We're missing `Quote` for example which lots of layouts change :)

src/lib/config.ts
```diff
@@ -525,16 +536,20 @@ export class default_config {
...
-        inputmaps: "false",
-        ignoremaps: "false",
-        exmaps: "false",
-        hintmaps: "false",
-    }
+    usekeytranslatemap: "true" | "false" = "false"
+
+    /**
+     * Instead of fetching actual character which depends on selected layout,
+     * use machine code of a key. Makes any keyboard work as if it is qwerty.
```
bovine3dom, line 545:
```suggestion
     * use machine code of a key. Makes any keyboard work as if it is using a US QWERTY layout for all Tridactyl modes except free text input
```

src/lib/config.ts
```diff
@@ -525,16 +536,20 @@ export class default_config {
     }
 
     /**
-     * Whether to use the keytranslatemap in various maps.
+     * Whether to use the keytranslatemap
```
bovine3dom, line 539:
```suggestion
     * Whether to use the keytranslatemap. For example, useful for changing the effective base layout when using [[alwaysqwerty]].
     * 
     * To make your `alwaysqwerty` layout more like a GB layout, run `:keymap @ "` and then no matter which layout you are using, pressing the `"` key on a physical GB layout will trigger `"` binds rather than `@` binds.
     * 
     * As a rule of thumb: `:keymap [equivalent US key] [key you want to press on your physical layout]`
```

I finally understand the point of using `keymap` and `alwaysqwerty` at the same time now! This help message isn't perfect but I think it's at least better :)

bovine3dom (changes requested):
This looks great except for a few missing keys (see comments), but I still have problems with `alwaysqwerty` being used in hint mode.

1. `:set hintfiltermode vimperator`
2. `:set hintchars 1234567890`
3. go to some website, e.g. https://www2.hm.com/fr_fr/index.html
4.  `setxkbmap fr bepo` in your terminal
5. press the physical f key on your keyboard
6. say you want to visit the `Femme` page. now you have to type out the French word `femme`, which because it is in my muscle memory and I'm using a French layout my brain wants to type the physical keys `/f''f` which doesn't work. I need to type `epqqp` in the layout instead to ensure that I'm hitting the correct physical keys. For me, hint filtering is like a kind of text input, because I'm typing words.

---

If we can't see a solution for hint mode I'm happy to just add a disclaimer to `alwaysqwerty` since I think vimperator-filter users are a small minority and it is much less confusing to type random characters in QWERTY than it is whole words :)

bovine3dom:
@dydascal @salehmu @drogeek you all contributed `keymap` settings for Tridactyl here - https://github.com/tridactyl/tridactyl/wiki/Internationalisation . We're trying to remove the need to do that by having a setting that lets Tridactyl translate the keys automatically.

We are thinking of removing per-mode settings for keymap and instead just having it apply to all modes to simplify the code. Could you offer us any reasons for having it apply to some modes but not others?

Thanks and sorry for the unsolicited pings :)

treapster:
> This looks great except for a few missing keys (see comments), but I still have problems with `alwaysqwerty` being used in hint mode.
> 
>     1. `:set hintfiltermode vimperator`
> 
>     2. `:set hintchars 1234567890`
> 
>     3. go to some website, e.g. https://www2.hm.com/fr_fr/index.html
> 
>     4. `setxkbmap fr bepo` in your terminal
> 
>     5. press the physical f key on your keyboard
> 
>     6. say you want to visit the `Femme` page. now you have to type out the French word `femme`, which because it is in my muscle memory and I'm using a French layout my brain wants to type the physical keys `/f''f` which doesn't work. I need to type `epqqp` in the layout instead to ensure that I'm hitting the correct physical keys. For me, hint filtering is like a kind of text input, because I'm typing words.
> 
> 
> If we can't see a solution for hint mode I'm happy to just add a disclaimer to `alwaysqwerty` since I think vimperator-filter users are a small minority and it is much less confusing to type random characters in QWERTY than it is whole words :)

Well, what you're describing is not exclusive to vimperator-style hint mode. Vanilla alwaysqwerty will work like qwerty in all shortcuts mode. Regarding vimperator-style hints, if this mode expects text input, then we should provide it with original characters - because the content can be in any layout and we're typing words. But checking for whether we're in this partucular hint mode inside minimalKeyFromKeyboardEvent will be too cumbersome. Another solution may be adding .character field to MinimalKey which will be copied from original KeyboardEvent.key and passed down the pipeline. This way we can use it in hint mode when we're using vimperator-style hints, but other parsers will only use translated key. But the problem is that even in current version vimperator-style hints are ambiguous since user input can be interpreted either as a letter in a word or a letter in a hint, at least i's how i observed it after your instructions. Since alwaysqwerty introduces separation between typing and shortcuts, the only way i see it working with alwaysqwerty is restricting this mode to just characters, and not generating any hints, or at least not comparing input with hints.

For usual hint mode and other shortcuts, you can create map from qwerty to whatever layout you want and have it permanently. I think for convenience we can even ship the most popular maps like azerty, dvorak and couple others, and probably have a command like `:setlayoutmap azerty` which will assign shipped objects to kaytranslatemap.
Example how shipped maps can look in config(and you can grab azerty mapping):
<details>

```
    layoutMaps = {
        azerty: {
            "q": "a",
            "w": "z",
            "e": "e",
            "r": "r",
            "t": "t",
            "y": "y",
            "u": "u",
            "i": "i",
            "o": "o",
            "p": "p",
            "[": undefined, // dead keys
            "]": "$",
            "\\": "*",
            "a": "q",
            "s": "s",
            "d": "d",
            "f": "f",
            "g": "g",
            "h": "h",
            "j": "j",
            "k": "k",
            "l": "l",
            ";": "m",
            "'": "ù",
            "z": "w",
            "x": "x",
            "c": "c",
            "v": "v",
            "b": "b",
            "n": "n",
            "m": ",",
            ",": ";",
            ".": ":",
            "/": "!",
            "`": "²",
            "1": "&",
            "2": "é",
            "3": "\"",
            "4": "'",
            "5": "(",
            "6": "-",
            "7": "è",
            "8": "_",
            "9": "ç",
            "0": "à",
            "-": ")",
            "=": "=",
            "Q": "A",
            "W": "Z",
            "E": "E",
            "R": "R",
            "T": "T",
            "Y": "Y",
            "U": "U",
            "I": "I",
            "O": "O",
            "P": "P",
            "{": undefined, // dead
            "}": "£",
            "|": "µ",
            "A": "Q",
            "S": "S",
            "D": "D",
            "F": "F",
            "G": "G",
            "H": "H",
            "J": "J",
            "K": "K",
            "L": "L",
            ":": "M",
            "\"": "%",
            "Z": "W",
            "X": "X",
            "C": "C",
            "V": "V",
            "B": "B",
            "N": "N",
            "M": "?",
            "<": ".",
            ">": "/",
            "?": "§",
            "~": "²",
            "!": "1",
            "@": "2",
            "#": "3",
            "$": "4",
            "%": "5",
            "^": "6",
            "&": "7",
            "*": "8",
            "(": "9",
            ")": "0",
            "_": "°",
            "+": "+",
        },
        dvorak: {
            /*...*/
        },
    }
```
</details>

However, on the other hand, this won't work out of the box, and there are probably dozens of qwerty variations in different european countries, and i suspect we can't cover them all, which will force users to do all those mapping manually which is no fun.  So i have another hacky idea: introduce third hybrid mode in which we check whether a character is ascii, and in case it is treat it literally, otherwise use keycode. It will roughly resemble how vscode works, but may behave unexpected with accented characters in qwerty-derived layouts, and we also assume that people with non-latin layouts are expecting qwerty for shortcuts, which i think true in 95% of cases.

src/lib/config.ts
```diff
@@ -525,16 +536,20 @@ export class default_config {
...
      */
-    keytranslatemodes: { [key: string]: "true" | "false" } = {
-        nmaps: "true",
-        imaps: "false",
-        inputmaps: "false",
-        ignoremaps: "false",
-        exmaps: "false",
-        hintmaps: "false",
-    }
+    usekeytranslatemap: "true" | "false" = "false"
```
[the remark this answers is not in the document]
treapster: 1. I'd like to hear feedback from users too.  
2. After we reach agreement on changes to config itself, we'll add config updater.  
3. Agree on usekeytranslatemap = true by default. But the switch itself may be useful because someone probably wants to disable their map but not lose it?  
4. alwaysqwerty  and usekeytranslatemap are not commands, and i'm not sure there are better names for them. But i18n namespace is a good idea. Or it can be `input` or `keyboard`, i don't mind.

src/lib/config.ts
treapster: [Here](https://github.com/tridactyl/tridactyl/pull/4439#issuecomment-1315772177) i propose hybrid mode idea.

src/lib/keyseq.ts
treapster: Sure, i missed `Quote`, brackets and minus. Link may be useful too, even though i generated this map just by doing  
``console.log(`{e.code}: "{e.key}",`)``on en layout and a bit of python for letters and digits.

src/excmds.ts
```diff
@@ -4251,14 +4251,34 @@ export async function bind(...args: string[]) {
             await browser.commands.update({ name: command.name, shortcut: minimalKeyToMozMap(mapstrToKeyseq(args_obj.key)[0]) })
             await commandsHelper.updateListener()
         }
-        p = config.set(args_obj.configName, args_obj.key, args_obj.excmd)
+        p = config.set(args_obj.configName, args_obj.key, args_obj.excmd).then(() => fillcmdline_nofocus(`${args_obj.key} is bound to ${args_obj.excmd}` + ` in ${args_obj.configName}`))
```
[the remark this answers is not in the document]
treapster: Ok, then we need to find another way to provide feedback after :bindkey

src/excmds.ts
```diff
@@ -4251,14 +4251,34 @@ export async function bind(...args: string[]) {
...
+
+     If you have alwaysqwerty enabled, it will bind commands to physical keys.
+
+     Used same as bind, but without key provided:
+
+         - `bindkey [command]`, then press the keys you want to bind, then hit Enter.
+         - `bindkey --mode=[mode] [command]` also works.
+
+     You can execute it without arguments to see what is bound to the keys you type.
+*/
```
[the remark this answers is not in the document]
treapster: Well, currently it does not prefill but executes bind right after you hit enter. And this is why i added output to bind, since otherwise you can't tell what keys have you bound. Do you think it's better to prefill?

treapster:
After thinking a bit more about hybrid ascii-thing, i think it's not that good, because punctuation is ascii, but varies across layouts, so in hybrid mode even in a combination such as qwerty+russian all keys which are not letters will change behavior when switching layout. Thus i think we're better off disabling alwaysqwerty by default and providing maps for the layouts we're not too lazy to provide.

treapster:
> After thinking a bit more about hybrid ascii-thing, i think it's not that good, because punctuation is ascii, but varies across layouts, so in hybrid mode even in a combination such as qwerty+russian all keys which are not letters will change behavior when switching layout. Thus i think we're better off disabling alwaysqwerty by default and providing maps for the layouts we're not too lazy to provide.

On the other hand, most shortcuts are just letters, and most keys in qwerty variations are ascii, so there is a chance that hybrid node will provide the best average experience out of the box without any config changes, but it can quickly become a nightmare if combined with keytranslatemap

treapster:
@bovine3dom, do you have time to look into this?

bovine3dom:
I'll try to find some time this week, sorry that it's taking so long. Thanks again for all the work you're putting into it - it is a feature I have wanted personally in Tridactyl for a long time :)

treapster:
> I'll try to find some time this week, sorry that it's taking so long. Thanks again for all the work you're putting into it - it is a feature I have wanted personally in Tridactyl for a long time :)

Thanks for the kind words! Take all the time you need:)

bovine3dom:
I think actually I'm just going to test this and merge it as is, bumping the priority of #3635 up to a P3. That way we should be able to have `alwaysqwerty` be false in hint mode for free and there shouldn't be any ugliness with trying to check the mode inside the key handling code.

treapster:
I think in current implementation of :bindkey feedback will greatly improve UX, how about -f flag to bind which will make it produce feedback and will be used from bindkey?

bovine3dom:
Oh, I had forgotten that `:bindkey` could accept arguments, I was always using it without any and then finishing off the command myself. Do you think allowing it to accept arguments adds any value?

`:bindkey ---mode=[e.g. insert]` with no other arguments seems to forget the mode for what it is worth.

---

I'm trying to work out if it makes any sense to leave `keymap` in if we are allowing the `keylayoutforce` layout to be changed. `keymap` is simpler because you don't have to work out the internal name for the key...

treapster:
> Oh, I had forgotten that `:bindkey` could accept arguments, I was always using it without any and then finishing off the command myself. Do you think allowing it to accept arguments adds any value?

I think it's logical to accept arguments to keep it a fully-functioning wrapper around bind. And what do you mean by finishing off by yourself? Bindkey executes bind as soon as you hit enter, and that is why we accept arguments - to forward them.

> `:bindkey ---mode=[e.g. insert]` with no other arguments seems to forget the mode for what it is worth.

Well, it's not bindkey, it's bind itself forgetting the mode:). If you do e.g `:bind --mode=visual j`, you'll get `bind j js document.getSelection().modify("extend","forward","line")` - with no mention of visual in it. And you get the same result for `:bindkey --mode=visual<CR><j><CR>`.

> I'm trying to work out if it makes any sense to leave `keymap` in if we are allowing the `keylayoutforce` layout to be changed.

Well, when keylayoutforce is enabled it certainly does not make sense to use keymap, and i think it should either fail with error or ignore and warn when called with keylayoutforce set. As of leaving keymap in at all - well, to ditch it we also have to ditch keytranslatemap, and thus break setups of all it's users, and force them to recreate their keymaps with keylayoutforcemapping. Instead of doing that, we can mark it as deprecated and unrecommended in documentation/wiki and promote keylayoutforcemapping, and probably remove it a couple releases after.

>`keymap` is simpler because you don't have to work out the internal name for the key...

It mostly affects those unfamiliar with qwerty(since names themselves are either "KeyX" or their obvious names found on mozilla website), and it can be solved by:
 - Providing built-in keyboard layouts so that user can simply `:setkeyboardlayout azerty` which will assign requested layout map to keylayoutforcemapping
 - Providing command analogous to bindkey which calls `:set keylayoutforcemapping.{scanned code} {user-provided values}`

bovine3dom:
Ok, aside from documentation (which I am happy to write) I think I'm happy with this now.

1) I renamed it to bindwizard to emphasise that it is interactive and isn't just another name for bind

2) I removed the automatic invocation of the `:bind` command since I can see lots of other reasons for using it - people will want to use it for bindurl, unbind, or maybe even just trying to understand our key sequence notation. And that also made it easier to give people feedback since the command displays by default

---

I'll probably add a page to tutor, a note to the new tab page / FAQ and some extra text in [[bind]].

Are you happy with me to merge this aside from that?

treapster:
Now that there's no alwaysqwerty mentions left, i'm also happy with it. Extra notes on new page and tutor will be very useful.

I have one question regarding merging itself - why aren't you doing squash+merge or at least rebase+merge? Right now all the commits from MRs get intermixed with others after merge, and most of them are some unimportant cosmetic fixes. This makes commit history noisy and hard to follow. Since most MRs are rather atomic changes related to one feature, how about auto squashing on merge? And if you want to preseve commits of some big MR, i think it's better to rebase before merge so that all related commits are consecutive.

bovine3dom:
I don't like squash and merge because I like git blame / git bisect to have useful information for each line, usually at a more granular level than pull requests. If something is just formatting it usually isn't hard to just reblame again. And the fewer lines in each commit, the more precisely git bisect can pinpoint the cause of a bug.

If I was less lazy I would rebase and merge, rewriting the history to be as useful as possible and trying to remove any commits that led to broken builds. But that would take a lot of time and I'm not sure it would be worth it.

The benefit I can see of squashing and merging is simplifying making changelogs, but it doesn't take me that long to make them already.

Why do you care that the commit history is hard to follow? What do you use it for? :)

treapster:
Not that i read commit histories in my free time, but it matters. There is even a [specification](https://www.conventionalcommits.org/en/v1.0.0/) for commits. And when the history is well-structured, it may be interesting to read to get familiar with the latest changes. If you don't wanna squash, rebasing can still be useful to keep things grouped, and it's just one command.

The other problem when you don't rebase is that history becomes non-linear. For example, now the history of this branch looks like some of commits from this PR are included in release 1.23 since they come before, but it is not true. With linear history if you checkout A you get everything before A, which is simple and predictable.

treapster:
Gentle ping. Rebase or not rebase, let's merge it.

bovine3dom:
Sorry, I actually worked on the docs for this while travelling for Christmas, but now I'm full on holiday mode for the next couple of days, can't really touch any code directly for a while

But thanks for the ping :) In most other circumstances I would have genuinely forgotten...

treapster:
No problem, then happy holidays:) And thanks for your time and review!

treapster:
Just for the record and probably next PR, i was procrastinating and created [this](https://gist.github.com/treapster/324119c773fac31651f6422087b36804#file-getmap-html) stupid tool ([gistpreview](https://gistpreview.github.io/?324119c773fac31651f6422087b36804)) for creating langmaps, and now have azerty, german and dvorak maps built with this, which i think we can ship with tridactyl and a new command like `:setkeyboardlayout`. Also, we can change it to create keylayoutforcemapping commands on the fly, integrate it into tridactyl and automate the mapping process entirely, by basically scanning user's layout until he presses enter or something. However, providing main European layouts will likely be enough:)

bovine3dom:
That sounds great, the little gist works well :)

bovine3dom:
We should probably add a setting of like "base layout" that's just e.g. en-us, en-gb etc and then merge the user settings on top of that rather than having the ugliness of some Tridactyl default settings copied across into user config. That way we can fix the bugs that we will inevitably introduce.

treapster:
> We should probably add a setting of like "base layout" that's just e.g. en-us, en-gb etc and then merge the user settings on top of that rather than having the ugliness of some Tridactyl default settings copied across into user config. That way we can fix the bugs that we will inevitably introduce.

I don't fully understand what you mean here. If you mean assigning map values per key instead of assigning the map itself, it's interesting. However i don't see how it will help in case we ship wrong layout map -- the user will have to fix it manually anyway. And if we prioritize existing values in keylayoutforce, we won't be able to change them

>Edit: I forgot that I was supposed to make a decision about rebasing/squashing ... the absence of that was negligence rather than actual thought :)

Then you can give it a thought on the next MR:)
 
Also, we as soon as beta comes out, we should edit [wiki](https://github.com/tridactyl/tridactyl/wiki/Internationalisation) to something like this:

--------

# Make tridactyl always behave like you're using qwerty

`set keylayoutforce true` (since \*next beta release\*)

# Before \*next beta release\*
\*all the legacy stuff with headers one step smaller\*

bovine3dom:
Does some pseudocode help?

```js
const layout = R.mergeLeft(config.get("keylayoutforcemapping"),KEYLAYOUT_DEFAULTS[config.get("keylayoutforcebaselayout")])
```

No promises on whether I used mergeLeft correctly... I always have to look it up

Wiki is a good idea :)

treapster:
> Does some pseudocode help?
> 
> ```js
> const layout = R.mergeLeft(config.get("keylayoutforcemapping"),KEYLAYOUT_DEFAULTS[config.get("keylayoutforcebaselayout")])
> ```
> 
> No promises on whether I used mergeLeft correctly... I always have to look it up

Not that much, since in your example user settings which is usually qwerty will override anything on the right, making it no-op. And i assume on the right we'll have maps for some non-qwerty layouts. But i think we can continue this discussion when i create the corresponding MR(i'm not sure when).

bovine3dom:
Yeah, we'd delete all the current mappings and move them into e.g. `KEYLAYOUT_DEFAULTS.en-us`, so the only thing left there would be used overrides.

It would make how to override individual keys less clear (like knowing that the 1 key is `Digit1`) but I think almost no-one would want to do that compared to changing the base layout, so it would be a net win.

PR #4440, opened 2022-10-30
Use MinimalKey internally, add terminator key to gobble mode

treapster:
A patch which grew out of [this](https://github.com/tridactyl/tridactyl/pull/4439) PR. The main goal is to allow gobbling with terminator and also register things like `<C-x>` in gobble mode. It's easier to implement on concrete type than on KeyEventLike, so i changed types to MinimalKey. Because of that tests will likely fail, so i expect to spend some time fixing before it's really mergeble. But it'll simplify merging of original PR.

### DIFF
--- a/src/content/controller_content.ts
+++ b/src/content/controller_content.ts
@@ -5,7 +5,7 @@ import * as controller from "@src/lib/controller"
 import {
     KeyEventLike,
     ParserResponse,
-    minimalKeyFromKeyboardEventLike,
+    minimalKeyFromKeyboardEvent,
     MinimalKey,
 } from "@src/lib/keyseq"
 import { deepestShadowRoot } from "@src/lib/dom"
@@ -138,7 +138,7 @@ function* ParserController() {
                     // binding, so it can't grow indefinitely unless you
                     // have a combination of maps that permits bindings of
                     // unbounded length.
-                    keyEvents.push(minimalKeyFromKeyboardEventLike(keyevent))
+                    keyEvents.push(minimalKeyFromKeyboardEvent(keyevent))
                 } else {
                     keyEvents.push(keyevent)
                 }
--- a/src/content/controller_content.ts
+++ b/src/content/controller_content.ts
@@ -2,7 +2,12 @@ import { isTextEditable } from "@src/lib/dom"
 import { contentState, ModeName } from "@src/content/state_content"
 import Logger from "@src/lib/logging"
 import * as controller from "@src/lib/controller"
-import { KeyEventLike, ParserResponse } from "@src/lib/keyseq"
+import {
+    KeyEventLike,
+    ParserResponse,
+    minimalKeyFromKeyboardEvent,
+    MinimalKey,
+} from "@src/lib/keyseq"
 import { deepestShadowRoot } from "@src/lib/dom"
 
 import * as hinting from "@src/content/hinting"
@@ -97,7 +102,7 @@ export const canceller = new KeyCanceller()
 /** Accepts keyevents, resolves them to maps, maps to exstrs, executes exstrs */
 function* ParserController() {
     const parsers: {
-        [mode_name in ModeName]: (keys: KeyEventLike[]) => ParserResponse
+        [mode_name in ModeName]: (keys: MinimalKey[]) => ParserResponse
     } = {
         normal: keys => generic.parser("nmaps", keys),
         insert: keys => generic.parser("imaps", keys),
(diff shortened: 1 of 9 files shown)


src/parsers/gobblemode.ts
```diff
@@ -27,18 +30,29 @@ function reset() {
 }
 
 /** Receive keypress. If applicable, execute a command. */
-export function parser(keys: KeyEventLike[]) {
+export function parser(keys: MinimalKey[]) {
```
bovine3dom, line 33: wouldn't this PR be much shorter if we wrote/used a KeyEventLikeToMinimalKey function here? Do we gain a lot by changing it everywhere?

treapster:
There is one thing i'm not sure about: in this patch gobble interprets shifted keys as <S-key>. I saw comments that we expect printable keys to be a single letter keys, but if inside <S-key> key is capitalized(whick it is, until caps lock is on), it should work. Or should we bother and treat all printable keys as if there is no shift?

src/parsers/gobblemode.ts
treapster: Why use KeyEventLikeToMinimalKey everywhere when we can use it once or twice in the top of a pipeline and work with internal type everywhere else? It eliminates checking for InstanceOf and allows direct usage of MinimalKey's methods and helper functions

treapster: Also, why it only treats the first key? Is it guaranteed it always gets no more than one?

treapster:
Since Windows test also fails on [other](https://github.com/tridactyl/tridactyl/pull/4442) PRs, i guess I didn't break anything here?

bovine3dom:
> Since Windows test also fails on https://github.com/tridactyl/tridactyl/pull/4442 PRs, i guess I didn't break anything here?

Yeah, the windows esr build is broken at the moment, it's nothing to worry about :)

src/parsers/gobblemode.ts
bovine3dom: > Also, why it only treats the first key? Is it guaranteed it always gets no more than one?

Yeah, it only ever gets one key.

controller_content works like this:

```
1. initialise an empty list of key events
2. receive a key event, add it to that list
3. send the list of key events to a parser and get a new list of key events back - go to 1. Or get an ex string to execute - go to 2.
```

gobble mode only ever returns and empty list of key events https://github.com/tridactyl/tridactyl/blob/74d357e1e2bdbc31ce87d06bfc5ac2ec3843e888/src/parsers/gobblemode.ts#L57 so you can only ever get one key.

Parsing multiple keys at once is so that e.g. the normal mode parser can say "yes the g key matches loads of binds, give me it again with the next key and we'll see if it matches a bind".

src/content/controller_content.ts
```diff
@@ -153,17 +168,10 @@ function* ParserController() {
 
                 const newMode = contentState.mode
                 if (newMode !== currentMode) {
-                    keyEvents = []
+                    keyEvents = keyEvents.slice(-1)
```
bovine3dom, line 171: Could you explain what you're trying to do in this line?

treapster: Before it was `check mode -> clear events if it changed-> push new event`, now i grouped KeyboardEvent-related things above including push, and thus instead of clearing we leave the one we pushed

src/lib/keyseq.ts
```diff
@@ -473,37 +475,44 @@ function numericPrefixToExstrSuffix(numericPrefix: KeyEventLike[]) {
...
+            keyEvents[index].translated = true
         }
     }
 }
+/**
+ * Convert keyboardEvent to internal type MinimalKey
+ * for further use. Key is obtained through layout-independent
+ * code if config says so.
+ */
+export function minimalKeyFromKeyboardEventLike(keyEvent: {
```
treapster, line 502: Why not accept keyboardEvent? We need event.code and in future we'll probably want more context from the actual event

src/lib/keyseq.ts
```diff
@@ -473,37 +475,44 @@ function numericPrefixToExstrSuffix(numericPrefix: KeyEventLike[]) {
...
-            // through, so we just swap the key event out for a new
-            // MinimalKey with the right key and modifiers copied from
-            // the original.
-            keyEvents[index] = new MinimalKey(newkey, {
-                altKey: keyEvent.altKey,
-                ctrlKey: keyEvent.ctrlKey,
-                metaKey: keyEvent.metaKey,
-                shiftKey: keyEvent.shiftKey,
+        if (!keyEvent.translated && newkey !== undefined) {
+            keyEvents[index] = minimalKeyFromKeyboardEventLike({
```
treapster, line 489: minimalKeyFromKeyboardEvent was supposed to convert from keyboardEvemt to MinimalKey, using code if configured, once per event. I don't think it's a good idea to make it general and use here, if we want to be more concise we should either use a separate helper to update or make .key mutable.

src/lib/keyseq.ts
bovine3dom: it was so it could be used on line 489. KeyboardEvent has lots and lots of properties that aren't used in the function, I didn't see why it should be restricted :)

src/lib/keyseq.ts
bovine3dom: I don't see the problem. Making key mutable seems dangerous since lots of other things might need to change when the key changes.

This way if we make a new requirement like needing a keycode we'll notice that it is missing from translated keys.

And it just seemed neater.

treapster: My idea is that we don't need code in internal representation of keys, but we need it to convert to this representation once in the top of the pipeline. We convert either trivially or (after second PR) using code, and then use .key property same as before and apply translations. It doesn't make sense to use conversion function twice, we better create someting like updateKey(), or make .key private and create setter.

treapster:
We of course can make conversion function accept opaque object or union type, but why, if we only ever give it real keyboard event, and there is no reason to give it anything else?

bovine3dom:
LGTM, I'll merge once it passes the tests :)

treapster:
Do you have an idea why they broke?

PR #4453, opened 2022-11-05
Option to exclude selectors from hint result

GHolk:
Sometimes, a real exclude operation after `-c` selector is really useful. It is labour to construct a selector to match anything you want to click except some element.

Another thing I am interested in is to add a convient interface in hint mode for javascript to filter elements.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -432,8 +432,7 @@ export function hintElements(elements: Element[], option = {}) {
                 const first = await Promise.race([onSelect.promise, endPromise])
                 if (first && typeof first === "object" && key in first) {
                     yield first[key]
-                }
-                else return await endPromise
+                } else return await endPromise
             }
         }
         const result = wrap()
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -397,7 +397,8 @@ interface Hintables {
 }
 
 /** A convient javascript interface to hint on specified html elements.
- *  The return value is a promise resolve to the selected element.
+ *  The return value is a promise resolve to the selected element,
+ *  or a AsyncIterator resolve to the selected elements in rapid mode.
  */
 export function hintElements(elements: Elements[], rapid = false) {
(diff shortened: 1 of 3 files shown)


GHolk:
I don't understand why linter complain. I think `src/content/hinting.ts` line 315 is fine.

bovine3dom:
it was just a minor thing, it always wants `else` to be on the same line as the closing brace of the `if`

src/content/hinting.ts
```diff
@@ -396,6 +396,64 @@ interface Hintables {
...
+  @param elements a iterator yield html elements
+  @param option a option object. The `option.rapid` specify whether hint in rapid mode. Default value is false. The `option.callback` is execute when a hint is selected if specified.
+
+  @returns promise resolve to the selected element, or a async iterator in rapid mode.
+ */
+export function hintElements(elements: Element[], option = {}) {
+    const hintable = toHintablesArray(Array.from(elements))
+    const rapid = option["rapid"] ?? false
+    const callback = typeof option["callback"] === "function" ?
+        option["callback"] : x => x
```
bovine3dom, line 412: Shouldn't we provide a more useful default like trying to click on the element? Otherwise LGTM :)

GHolk: If we introduce a default action, the async iterator will be useless without specifying a callback do nothing.

pros:

* default action may be useful.
* Async iterator code is complex; callback will be easier.

cons:

* Callback is ugly; promise/async is beautiful.

How about treat `{callback: null}` as `NOP` and `{callback: undefined}` as `e => e.click()`?

bovine3dom: Oh, I see. I hadn't thought about that.

Maybe we should add some short examples to the docstring instead of a useful default, like `hintElements(...).then(e=>tri.dom.simulateClick())` and `for (await e of hintElements(..., {rapid: true}){tri.dom.simulateClick(e)}` ?

bovine3dom:
LGTM, sorry it took so long to merge!

Issue #1856, opened 2019-09-16
Function/cmd to add a bookmark to a specyfic folder with a current page title


frifix:
Tridactyl version: 1.16.3pre3122-c3f45f5
Firefox version: Mozilla Firefox 69.0
Operating system: win

I want to bind a function/cmd that adds a bookmark to a specyfic folder, but I want to preserve a current page title. Something that behaves like `bmark` command when we skip an URL, but with a specyfic folder. This doesn't work properly, because it doesn't add a bookmark to a specyfic folder:
`bind ,a run_exstr bmark /Bookmarks Menu/readlater/`

I would be grateful if anyone would help me.

bovine3dom:
`run_exstr` isn't required - you just need `bind ,a bmark ...`.

There are two options: submit a PR that adds a -f flag to bmarks telling it to add bmark to current URL in given folder with current title.

Other option is to make your own with JavaScript. I don't have time to write it out for you but the gist is this: `jsb /*set current_url and current_title... */ tri.excmds.bmark(current_url,"/folder/you/want/"+current_title_with_slashes_removed)`

You'll probably need to do some hunting in the code to get the current URL and title but I hope that puts you on the right track.

frifix:
Thank you for your tips. I came with this solution:
`bind ,a composite js "bmark " + window.location.href + " /Bookmarks Menu/readlater/" + window.document.title.replace(/\//g, '\\') | composite`
I suppose that there is a more simplier solution, but this works for me. As you pointed out I had to remove all slashes (I replace them with backslashes).

A solution with -f flag would be the most approriate. Feel free to close this issue if you feel that -f flag is not worth the effort to write by some contributor.

forgetmenotz:
I'm not sure if this would be of help to you but what I use is a firefox addon called QuickBookmarkToFolder. Ctrl+Shift+k opens a dialog box, it fuzzy finds one of your folders and you just select with arrow keys and press enter to add the page as current title. I have a key bound to Mode_switch in xmodmap that changes vim keys to arrow keys so it works well for me. Would be better if it allowed folder creation, not sure if that is possible to do from tridactyl.

bovine3dom:
Fixed in #4457

PR #4457, opened 2022-11-09
Use tab title if bookmarking current page

badcold:
There's currently no way to add a bookmark to a folder AND keep the title.

Most of the feature was implemented via PR https://github.com/tridactyl/tridactyl/pull/862/

Should fix issue https://github.com/tridactyl/tridactyl/issues/1856


bovine3dom:
Looks great, thanks! I'll merge once the tests have passed

PR #4458, opened 2022-11-09
Make E2E tests more helpful for PRs

bovine3dom:
PRs are more likely to have genuine failures than commits to master, so retry them less often and try all the tests even if some fail


bovine3dom:
Well 1 retry clearly wasn't enough. Also, it doesn't look like attempting to unset the fail-fast is working on the CI but it worked fine on my machine. :(

PR #4460, opened 2022-11-11
[Tutor] Fix keybinding on hint mode

ericliou:
`<Enter>` key is not mapped to exit hint mode, at least not anymore. It's now mapped to `hint.selectFocusedHint`.
https://github.com/tridactyl/tridactyl/blob/44efcf38af9290a6bc8061c59119b0ce24b52756/src/lib/config.ts#L411


bovine3dom:
You're right, it must be a typo. Thanks!

PR #4467, opened 2022-11-17
 Do not hide ex cmdline after fillcmdline_tmp if it is focus

GHolk:
Sometimes the cmdline is closed before I start editing it.
Determine with focus state should be more appropriate than determine with the modification,
or are there any reason to do so?

I know check a hard-code id is a bad idea, but I can not figure out other ways to check
whether the cmdline is focused; please help.
Should I add a new exported function in `src/commandline_frame.ts` to check focus or blur?
Any better solutions?

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3843,7 +3843,7 @@ export async function fillcmdline_tmp(ms: number, ...strarr: string[]) {
     Messaging.messageOwnTab("commandline_frame", "fillcmdline", [strarr.join(" "), false, false])
     return new Promise<void>(resolve =>
         setTimeout(async () => {
-            if ((await Messaging.messageOwnTab("commandline_frame", "getContent", [])) === str) {
+            if (document.activeElement?.id !== "cmdline_iframe") {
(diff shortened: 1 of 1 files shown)


bovine3dom:
Sorry that it has taken me over a year to review a one-line PR ... but if instead we modify the hide_and_blur function here https://github.com/tridactyl/tridactyl/blob/2c0798ffe5bc757057177ec95056831d237886f7/src/content/commandline_content.ts#L120

to accept an argument `force=true`, it should be as simple as adding a guard `if (!(force || (document.activeElement !== cmdline_iframe)) { return }` at the top of that function. Right?

GHolk:
To be honest, I don't like the idea to change the behavior of `ex.hide_and_clear`, because when I call `hide_and_clear`,
it mean that I want it be hidden and cleared.

Maybe rename the `force` to a `smart` option?
If the smart option presents, it will check the active element before close.

Anyway, I still prefer to modify the behavior of the `fillcmdline_tmp`.

Issue #2944, opened 2020-10-23
Script not running with `composite hint -y | exclaim /myscript`


mrhieu059:
Tridactyl version: 1.20.2
Firefox version: Mozilla Firefox 82.0
Operating system: FreeBSD

I had a script which which pass _url_ to youtube-dl. The script is work perfectly fine with this command:` composite get_current_url | exclaim /myscript` however, when I use this command `composite hint -y | exclaim /myscript` to pass a specific url onto myscript, it doesn't work and doesn't seem to activate the script.

bovine3dom:
You need to use `hint -pipe a href` instead of `hint -y` : )

bovine3dom:
Oh, if you're piping to exclaim it's a good idea to pipe through `:shellescape`. e.g. `composite hint ... | shellescape | exclaim ...`. Without it a site could execute whatever it liked on your machine if you clicked a malicious link which targeted users who weren't using shellescape.

mrhieu059:
Thank you @bovine3dom, it worked when I pipe through `shellescape`.

PR #4481, opened 2022-11-28
Docs: add notice about `shellescape` usage in `exclaim` help.

pataquets:
Found it out at #2944 while troubleshooting a similar problem.
I think it's worth it to add to docs, as it might help others avoid some trouble.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -692,7 +692,7 @@ export async function nativeopen(...args: string[]) {
  *
  * Requires the native messenger, obviously.
  *
- * If you're using `exclaim` with arguments coming from a pipe, consider using `shellescape` to properly escape arguments and to prevent unsafe commands.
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -692,6 +692,8 @@ export async function nativeopen(...args: string[]) {
  *
  * Requires the native messenger, obviously.
  *
+ * If you're using `exclaim` with arguments coming from a pipe, consider using `shellescape` to properly escape arguments and to prevent unsafe commands.
```
bovine3dom, line 695:
```suggestion
 * If you're using `exclaim` with arguments coming from a pipe, consider using [[shellescape]] to properly escape arguments and to prevent unsafe commands.
```

bovine3dom:
Thanks!

Issue #64, opened 2017-11-16
Support 'n' for Find Again


seren:
Vimperator (and Vim) use lower-case `n` to jump to the next item matching a previous search. Firefox has this built-in with Command-G. Would it be possible to wire up `n` to trigger `Find Again`? Thanks.

dumblob:
If `Command-G` somehow supports also a search in a reverse order (`N` in vim), then it sounds feasible.

bovine3dom:
We were just going to implement our own find mode because we wanted incsearch, hlsearch and regex.

I believe @cmcaine has some code locally that he was working on.

pholat:
I've started to use the plugin as I found it looking for Vimperator replacement 👍 
Installing the latest version of plugin I believe I saw a few times search bar and could use Vim/Vimperator like search - I'll appreciate if someone post what status is of this feature.

bovine3dom:
@pholat we don't have it yet.

pholat:
@bovine3dom thanks! I'm looking forward to it.

mrlinuxfish:
I would also like this or something similar. As it is right now, I have no idea how to go to the next instance of text while searching.

bovine3dom:
@mrlinuxfish Ctrl-G

wschmrdrNS:
And how about in the other direction?

bovine3dom:
@wschmrdrNS Ctrl-Shift-G

Koushien:
NB F3 and Shift+F3 do find again, find previous respectively as well.
https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly

wschmrdrNS:
The point of having 'n' and 'N' available is that you can continue to
search on whatever is in the find buffer after making an update to
text. You should not have to constantly re-type the same search
parameter over and over again.

mMontu:
Any news about this issue?

bovine3dom:
@mMontu there's a branch with most of the work done. It's blocked by the "provide feedback to users" bug: you can happily search on the branch but you have to do it blind, it doesn't tell you what you're searching for.

mMontu:
@bovine3dom well, that is more than enough for me. Is it possible to add it to master and enable it through an option? Or the best option is to cherry pick it and do a local build?

bovine3dom:
[...]

PR #4482, opened 2022-11-29
Comments: Change broken "findnext -1" to "findnext 1 -?" in config

demiaus:
"findnext -1" doesn't seem to work anymore. Although commented out, the binding should reflect the working "findnext 1 -?" for possible future use and for people looking to use it in their rc.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -293,7 +293,7 @@ export class default_config {
         // "/": "fillcmdline find",
         // "?": "fillcmdline find -?",
         // n: "findnext 1",
-        // N: "findnext -1",
+        // N: "findnext 1 -?",
         // ",<Space>": "nohlsearch",
         M: "gobble 1 quickmark",
         B: "fillcmdline taball",


bovine3dom:
Oh no, did we actually break this? We probably broke a lot of people's binds if -1 ever worked.

I'll try to look later to see if it did work because I'd rather fix that than change the docs :)

Thanks for the PR!

demiaus:
> if -1 ever worked.

It did, but I'm not sure when it broke. I think I first copied the binding from your comment way back [comment](https://github.com/tridactyl/tridactyl/issues/64#issuecomment-496913151) :smile: 

Some shallow googling shows it's been in some custom configs at least

GHolk:
Yes, it did break it. The [vercel/arg](https://github.com/vercel/arg) treat negative number as option.
I would like to fork **arg** to change its behavior.

demiaus:
That's better for sure! :) I think I've seen negative numbers used with `tabmove`, that probably broke also. I'll go ahead and close this.

PR #4486, opened 2022-12-05
Correct highlight position while scrolling

GHolk:
I introduce `compute-scroll-into-view` package to scroll a selection range into view.
The highlight's position is relative to the whole document (`position: absolute`),
so it would not match if the hightlight text inside a scrollable area,
or the matched text change its position relative to the absolute position.

I overwrite the `getBoundingClientRect` and `getClientRects` so the `DOM.isVisible` will work on it.
For `compute-scroll-into-view`, I create a fake node with same position
so it can compute the right scroll method.

A new method `queryInRange` is used to find the focusable element containing the matched text.
It will find all parent nodes of the node which is the beginning of the selection;
Then all the following nodes in *document order* until reach the node which is the end of the selection.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "arg": "^5.0.2",
     "cleanslate": "^0.10.1",
+    "compute-scroll-into-view": "^1.0.17",
     "csp-serdes": "github:cmcaine/csp-serdes",
     "css": "^3.0.0",
     "editor-adapter": "^0.0.3",
--- a/src/content/finding.ts
+++ b/src/content/finding.ts
@@ -117,24 +117,47 @@ class FindHighlight extends HTMLSpanElement {
         if (!DOM.isVisible(this)) {
             this.scrollIntoView({ block: "center", inline: "center" })
         }
-
-        // TODO: find in all nodes but not just start container
-        let parentElement = this.range.startContainer.parentElement
-        loop: while (parentElement) {
-            switch (parentElement.nodeName.toLowerCase()) {
-                case "a":
-                case "input":
-                case "button":
-                case "details":
-                    parentElement.focus()
-                    break loop
-            }
-            parentElement = parentElement.parentElement
-        }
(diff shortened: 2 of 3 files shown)


bovine3dom:
Thanks for this PR! The CI is failing because we are getting a lot of type errors on this:

```
src/content/finding.ts:110:39 - error TS2345: Argument of type '{ nodeType: number; getBoundingClientRect: () => DOMRect; parentElement: any; }' is not assignable to parameter of type 'Element'.
  Type '{ nodeType: number; getBoundingClientRect: () => DOMRect; parentElement: any; }' is missing the following properties from type 'Element': assignedSlot, attributes, classList, className, and 120 more.

110         const actions = scrollCompute(fakeNode, option)
                                          ~~~~~~~~

src/content/finding.ts:121:34 - error TS2339: Property 'focus' does not exist on type 'Element'.

121         if (focusable) focusable.focus()
                                     ~~~~~

src/content/finding.ts:124:18 - error TS2339: Property 'style' does not exist on type 'Element'.

124             node.style.background = `rgba(255,127,255,0.5)`
                     ~~~~~

src/content/finding.ts:147:30 - error TS2339: Property 'matches' does not exist on type 'Node'.

147                     if (node.matches(selector)) {
                                 ~~~~~~~

src/content/finding.ts:155:34 - error TS2740: Type 'Node' is missing the following properties from type 'Element': assignedSlot, attributes, classList, className, and 72 more.

155         if (walker.parentNode()) return walker.currentNode
                                     ~~~~~~~~~~~~~~~~~~~~~~~~~

src/content/finding.ts:158:13 - error TS2322: Type 'Node' is not assignable to type 'Element'.

158             return walker.currentNode
                ~~~~~~~~~~~~~~~~~~~~~~~~~


Found 6 errors.

error Command failed with exit code 1.
```

I'll try to look at it soon but I can't promise I'll get round to it before I disappear over Christmas :)

GHolk:
ok, I will try

GHolk:
Now it should work

PR #4532, opened 2023-01-08
Add multiple keyboard layouts

treapster:
So it's the next step after #4439, and now we can continue the discussion we began there. I didn't implement any fancy layout merging for now, just the layouts.


bovine3dom:
Broken fancy merging patch here. I don't have time to debug it right now but I didn't want to lose it / commit broken code

I'll look tomorrow if you don't beat me to it :)

```patch
diff --git a/src/excmds.ts b/src/excmds.ts
index 736973fa..22e03655 100644
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -4299,15 +4299,6 @@ export async function bindwizard(...args: string[]) {
     return gobble("<CR>", `fillcmdline_notrail bind --mode=${mode}`, ...args)
 }
 
-/** Change keyboard layout for [[kaylayoutforce]].
- *
- * Executing the command does not affect your physical layout when typing, it is only used for interpreting shortcuts.
- */ // Compiler doesn't like the keyof type, why is that?
-export function setkeyboardlayout(arg /*: keyof config.default_config["keyboardlayouts"]*/): void {
-    const layout = config.get("keyboardlayouts")[arg]
-    if (layout) config.set("keylayoutforcemapping", layout)
-    else fillcmdline_notrail(`${arg} is unknown layout`)
-}
 /**
  * Like [[bind]] but for a specific url pattern (also see [[seturl]]).
  *
index b687a1d4..7b900d21 100644
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -555,211 +555,7 @@ export class default_config {
      */
     keylayoutforce: "true" | "false" = "false"
 
-    /**
-     * Key codes for printable keys for [[keylayoutforce]], lower and upper register.
-     * See https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values
-     * These maps are assigned to [[keylayoutforcemapping]] via `:setkeyboardlayout` command,
-     * but keylayoutforcemapping can also be changed manually with `:set`.
-     *
-     * If your layout is missing here, you can contribute it with the help of [this](https://gistpreview.github.io/?324119c773fac31651f6422087b36804)
-     * tool.
-     */
-    keyboardlayouts = {
-        qwerty: {
-            KeyA: ["a", "A"],
-            KeyB: ["b", "B"],
-            KeyC: ["c", "C"],
-            KeyD: ["d", "D"],
-            KeyE: ["e", "E"],
-            KeyF: ["f", "F"],
-            KeyG: ["g", "G"],
-            KeyH: ["h", "H"],
-            KeyI: ["i", "I"],
-            KeyJ: ["j", "J"],
-            KeyK: ["k", "K"],
-            KeyL: ["l", "L"],
-            KeyM: ["m", "M"],
-            KeyN: ["n", "N"],
-            KeyO: ["o", "O"],
-            KeyP: ["p", "P"],
-            KeyQ: ["q", "Q"],
-            KeyR: ["r", "R"],
-            KeyS: ["s", "S"],
-            KeyT: ["t", "T"],
-            KeyU: ["u", "U"],
-            KeyV: ["v", "V"],
-            KeyW: ["w", "W"],
-            KeyX: ["x", "X"],
-            KeyY: ["y", "Y"],
-            KeyZ: ["z", "Z"],
-            Digit0: ["0", ")"],
-            Digit1: ["1", "!"],
-            Digit2: ["2", "@"],
-            Digit3: ["3", "#"],
-            Digit4: ["4", "$"],
-            Digit5: ["5", "%"],
-            Digit6: ["6", "^"],
-            Digit7: ["7", "&"],
-            Digit8: ["8", "*"],
-            Digit9: ["9", "("],
-            Equal: ["=", "+"],
-            Backquote: ["`", "~"],
-            Backslash: ["\\", "|"],
-            Period: [".", ">"],
-            Comma: [",", "<"],
-            Semicolon: [";", ":"],
-            Slash: ["/", "?"],
-            BracketLeft: ["[", "{"],
-            BracketRight: ["]", "}"],
-            Quote: ["'", '"'],
-            Minus: ["-", "_"],
-        },
-        azerty: {
-            Backquote: ["²", "²"],
-            Digit1: ["&", "1"],
-            Digit2: ["é", "2"],
-            Digit3: ['"', "3"],
-            Digit4: ["'", "4"],
-            Digit5: ["(", "5"],
-            Digit6: ["-", "6"],
-            Digit7: ["è", "7"],
-            Digit8: ["_", "8"],
-            Digit9: ["ç", "9"],
-            Digit0: ["à", "0"],
-            Minus: [")", "°"],
-            Equal: ["=", "+"],
-            KeyQ: ["a", "A"],
-            KeyW: ["z", "Z"],
-            KeyE: ["e", "E"],
-            KeyR: ["r", "R"],
-            KeyT: ["t", "T"],
-            KeyY: ["y", "Y"],
-            KeyU: ["u", "U"],
-            KeyI: ["i", "I"],
-            KeyO: ["o", "O"],
-            KeyP: ["p", "P"],
-            BracketRight: ["$", "£"],
-            Backslash: ["*", "µ"],
-            KeyA: ["q", "Q"],
-            KeyS: ["s", "S"],
-            KeyD: ["d", "D"],
-            KeyF: ["f", "F"],
-            KeyG: ["g", "G"],
-            KeyH: ["h", "H"],
-            KeyJ: ["j", "J"],
-            KeyK: ["k", "K"],
-            KeyL: ["l", "L"],
-            Semicolon: ["m", "M"],
-            Quote: ["ù", "%"],
-            KeyZ: ["w", "W"],
-            KeyX: ["x", "X"],
-            KeyC: ["c", "C"],
-            KeyV: ["v", "V"],
-            KeyB: ["b", "B"],
-            KeyN: ["n", "N"],
-            KeyM: [",", "?"],
-            Comma: [";", "."],
-            Period: [":", "/"],
-            Slash: ["!", "§"],
-        },
-        german: {
-            Digit1: ["1", "!"],
-            Digit2: ["2", '"'],
-            Digit3: ["3", "§"],
-            Digit4: ["4", "$"],
-            Digit5: ["5", "%"],
-            Digit6: ["6", "&"],
-            Digit7: ["7", "/"],
-            Digit8: ["8", "("],
-            Digit9: ["9", ")"],
-            Digit0: ["0", "="],
-            Minus: ["ß", "?"],
-            KeyQ: ["q", "Q"],
-            KeyW: ["w", "W"],
-            KeyE: ["e", "E"],
-            KeyR: ["r", "R"],
-            KeyT: ["t", "T"],
-            KeyY: ["z", "Z"],
-            KeyU: ["u", "U"],
-            KeyI: ["i", "I"],
-            KeyO: ["o", "O"],
-            KeyP: ["p", "P"],
-            BracketLeft: ["ü", "Ü"],
-            BracketRight: ["+", "*"],
-            Backslash: ["#", "'"],
-            KeyA: ["a", "A"],
-            KeyS: ["s", "S"],
-            KeyD: ["d", "D"],
-            KeyF: ["f", "F"],
-            KeyG: ["g", "G"],
-            KeyH: ["h", "H"],
-            KeyJ: ["j", "J"],
-            KeyK: ["k", "K"],
-            KeyL: ["l", "L"],
-            Semicolon: ["ö", "Ö"],
-            Quote: ["ä", "Ä"],
-            KeyZ: ["y", "Y"],
-            KeyX: ["x", "X"],
-            KeyC: ["c", "C"],
-            KeyV: ["v", "V"],
-            KeyB: ["b", "B"],
-            KeyN: ["n", "N"],
-            KeyM: ["m", "M"],
-            Comma: [",", ";"],
-            Period: [".", ":"],
-            Slash: ["-", "_"],
-            Backquote: ["", "°"],
-        },
-        dvorak: {
-            Backquote: ["`", "~"],
-            Digit1: ["1", "!"],
-            Digit2: ["2", "@"],
-            Digit3: ["3", "#"],
-            Digit4: ["4", "$"],
-            Digit5: ["5", "%"],
-            Digit6: ["6", "^"],
-            Digit7: ["7", "&"],
-            Digit8: ["8", "*"],
-            Digit9: ["9", "("],
-            Digit0: ["0", ")"],
-            Minus: ["[", "{"],
-            Equal: ["]", "}"],
-            KeyQ: ["'", '"'],
-            KeyW: [",", "<"],
-            KeyE: [".", ">"],
-            KeyR: ["p", "P"],
-            KeyT: ["y", "Y"],
-            KeyY: ["f", "F"],
-            KeyU: ["g", "G"],
-            KeyI: ["c", "C"],
-            KeyO: ["r", "R"],
-            KeyP: ["l", "L"],
-            BracketLeft: ["/", "?"],
-            BracketRight: ["=", "+"],
-            Backslash: ["\\", "|"],
-            KeyA: ["a", "A"],
-            KeyS: ["o", "O"],
-            KeyD: ["e", "E"],
-            KeyF: ["u", "U"],
-            KeyG: ["i", "I"],
-            KeyH: ["d", "D"],
-            KeyJ: ["h", "H"],
-            KeyK: ["t", "T"],
-            KeyL: ["n", "N"],
-            Semicolon: ["s", "S"],
-            Quote: ["-", "_"],
-            KeyZ: [";", ":"],
-            KeyX: ["q", "Q"],
-            KeyC: ["j", "J"],
-            KeyV: ["k", "K"],
-            KeyB: ["x", "X"],
-            KeyN: ["b", "B"],
-            KeyM: ["m", "M"],
-            Comma: ["w", "W"],
-            Period: ["v", "V"],
-            Slash: ["z", "Z"],
-        },
-    }
+    keylayoutforcebase: "qwerty"
 
     /**
      * Key code to character mapping to use when keylayoutforce is enabled. You can change it with [[setkeyboardlayout]]
@@ -770,7 +566,7 @@ export class default_config {
      * You must instead set the lower and upper registers using a string with no spaces in it, for example
      * `:set keylayoutforcemapping Digit2: 2"` for the British English layout.
      */
-    keylayoutforcemapping = this.keyboardlayouts["qwerty"]
+    keylayoutforcemapping = {}
 
     /**
      * Automatically place these sites in the named container.
diff --git a/src/lib/keyboardlayouts.ts b/src/lib/keyboardlayouts.ts
new file mode 100644
index 00000000..406292db
--- /dev/null
+++ b/src/lib/keyboardlayouts.ts
@@ -0,0 +1,206 @@
+/**
+ * Key codes for printable keys for [[keylayoutforce]], lower and upper register.
+ * See https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values
+ * These maps are assigned to [[keylayoutforcemapping]] via `:setkeyboardlayout` command,
+ * but keylayoutforcemapping can also be changed manually with `:set`.
+ *
+ * If your layout is missing here, you can contribute it with the help of [this](https://gistpreview.github.io/?324119c773fac31651f6422087b36804)
+ * tool.
+ */
+export const keyboardlayouts = {
+    qwerty: {
+        KeyA: ["a", "A"],
+        KeyB: ["b", "B"],
+        KeyC: ["c", "C"],
+        KeyD: ["d", "D"],
+        KeyE: ["e", "E"],
+        KeyF: ["f", "F"],
+        KeyG: ["g", "G"],
+        KeyH: ["h", "H"],
+        KeyI: ["i", "I"],
+        KeyJ: ["j", "J"],
+        KeyK: ["k", "K"],
+        KeyL: ["l", "L"],
+        KeyM: ["m", "M"],
+        KeyN: ["n", "N"],
+        KeyO: ["o", "O"],
+        KeyP: ["p", "P"],
+        KeyQ: ["q", "Q"],
+        KeyR: ["r", "R"],
+        KeyS: ["s", "S"],
+        KeyT: ["t", "T"],
+        KeyU: ["u", "U"],
+        KeyV: ["v", "V"],
+        KeyW: ["w", "W"],
+        KeyX: ["x", "X"],
+        KeyY: ["y", "Y"],
+        KeyZ: ["z", "Z"],
+        Digit0: ["0", ")"],
+        Digit1: ["1", "!"],
+        Digit2: ["2", "@"],
+        Digit3: ["3", "#"],
+        Digit4: ["4", "$"],
+        Digit5: ["5", "%"],
+        Digit6: ["6", "^"],
+        Digit7: ["7", "&"],
+        Digit8: ["8", "*"],
+        Digit9: ["9", "("],
+        Equal: ["=", "+"],
+        Backquote: ["`", "~"],
+        Backslash: ["\\", "|"],
+        Period: [".", ">"],
+        Comma: [",", "<"],
+        Semicolon: [";", ":"],
+        Slash: ["/", "?"],
+        BracketLeft: ["[", "{"],
+        BracketRight: ["]", "}"],
+        Quote: ["'", '"'],
+        Minus: ["-", "_"],
+    },
+    azerty: {
+        Backquote: ["²", "²"],
+        Digit1: ["&", "1"],
+        Digit2: ["é", "2"],
+        Digit3: ['"', "3"],
+        Digit4: ["'", "4"],
+        Digit5: ["(", "5"],
+        Digit6: ["-", "6"],
+        Digit7: ["è", "7"],
+        Digit8: ["_", "8"],
+        Digit9: ["ç", "9"],
+        Digit0: ["à", "0"],
+        Minus: [")", "°"],
+        Equal: ["=", "+"],
+        KeyQ: ["a", "A"],
+        KeyW: ["z", "Z"],
+        KeyE: ["e", "E"],
+        KeyR: ["r", "R"],
+        KeyT: ["t", "T"],
+        KeyY: ["y", "Y"],
+        KeyU: ["u", "U"],
+        KeyI: ["i", "I"],
+        KeyO: ["o", "O"],
+        KeyP: ["p", "P"],
+        BracketRight: ["$", "£"],
+        Backslash: ["*", "µ"],
+        KeyA: ["q", "Q"],
+        KeyS: ["s", "S"],
+        KeyD: ["d", "D"],
+        KeyF: ["f", "F"],
+        KeyG: ["g", "G"],
+        KeyH: ["h", "H"],
+        KeyJ: ["j", "J"],
+        KeyK: ["k", "K"],
+        KeyL: ["l", "L"],
+        Semicolon: ["m", "M"],
+        Quote: ["ù", "%"],
+        KeyZ: ["w", "W"],
+        KeyX: ["x", "X"],
+        KeyC: ["c", "C"],
+        KeyV: ["v", "V"],
+        KeyB: ["b", "B"],
+        KeyN: ["n", "N"],
+        KeyM: [",", "?"],
+        Comma: [";", "."],
+        Period: [":", "/"],
+        Slash: ["!", "§"],
+    },
+    german: {
+        Digit1: ["1", "!"],
+        Digit2: ["2", '"'],
+        Digit3: ["3", "§"],
+        Digit4: ["4", "$"],
+        Digit5: ["5", "%"],
+        Digit6: ["6", "&"],
+        Digit7: ["7", "/"],
+        Digit8: ["8", "("],
+        Digit9: ["9", ")"],
+        Digit0: ["0", "="],
+        Minus: ["ß", "?"],
+        KeyQ: ["q", "Q"],
+        KeyW: ["w", "W"],
+        KeyE: ["e", "E"],
+        KeyR: ["r", "R"],
+        KeyT: ["t", "T"],
+        KeyY: ["z", "Z"],
+        KeyU: ["u", "U"],
+        KeyI: ["i", "I"],
+        KeyO: ["o", "O"],
+        KeyP: ["p", "P"],
+        BracketLeft: ["ü", "Ü"],
+        BracketRight: ["+", "*"],
+        Backslash: ["#", "'"],
+        KeyA: ["a", "A"],
+        KeyS: ["s", "S"],
+        KeyD: ["d", "D"],
+        KeyF: ["f", "F"],
+        KeyG: ["g", "G"],
+        KeyH: ["h", "H"],
+        KeyJ: ["j", "J"],
+        KeyK: ["k", "K"],
+        KeyL: ["l", "L"],
+        Semicolon: ["ö", "Ö"],
+        Quote: ["ä", "Ä"],
+        KeyZ: ["y", "Y"],
+        KeyX: ["x", "X"],
+        KeyC: ["c", "C"],
+        KeyV: ["v", "V"],
+        KeyB: ["b", "B"],
+        KeyN: ["n", "N"],
+        KeyM: ["m", "M"],
+        Comma: [",", ";"],
+        Period: [".", ":"],
+        Slash: ["-", "_"],
+        Backquote: ["", "°"],
+    },
+    dvorak: {
+        Backquote: ["`", "~"],
+        Digit1: ["1", "!"],
+        Digit2: ["2", "@"],
+        Digit3: ["3", "#"],
+        Digit4: ["4", "$"],
+        Digit5: ["5", "%"],
+        Digit6: ["6", "^"],
+        Digit7: ["7", "&"],
+        Digit8: ["8", "*"],
+        Digit9: ["9", "("],
+        Digit0: ["0", ")"],
+        Minus: ["[", "{"],
+        Equal: ["]", "}"],
+        KeyQ: ["'", '"'],
+        KeyW: [",", "<"],
+        KeyE: [".", ">"],
+        KeyR: ["p", "P"],
+        KeyT: ["y", "Y"],
+        KeyY: ["f", "F"],
+        KeyU: ["g", "G"],
+        KeyI: ["c", "C"],
+        KeyO: ["r", "R"],
+        KeyP: ["l", "L"],
+        BracketLeft: ["/", "?"],
+        BracketRight: ["=", "+"],
+        Backslash: ["\\", "|"],
+        KeyA: ["a", "A"],
+        KeyS: ["o", "O"],
+        KeyD: ["e", "E"],
+        KeyF: ["u", "U"],
+        KeyG: ["i", "I"],
+        KeyH: ["d", "D"],
+        KeyJ: ["h", "H"],
+        KeyK: ["t", "T"],
+        KeyL: ["n", "N"],
+        Semicolon: ["s", "S"],
+        Quote: ["-", "_"],
+        KeyZ: [";", ":"],
+        KeyX: ["q", "Q"],
+        KeyC: ["j", "J"],
+        KeyV: ["k", "K"],
+        KeyB: ["x", "X"],
+        KeyN: ["b", "B"],
+        KeyM: ["m", "M"],
+        Comma: ["w", "W"],
+        Period: ["v", "V"],
+        Slash: ["z", "Z"],
+    },
+}
+
diff --git a/src/lib/keyseq.ts b/src/lib/keyseq.ts
index db90bbda..bf880feb 100644
--- a/src/lib/keyseq.ts
+++ b/src/lib/keyseq.ts
@@ -24,6 +24,7 @@
 import { filter, find, izip } from "@src/lib/itertools"
 import { Parser } from "@src/lib/nearley_utils"
 import * as config from "@src/lib/config"
+import { keyboardlayouts } from "@src/lib/keyboardlayouts.ts"
 import * as R from "ramda"
 import grammar from "@src/grammars/.bracketexpr.generated"
 const bracketexpr_grammar = grammar
@@ -494,7 +495,7 @@ export function minimalKeyFromKeyboardEvent(
     }
     if (config.get("keylayoutforce") === "true") {
         let newkey = keyEvent.key
-        const keycodetranslatemap = config.get("keylayoutforcemapping")
+        const keycodetranslatemap = R.mergeRight(keyboardlayouts[config.get("keylayoutforcebase")], config.get("keylayoutforcemapping"))
         const translation = keycodetranslatemap[keyEvent.code]
         if (translation) newkey = translation[+keyEvent.shiftKey]
         return new MinimalKey(newkey, modifiers)
```

treapster:
I like splitting to a separate file, but i didn't expect merging on every keypress. But to not do that we need to put keycodetranslatemap in the config too and update on every `:set` which is too bothersome.

But even with your current suggested patch, i'm not sure there are serious enough usecases to bother with adding new config var and merging. What issues are we solving with it? All the functionality user wants can be achieved by doing `:setkeyboardlayout`  + manual overrides, and if he wants to preserve it he can add it to .tridactylrc. If the point of merging is to preserve overrides across different layouts, i think doing so may be counterintuitive and unexpected for some users, plus why would anyone set layout more than once?  Also, i think we should try to minimize the need for individual overrides by just providing more layouts. Now that we have german and azerty, i guess the majority of european users won't use overrides anyway, and the rest of the world uses qwerty. That said, this changes do look like unnecessary indirection

bovine3dom:
It isn't about preserving overrides across different base layouts, that's a slightly confusing side-effect of my patch.

I don't like the idea of copying what is essentially Tridactyl default configuration into the user's configuration. It makes RC files much longer and less clear, but more importantly we cannot easily update the user's configuration if the Tridactyl default configuration changes.

We change the default configuration of Tridactyl when we find bugs in it. For example if we forgot a key on some obscure layout then we could update it easily with my patch. I think that bugs in this specific area are unlikely but to some extent that's true for any individual area ...

I agree that merging on each keypress is untidy, but we can probably improve it by caching it.

---

Maybe we just want to remove the ability to remap individual keys of the provided layouts? Maybe `keylayoutforcemapping` could become `object | string` with e.g. `en-us` or a big load of mappings that a user has chosen

treapster:
Ok, now i get the point of your patch. Then i'm fine with something along the lines of what you suggested, especially with caching. On the other hand, despite untidiness one merge should be cheap enough to not care and leave it as it is. As of removing custom overrides entirely, i think it would be weird and base layout + overrides seems more natural.

treapster:
What is a proper way to disallow bogus keylayoutforcebase values? Something like `keyof typeof keyboardlayouts` gives error in complier for me. Also, how about renaming keylayoutforce -> forcekeyboardlayout, keylayoutforcebase->keyboardlayoutbase, keylayoutforcemapping->keyboardlayoutmap? And once the patch is stable, we should not forget to update the comments to reflect the new functionality and names.

src/lib/keyseq.ts
```diff
@@ -515,3 +518,7 @@ browser.storage.onChanged.addListener(changes => {
         KEYMAP_CACHE = {}
     }
 })
+
+function updateBaseLayout() {
+    KEYCODETRANSLATEMAP = R.mergeRight(keyboardlayouts[config.get("keylayoutforcebase")], config.get("keylayoutforcemapping"))
+}
```
bovine3dom, line 524: Something like

```

function updateBaseLayout() {
    KEYCODETRANSLATEMAP = R.mergeRight(keyboardlayouts[config.get("keylayoutforcebase")], config.get("keylayoutforcemapping"))
}

setTimeout(() => config.addChangeListener("keylayoutforcebase", updateBaseLayout), 1000)
setTimeout(() => config.addChangeListener("keylayoutforcemapping", updateBaseLayout), 1000)
```

should work for invalidating but doesn't run at all and I can't work out why. (the `setTimeout`s are there because without them Tridactyl is just totally dead)

bovine3dom:
Yeah the names are pretty lame. We try to follow the pseudo namespacing of `[noun][verb]` like `tabopen` rather than `opentab` so that you can find things fairly easily with completions without using `:apropos`, so we shouldn't use `forcekeyboardlayout`.

I'll try to think about them a bit ... usually I just give up and subject people to terrible names.

---

The runtime type checking is something glacambre wrote for us here https://github.com/tridactyl/tridactyl/blob/77cd61147cd7e708320084bae958232bdff6122b/compiler/gen_metadata.ts#L90
. I think it only supports simple types so the `keyof typeof` stuff won't work unless we improve that.

treapster:
I tried to wrap my head around type checking and it seems like a typescript problem, because `checker.getTypeOfSymbolAtLocation()` for `keyof typeof keyboardlayouts` gives `string | number | symbol` instead of `"qwerty" | "azerty" | "german" | "dvorak"`, which is too broad and our toSimpleType doesn't know about symbol. But even if it knew, it wouldn't help since it's not the types we want.
This is super weird because if you define `type layouts = keyof typeof keyboardlayouts` and hover over `layouts` in vscode, you'll get a proper union type. It's also interesting that if you go to the defenition of `keyboardlayouts` and add a colon after identifier but omit type, and then hover over the previous defenition of `layouts`, you'll get `string | number | symbol` as from the checker. So, if you define an object without colons and type, VS code tools and typescript typechecker disagree on what's the right type of it's keyof:(

treapster:
But if i put keyboardlayouts in config.ts, everything works fine. So the problem is not the `keyof typeof`, but the type definition being in a different file.

treapster:
> Yeah the names are pretty lame. We try to follow the pseudo namespacing of `[noun][verb]` like `tabopen` rather than `opentab` so that you can find things fairly easily with completions without using `:apropos`, so we shouldn't use `forcekeyboardlayout`.

My point was to include a whole "keyboard" word in all the names to make them easier to find, i don't mind the ordering. And removing "force" from everything but the switch itself will make it shorter.

bovine3dom:
It was more aesthetically pleasing to have the layouts in a separate file but the runtime type checking is worth the ugliness of having it all in one file IMO - thanks for spotting that that was the problem.

I think this is good to merge now - let me know if you disagree, and thanks for all your work on this and the previous PR!

treapster:
I agree that it's good to merge. How about releasing beta after that?

bovine3dom:
Yeah, this PR is what I was waiting for :)

PR #4541, opened 2023-01-11
Skip `loadaucmds` when no config for cmdType

### LINKED ISSUE #4487: Repeated can't convert undefined to object; seems to be missing autocmd entry.
-   Ubuntu, Firefox 107.0
-   Tridactyl 1.22.1pre6524-78d0fa91

On many websites, I kept seeing this over and over in the console:

```
TypeError: can't convert undefined to object   .excmds_content.generated.ts:2097:27
```

When I view the source of the error, it pointed me to the line `const ausites = Object.keys(aucmds)` in `function loadaucmds()`.  It seems `autocmds` doesn't have keys for some event types (that is, no `{}` value), and this is causing an error.  It may be because I have an old install with an older config.

What made it go away was running:

```
autocmd HistoryState blahblahblahblah echo test
```

This vivified the `HistoryState` entry in `autocmds`.  Thus, this event was presumably the culprit.

So, this is fixed for me, but it seems it should either not spit out errors in this case, or somehow assure that necessary keys are present in `autocmds`.

paescuj:
Fixes #4487

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2359,6 +2359,7 @@ if (fullscreenApiIsPrefixed) {
 //#content
 export async function loadaucmds(cmdType: "DocStart" | "DocLoad" | "DocEnd" | "TabEnter" | "TabLeft" | "FullscreenEnter" | "FullscreenLeft" | "FullscreenChange" | "UriChange" | "HistoryState") {
     const aucmds = await config.getAsync("autocmds", cmdType)
+    if (!aucmds) return
     const ausites = Object.keys(aucmds)
     const aukeyarr = ausites.filter(e => window.document.location.href.search(e) >= 0)
     const owntab = await ownTab()


jcelliott:
I'm running into this same issue. @paescuj's fix looks pretty straightforward to me. Does anything else need to happen to get this merged?

Thanks!

paescuj:
Hey @jcelliott, thanks for your feedback! From my point of view, we're good to go 👍 I briefly went through the code and couldn't spot any negative side effects. In fact, I think we're in any case on the safer side with this fix (since `config` might also be undefined for other unexpected reasons)... But let's wait for the feedback of the maintainer (btw. thank you so much for `tridactyl` ❤️).

bovine3dom:
I wanted to merge it yesterday but I had to manually start the tests, LGTM and thanks for the PR!

paescuj:
Thanks! Much appreciated 😃

PR #4546, opened 2023-01-14
Fix container icon colours in completions

rddunphy:
Container icons in tab completions for `:tab` etc. were being coloured using `-moz-context-properties: fill`, which doesn't appear to work (FF108) - all container icons were displayed as black. Using `background-color` and `mask-image` instead fixes it.


bovine3dom:
Looks great, thanks!

PR #4547, opened 2023-01-14
Center description in readme

paescuj:
Seems to fit a little better since all other elements in the header are centered as well

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -6,7 +6,7 @@ Tridactyl
 <br>
 </h1>
 
-<h4 align="center">Replace Firefox's default control mechanism with one modelled on the one true editor, Vim.</h4>    
+<h4 align="center">Replace Firefox's default control mechanism with one modelled on the one true editor, Vim.</h4>
 
 <p align="center">
 <a href="https://travis-ci.org/tridactyl/tridactyl"><img src="https://travis-ci.org/tridactyl/tridactyl.svg?branch=master" alt="Build Status"></a>
--- a/readme.md
+++ b/readme.md
@@ -6,7 +6,7 @@ Tridactyl
 <br>
 </h1>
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -6,7 +6,7 @@ Tridactyl
 <br>
 </h1>
 
-<h4>Replace Firefox's default control mechanism with one modelled on the one true editor, Vim. </h4>
+<h4 align="center">Replace Firefox's default control mechanism with one modelled on the one true editor, Vim.</h4>
```
paescuj, line 9:
```suggestion
<h4 align="center">Replace Firefox's default control mechanism with one modelled on the one true editor, Vim.</h4>
```

paescuj:
Should we skip e2e tests on readme changes? Could be achieved with the [`paths-ignore`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths) option:

```yaml
    paths-ignore:
      - 'readme.md'
      # or maybe even
      - '**.md'
```

bovine3dom:
Thanks for the PR. I need to just check how it looks on my PC.

As for the markdown files, some of them are compiled to HTML in the build process so we shouldn't exclude all of them. We could exclude the readme but I don't think it's a problem that arises often.

The concurrency cancelling seems like a good idea :)

paescuj:
Thanks!

> As for the markdown files, some of them are compiled to HTML in the build process so we shouldn't exclude all of them. We could exclude the readme but I don't think it's a problem that arises often.

Makes sense! However, since there are only two additional lines, I'd do the filtering for the `readme.md` file.

> The concurrency cancelling seems like a good idea :)

Great! I'm going to create a pull request for this 👍

PR #4573, opened 2023-01-30
Sort history completions by frecency

### LINKED ISSUE #4572: :tabopen completions should display sooner
Steps to reproduce:

1. `:tabopen ` - no completions shown
2. `:tab ` - all completions are shown
3. `:tabopen  ` - completions are shown which match ` `

`:tabopen` should behave more like `:tab`, showing completions even when no filter is provided

Related - #4562 has made this more obviously useful, since it would let people discover `searchurls`

rddunphy:
- Add exponential decay frecency algorithm (https://wiki.mozilla.org/User:Jesse/NewFrecency). Each new visit gets a count of 2, and loses half that score every halflife (default is 30 days).
- Add `frecencyhalflife` option - setting to 0 sorts by visitCount (current behaviour)
- SearchURLs are scored by the number of matching visits in history + `searchurlweight`
- Add searchurls to top sites (fixes #4572)
- Add `usetopsites` option (default true for current behaviour) - if false, completions are taken from history results even fore empty string
- Add `providers.HistoryItem` class to store history/bookmark etc. information in a more TypeScript-friendly way. This also means we can calculate frecency scores once and then store them as a property of the entry - previously, scores were being recalculated multiple times for each entry while sorting.
- Change default protocol (e.g. for `open github.com`) from `http://` to `https://` (Firefox redirects to `https://` now anyway, so this was just cluttering up the history with duplicate entries)

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1138,7 +1138,7 @@ export class default_config {
     usetopsites: "true" | "false" = "true"
 
     /**
-     * Half life in days of visits used for exponential decay in frecency calculation. Set to 0 to disable frecency and sort by visited count only.
+     * Half life in days of visits used for exponential decay in frecency calculation. Set to 0 to disable frecency and sort by visited count only. Smaller numbers mean that recency of visits matter more for ranking; bigger numbers mean that the absolute number of visits is more important.
      */
     frecencyhalflife = 30
 
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1143,6 +1143,11 @@ export class default_config {
     frecencyhalflife = 30
 
     /**
+     * Minimum interval between searchurl score updates, in seconds. (To improve performance, searchurl scores used to rank history completions are cached.)
+     */
+    searchurlscoreupdateinterval = 600
+
+    /**
      * When displaying bookmarks in history completions, how many page views to pretend they have.
      */
     bmarkweight = 100
@@ -1150,7 +1155,7 @@ export class default_config {
     /**
      * When displaying searchurls in history completions, how many page views to pretend they have.
      */
-    searchurlweight = 150
+    searchurlweight = 100
 
     /**
      * Default selector for :goto command.
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1133,7 +1133,7 @@ export class default_config {
     historyresults = 50
 
     /**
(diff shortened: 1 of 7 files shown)


src/lib/config.ts
```diff
@@ -1132,6 +1132,16 @@ export class default_config {
      */
     historyresults = 50
 
+    /**
+     * Whether to use the Firefox "top sites" API in history completions. This setting is ignored if [[historyresults]] is set to 0.
+     */
+    usetopsites: "true" | "false" = "true"
+
+    /**
+     * Half life in days of visits used for exponential decay in frecency calculation. Set to 0 to disable frecency and sort by visited count only.
```
bovine3dom, line 1141:
```suggestion
     * Half life in days of visits used for exponential decay in frecency calculation. Set to 0 to disable frecency and sort by visited count only. Smaller numbers mean that recency of visits matter more for ranking; bigger numbers mean that the absolute number of visits is more important.
```

bovine3dom:
Even if `usetopsites` is disabled, if I type quickly, the real results always seem to be replaced by the top sites.

And we should probably use the `completions` object to store the settings in right? https://github.com/tridactyl/tridactyl/blob/7b4011281873eb24ba2251709ad45bc275a2d5bb/src/lib/config.ts#L1163

rddunphy:
> Even if `usetopsites` is disabled, if I type quickly, the real results always seem to be replaced by the top sites.

I can't replicate that - but I also can't figure out a way of adding something to top sites without also adding it to history, so I'm not really sure how to test this!

> And we should probably use the `completions` object to store the settings in right?

That would make sense, but I notice there are quite a few completion-related options already that aren't in `completions` - e.g. `bmarkweight`, `historyresults`, `tabshowhidden`. I'm guessing moving them all for consistency is out of the question, as this would break everyone's `tridactylrc`s?

rddunphy:
Something like this? Now, the searchUrl scoring happens in the background, so there shouldn't be any lag.

I noticed that `sanitise tridactyllocal` doesn't clear this cache - is that not how you're supposed to reset state variables, or do I need to add some clean-up code somewhere?

bovine3dom:
oh man sorry @rddunphy, i totally forgot about this PR. if it's any small consolation i'm not convinced my implementation actually works https://github.com/tridactyl/tridactyl/issues/5502

rddunphy:
No problem!

Issue #4364, opened 2022-08-22
Transparency of the search bar and menu does not work with the "Dark Reader" extension


glowbigger:
-   Steps to reproduce: Install the "Dark Reader" extension. Use a theme like "midnight" and observe that there is no transparency between the menu and the bar when you press 'o'. There is no issue, however, on the new tab page where the dark reader extension is disabled.

-   Tridactyl version: 1.22.1

-   Firefox version: 103.0.2

bovine3dom:
Fixed in #4576

jasht1:
This still appears to be an issue, similarly to https://github.com/tridactyl/tridactyl/issues/4294 I use the [GruxBox Theme](https://github.com/jrolfs/gruvbox-material-tridactyl) which has some pretty large iframe containers. On any site where [darkreader](https://github.com/darkreader/darkreader) is active I still get the white boxes. 

This is potentially more relevant to darkreader so I have added a more comprehensive Issue [there](https://github.com/darkreader/darkreader/issues/14410), but I am also updating this thread for the sake of completeness.

- Tridactyl Version:  1.24.2
- LibreWolf Version: 139.0.4-1
&& 
- Firefox Version:  139.0.4 (64-bit)

PR #4576, opened 2023-01-31
Prevent Firefox from adding a white background to iframe in Midnight theme

### LINKED ISSUE #4294: Quirk with dark mode and transparent iframes
I originally noticed this with a third party theme:

https://github.com/jrolfs/gruvbox-material-tridactyl

But, the same issue is visible in the midnight theme that ships with Tridactyl, and as far as I can tell it's a general issue with transparent iframes and dark mode and will affect any theme that uses a transparent iframe for shadows etc.

To reproduce:

log into github.com and go to the settings page where it lets you choose a user theme.

If you select any of their dark themes, the Iframe that tridactyl uses for it's commandline gets a solid white color. If you select any of the non-dark themes, it becomes transparent again.

I couldn't figure out what was happening from using the dev tools to inspect it, but this blog posts suggests it may be an interaction between dark mode and iframes:

https://fvsch.com/transparent-iframes

Which quotes a spec saying:

> In order to preserve expected color contrasts, in the case of embedded documents typically rendered over a transparent canvas (such as provided via an HTML <iframe> element), if the used color scheme of the element and the used color scheme of the embedded document’s root element do not match, then the UA must use an opaque canvas of the Canvas color appropriate to the embedded document’s used color scheme instead of a transparent canvas.

The fix seems to be setting color-scheme: light on the iframe, wwhich is the opposite of what seems logical, but does seem to fix the problem for me.

rddunphy:
Fixes #4294 and #4364 - see also https://github.com/jrolfs/gruvbox-material-tridactyl/issues/2

As of FF102, Firefox adds a white background to the iframe - presumably this applies to all themes, but midnight is the only one where this is apparent due to the transparent space between the commandline and the completions.

(Also, pretty made a bunch of changes - the only actual change is line 120)

### DIFF
--- a/src/static/themes/midnight/midnight.css
+++ b/src/static/themes/midnight/midnight.css
@@ -1,6 +1,6 @@
 :root {
     --tridactyl-font-family: "San Francisco", sans-serif;
-    --tridactyl-bg: #161a20; 
+    --tridactyl-bg: #161a20;
     --tridactyl-fg: white;
 
     --tridactyl-status-font-size: 12px;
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM, thanks!

Issue #4555, opened 2023-01-17
Allow comments in multiline RC file commands


bovine3dom:
At the moment

```
js window.alert("hello \
" this is a comment
world")
```

doesn't work in our RC files.

We need to change the order in this function:

https://github.com/tridactyl/tridactyl/blob/e26865e4e4cbd000509bb72199ddb50bfc718a30/src/background/config_rc.ts#L61

so that we remove comments before we start joining lines together.

mozbugbox:
Should also consider javascript single line comment: `//`.

callmetushar123:
@bovine3dom Hello assign this also to me please

bovine3dom:
In our RC files

```
js window.alert("hello \
" this is a comment
world")
```

should have the comment removed and then the escaped lines should be merged, like this:

```
js window.alert("hello \
" this is a comment
world")

remove comments -->

js window.alert("hello \
world")

merge lines -->

js window.alert("hello world")

```

What currently happens is

```
js window.alert("hello \
" this is a comment
world")

merge lines --->

js window.alert("hello " this is a comment
world")

remove comments -->

js window.alert("hello " this is a comment
world")
```


---

You can test it by installing the native messenger (`:native`), putting snippet in the original post in a file and then running `:source [~/path/to/file]` in Tridactyl :)

bovine3dom:
Fixed in #4651

PR #4583, opened 2023-02-05
Allow comments in multiline RC file commands. Fixes issue #4555

callmetushar123:
Signed-off-by: Tushar <tusharbarman123@gmail.com>


callmetushar123:
@bovine3dom I think I made a mistake here can you guide me a bit?

bovine3dom:
> I think I made a mistake here can you guide me a bit?

Can you be a bit more precise? Why do you think you made a mistake? What works / doesn't work?

If it makes it easier for you, I could write you a failing test so you can see if you have fixed it

src/background/config_rc.ts
```diff
@@ -59,17 +59,21 @@ export async function runRc(rc: string) {
 }
 
 export function rcFileToExCmds(rcText: string): string[] {
-    // string-join lines that end with /
-    const joined = rcText.replace(/\\\n/g, "")
+   // Remove empty and comment lines
+   const rm= Array.from(rcText).filter(
```
callmetushar123, line 63: I used to convert the string to an Array because filter was showing error. @bovine3dom

src/background/config_rc.ts
```diff
@@ -59,17 +59,21 @@ export async function runRc(rc: string) {
...
+    x =>
+        /\S/.test(x) &&
+        !x.trim().startsWith('"') &&
+        !x.trim().startsWith("#"),
+   )
+   // Converting from an array to strings  
+   const str = rm.join(" ");
```
callmetushar123, line 70: Here I tried to convert the array back to string so as I could perform the string operations in the later stages. @bovine3dom

callmetushar123:
image

The failed test cases is showing this reason

src/background/config_rc.ts
bovine3dom: Have you tried running this code step by step in a REPL and seeing what it does?

The filter is expecting an array that contains one line of text per element.

`Array.from(string)` makes an array where each character in the string becomes an individual element, which isn't what we want.

src/background/config_rc.ts
bovine3dom: I don't think that's the right approach. Your command would add spaces where there were none before.

bovine3dom:
Those errors are linting errors that you can fix with `yarn run eslint --fix [file]`.

I think this PR might be a bit difficult for you if you haven't looked at regex/string handling before. Feel free to pick a different issue to work on if you are struggling.

Otherwise the approach I would take (because it requires the least thinking rather than because I think it is efficient) is this:

1. Split the string on all newline characters into an array
2. Remove lines that are comments
3. Join the array into a string restoring the newline characters
4. Remove the escaped newline characters
5. Split the array on remaining newline characters to get the individual ex commands

callmetushar123:
Thank you so much for the suggestion, I will try again

Issue #4580, opened 2023-02-02
feat: Traversing tab/buffer history (navigate throught the stack of previously active tabs)


maricn:
### `:tab` back and forth
I find `<C-^>` (`<c-6>` if you prefer) very useful to jump to previously open buffer. As I often have many open tabs (50+), it would be quite useful to not only navigate back-and-forth between the two last active buffers, but to be able to traverse deeper down the MRU (most recently used) buffers.

Of course, this would mean bidirectional traversal, like `:tab <` and `:tab >` to go back and forward through MRU tabs.

### `:tab` MRU list
Come to think of it, another related feature that could be useful would be `:tab-mru` that would open the tab list dialog but sort the tabs by their most recent active-ity instead of their natural order in firefox.

Regarding the command itself, perhaps it would be somewhat intuitive if `:tab #` would open such MRU list, and automatically select the first item to preserve backward compatibility, but if you'd type `:tab #2` it would select the 2nd item on the MRU stack.
 
What do you think?

bovine3dom:
Your `:tab-mru` already exists, sort of, with `:set tabsort mru`, but maybe there's a reason why you would want to be able to access both orders without changing a setting.

The first suggestion is a duplicate of #2378 - it's something I would really like but I'm not really sure how to do it as each time you navigate to a tab it then becomes the most recently used one, so we would need to keep a separate list and decide what interactions with a tab count as "using" it. We have a similar problem with `g;`.

maricn:
### `:tab` MRU list
Thanks for sharing the `:set tabsort mru`, didn't know about it. The global config issue is not a biggie.
<img width="820" alt="image" src="https://user-images.githubusercontent.com/3995223/216365008-00bbac05-68db-4f28-b759-3f93158d9619.png">

I think it used to be default for a while probably as I seem to remember having the tab ids all over the place. Anyways, that is also an issue, since there's a break-of-flow when I do `<b>` and then instead of saying "2nd most recent", I need to look for the tab in the MRU list and locate its ID, and then type it in. So that's one extra downside of this method that I experience immediately.

### `:tab` back and forth
Actually I believe having the MRU idea in my mind polluted the initial feature request I was about to make. Instead of thinking of this path of traversal as MRU, it makes more sense to think of it as "undo", but instead of it being "change" history it would be a "navigation" history. Navigation, ofc, not in the context of the `window.location` of the current tab, but in the sense of the active tab in the firefox window.

[...]

PR #4595, opened 2023-02-14
Respect MRU order in tab completions and :tab command

treapster:
Related: #4580.
Do we need to do the same for :taball, or it's not used that much?


bovine3dom:
This looks great, thanks :)

We should improve the docs for `gt` and the `tabsort` setting.

I am not sure how I feel about the tab IDs being totally replaced with recency of access but that might be just because it is a big change. I think I might try to live with it a few days to see if I get used to it :)

treapster:
Well, theoretically we can create a separate setting like tabindex which will be either mru or default, but i think there is no reason to use or know real IDs when you use MRU sorting.

treapster:
That's right, take your time to test it out

PR #4616, opened 2023-03-06
urlincrement/decrement should operate on decodedURI Fix #3792

### LINKED ISSUE #3792: urlincrement/decrement should operate on decodedURI
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI

so e.g. `%20` doesn't change to `%21`

callmetushar123:
Have made the necessary changes to fix the issue

### DIFF
--- a/src/lib/url_util.ts
+++ b/src/lib/url_util.ts
@@ -11,8 +11,10 @@
  * @return          the incremented URL, or null if cannot be incremented
  */
 export function incrementUrl(url, count) {
+    const url_en=decodeURI(url)
+    const regex = /(.*?)(\d+)(\D*)$/
     // Find the final number in a URL
-    const matches = url.match(/(.*?)(\d+)(\D*)$/)
+    const matches = regex.exec(url_en)
 
     // no number in URL - nothing to do here
     if (matches === null) {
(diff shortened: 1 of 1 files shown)


callmetushar123:
@bovine3dom lint was throwing errors for using match() so I used exec() instead, so please have a look

callmetushar123:
Thanks a lot for reviewing. Will check out this blog, meanwhile I will try the other issue i am assigned too.🥳

bovine3dom:
That documentation is extremely out of date, sorry, and was never finished.

If you want to understand Tridactyl the best thing to do is to read the code. This line here

https://github.com/tridactyl/tridactyl/blob/12fcc25efebc354b80ebc0244ce54bb0329b5dc6/src/content.ts#L159

is probably the most important one - it is how Tridactyl listens for keyevents. If you keep following that function you should end up understanding most of the architecture.

callmetushar123:
Okay thanks, will do. Will it be better if you raise an issue and someone can work on the documentation?

bovine3dom:
Maybe we could do a pass deleting old documentation.

I'm reluctant to ask someone to write new documentation because it has a habit of getting out of date without anyone noticing

PR #4617, opened 2023-03-06
Kill some copypated code in completions

treapster:
Aside from moving copypasted code to the base class, i also had to move updateOptions in TabHistory to please eslint, as well as suppress unused var warnings for overrides. Also, i discovered that our typescript is rather old(3.9.10 while the latest is 4.9.5) and doesn't know the override keyword. What stops us from bumping it to 4-something? It may also help with the case we bumped into [here](https://github.com/tridactyl/tridactyl/pull/4532#issuecomment-1399129366) and probably some other things. 

As for the patch itself, it's probably not the best solution but it removes some 100+ lines of repetition so i hope we can call it improvement, even though it's still OOP which i heard is not very liked here:)

### DIFF
--- a/src/commandline_frame.ts
+++ b/src/commandline_frame.ts
@@ -143,7 +143,9 @@ export function enableCompletions() {
             .map(constructorr => {
                 try {
                     return new constructorr(commandline_state.completionsDiv)
-                } catch (e) {}
+                } catch (e) {
+                    console.error(e)
+                }
             })
             .filter(c => c)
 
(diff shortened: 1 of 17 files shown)


treapster:
I completely forgot that there are also overrides of CompletionSourceFuse::filter with the same copypasta. And there are too many methods which call each other - filter and updateOptions are doing the same thing and onInput just confuses by unnecessary indirection, and sometimes inInput calls filter, other times it's vice versa.. I'll try to sort it out when i have time

treapster:
Even though more testing is needed, it seems like the main mess is gone, and from the first glance it looks like completions are still working. Will test with more commands in the following days.

Issue #4622, opened 2023-03-10
Update to TypeScript 4


bovine3dom:
The main thing blocking this issue is that TypeDoc 0.20+ broke something (#4459).

We would also need to remove the last vestiges of `tslint`.

Presumably there are other things that would need updating too.

(Raised by @treapster in #4617)

treapster:
Judging by the logs, removing--ignoreCompilerErrors should fix the breakage.

bovine3dom:
We even waited long enough to upgrade that an equivalent option got added back in :) https://github.com/TypeStrong/typedoc/pull/1975

We need to update TypeScript to get that though.

bovine3dom:
https://github.com/TypeStrong/typedoc/blob/master/internal-docs/custom-themes.md I've had this tab open for months, presumably it's relevant somehow to what we need to do

See also https://github.com/tridactyl/tridactyl/pull/4624/commits/92c1563264b4b5f594b1d7240cf9524649cd5f13

I'll try to look at this again soon.

I remember it being so frustrating the last time I looked at it that I was put off working on Tridactyl for a while ... this time it will go better :muscle:

bovine3dom:
Migrated to TS6 (lol) in #5450

PR #4623, opened 2023-03-10
Remove remaining traces of ts-lint

bovine3dom:
Progress towards #4622.


bovine3dom:
Somewhat ironically we are now told that we should be using prettier https://github.com/typescript-eslint/tslint-to-eslint-config/blob/main/docs/FAQs.md#user-content-should-i-use-prettier when we moved away from it to `tslint` a few years ago...

PR #4710, opened 2023-05-23
WIP: add ex.next_completion [n] cmd

### LINKED ISSUE #4164: Commands and exmaps bindings for paging through completions
Currently, there are `ex.{prev,next}completion` bound to `S-<tab>` and `<tab>`. It would be nice to have commands (or a parameter for these existing commands) to page through completions faster. Some use cases include quickly browsing `:goto` on large pages and `:tab` for tab addicts.

bovine3dom:
Strange behaviour currently: does not select any completions unless kick-started with ex.next_completion 1, and then seems to jump a different number down each time. Tested on : completions

Once merged will close #4164

### DIFF
--- a/src/lib/commandline_cmds.ts
+++ b/src/lib/commandline_cmds.ts
@@ -38,27 +38,27 @@ export function getCommandlineFns(cmdline_state: {
         /**
          * Selects the next completion.
          */
-        next_completion: async () => {
+        next_completion: async (n = 1) => {
             await awaitProxyEq(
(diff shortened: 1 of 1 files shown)


bovine3dom:
CI reports

```
src/commandline_frame.ts:222:33 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'never'.

222                                 args.length === 0 ? undefined : args.join(" "),
```

PR #4712, opened 2023-05-23
Fix #4704: add --nowrap flag to tabnext/prev

### LINKED ISSUE #4704: feature request - add an option to prevent "tabnext"/"tabprev" wrapping around
I know `:tabnext` and `:tabprev` can wrap around between the last and the first tab, but it seems make more sense to disable this function as `j` and `k` would not jump back from the last to the first line in `Vim`. It can be more useful especially when the tabbar was hidden(for example, in fullscreen mode or `:guiset tabs autohide` was set), as I can roughly know my current position without checking with `:tab`/`:taball`.
Is it possible to add an option to prevent `:tabnext`/`:tabprev` wrapping around? Such as something like `:tabnext --no-wrap`...

bovine3dom:


bovine3dom:
```
/home/circleci/project/src/excmds.ts
  2619:24  error  Missing radix parameter  radix
```

bovine3dom:
@zerobikappa it looks like I never merged this so #4704 was never done 🤔

i probably just saw the failing CI, fixed it and then forgot to check it ever again

PR #4727, opened 2023-06-11
WIP: add keyup/down bind support

bovine3dom:
I hate rebasing

Supercedes #1645

---

Observations:

- ~~implicit keyups aren't being ignored, j leads to scrolling twice~~
- hint mode broken
- user binds broken (e.g. `:bind ,q fillcmdline test`)
- <Shift> sometimes follows links in hint modes(!?)
- adding a user bind of `<D-j>` binds to `<<UD?-<>D<U?-D>-<U?-->j<U?-j>>` (i.e. the key sequence literally typed out)
- hardcoded bind of `<D-e>` doesn't work


bovine3dom:
I think I need to look back through https://github.com/tridactyl/tridactyl/pull/4440/ and make sure that everywhere that changed things, keyup/keydown/repeat/optional key events work.

I also need to look back at #1645 and try to remember what it was I had decided about repeat keys - there is a `.repeat` property on keyboard events, maybe we need to update minimalKeyboardEvent to include that information?

I think I just deleted some logic about repeat that I probably shouldn't have deleted...

Issue #4657, opened 2023-04-06
:reader should respect user themes / be configurable


bovine3dom:
Might be as simple as including some CSS and/or adding some classnames

ujjwalpandey922:
Hi can I work on this?

bovine3dom:
Hi, you can work on whatever you like :)

I will be busy for the next week or so but if you have any questions ask and I'll do my best to answer them anyway.

ujjwalpandey922:
sure 
thanks

MojtabaSediqi:
is any body working on it?

hejops:
Would this involve editing `userContent.css` as in `guiset`, or can the CSS rule(s) just be loaded into memory somewhere?

bovine3dom:
Once we add CSS variables you should be able to use them just like the variables for other parts of Tridactyl, so `userContent.css` but more also with themes in `:colours`

bovine3dom:
in the grand tridactyl traditition of "things mostly get better when i personally want them", i kinda want a side-scrolling newspaper style multi-column layout for `:reader`

but i don't think it makes much sense to bundle that with `:colours`. so maybe we need a separate command? `:colours --reader`, perhaps?

PR #4748, opened 2023-07-29
Provide hook for basic CSS override for new Reader mode

natrys:
Top level body now has the fixed DOM id: `tridactyl-reader`. Article content is wrapped in a `main` tag.

This doesn't really solve #4657 with proper theming support, but does bare minimum to make it configurable at least. Now one can do something like this in their `userContent.css`:

```css
#tridactyl-reader > main {
  width: 80vw !important;
  text-align: left;
  font-family: Ubuntu;
}
```

Wrapped the article content in a `main` tag for determinism, I checked a few articles and none of them had that. Also got rid of `grid` for the body without messing up current layout, I think it's not intuitive to have to set width for the child content in the parent (body) unless you know grid is being used.

I admit to wanting to add article link to the title for purely personal reason. My [time tracking software](https://arbtt.nomeata.de/) relies on pattern matching on window name derived from title, the [firefox plugin](https://github.com/erichgoldman/add-url-to-window-title) that I use to add that automatically seems unable to do so for the new reader mode because of the url mangling that happens. Let me know if that's undesirable.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1290,6 +1290,11 @@ export class default_config {
      * https://fusejs.io/api/options.html#threshold
      */
     completionfuzziness = 0.3
+
+    /**
+     * Whether to show article url in the document.title of Reader View.
+     */
+    readerurlintitle: "true" | "false" = "false"
 }
 
 const platform_defaults = {
--- a/src/reader.ts
+++ b/src/reader.ts
@@ -1,14 +1,30 @@
-// import * as config from "@src/lib/config"
+import * as config from "@src/lib/config"
 import xss from "xss"
 
-function updatePage(){
-    const article = JSON.parse(decodeURIComponent(atob(window.location.hash.substr(1))))
-    article.content = xss(article.content, {stripIgnoreTag: true})
-    document.body.innerHTML = article.content
+async function updatePage() {
+    const article = JSON.parse(
+        decodeURIComponent(atob(window.location.hash.substr(1))),
+    )
+    article.content = xss(article.content, { stripIgnoreTag: true })
+    const content = document.createElement("main")
+    content.innerHTML = article.content
+    document.body.appendChild(content)
     if (article.title !== undefined) {
         const header = document.createElement("header")
         const title = document.createElement("h1")
-        document.title = [article.siteName, article.title].filter(Boolean).join(": ") // sensible?
+        if (
+            (await config.getAsync("readerurlintitle")) == "true" &&
+            !(article.title ?? "").includes(article.link)
+        ) {
+            document.title =
+                [article.siteName, article.title].filter(Boolean).join(": ") +
+                " :: " +
+                article.link
+        } else {
+            document.title = [article.siteName, article.title]
+                .filter(Boolean)
+                .join(": ")
+        }
         title.textContent = article.title
         header.appendChild(title)
         if (article.byline !== undefined) {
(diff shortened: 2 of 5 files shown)


src/reader.ts
```diff
@@ -4,11 +4,17 @@ import xss from "xss"
...
+    content.innerHTML = article.content
+    document.body.appendChild(content)
     if (article.title !== undefined) {
         const header = document.createElement("header")
         const title = document.createElement("h1")
-        document.title = [article.siteName, article.title].filter(Boolean).join(": ") // sensible?
+        if (article.link === undefined) {
+            document.title = [article.siteName, article.title].filter(Boolean).join(": ")
+        } else {
+            document.title = [article.siteName, article.title].filter(Boolean).join(": ") + " :: " + article.link
```
bovine3dom, line 16: Could we hide this behind a config setting by default? I don't think people would expect it

I need to check that the CSS changes don't change too much but otherwise this looks great :)

natrys:
@bovine3dom hid it behind a config setting, thanks. 

It turned out to be a little more complicated. Readability.js uses a bunch of heuristics to determine article title. Original document.title is usually a fallback, and even then it trims and cleans it up if necessary. If original document.title has link in it (in my case the other extension makes sure it does), even then readability.js usually doesn't include it in the returned article.title. But sometimes if it needs to fallback to document.title and it's too small even including the link, it just returns whole thing in article.title. So needs an extra check is to make sure to not append link if it's already included in article.title.

src/lib/config.ts
```diff
@@ -1290,6 +1290,11 @@ export class default_config {
      * https://fusejs.io/api/options.html#threshold
      */
     completionfuzziness = 0.3
+
+    /**
+     * Whether to show article url in the document.title of Reader View.
+     */
+    urlinreaderviewtitle: "true" | "false" = "false"
```
bovine3dom, line 1297: Tridactyl (mostly) has a pseudo-namespacing where the category/noun comes first then the verb, so `:tabopen` rather than `:opentab`. It makes it easier to find things with completions

for this I guess `readerurlintitle` would be ok?

bovine3dom:
Your `readerurlintitle` setting doesn't seem to work for me. Steps to reproduce:

1. `:set readerurlintitle true`
2. `:open https://www.bbc.com/news/world-europe-66362769`
3. `:reader`
4. No URL in window title. Expected behaviour: URL appears at end of window title with `:: ` prepended

natrys:
It works if you change `readerurlintitle` default value to true in the source, but not in the config or manually. Something weird is happening reconciling userconfig, but only when coming from `reader.ts`.

Here are my findings with default value false in source, but manually set to true later like you did:

- `console.log(config.get("readerurlintitle"))` is true everywhere in code, except inside `updatePage` function in `reader.ts`.
- The problem is that `const user = getDeepProperty(USERCONFIG, target)` call [here](https://github.com/tridactyl/tridactyl/blob/5cbfb728bd185535383bfb4e15966cfa2730b06d/src/lib/config.ts#L1820) usually returns true as it should, except when coming from `updatePage`, then it returns `undefined`.
- I put this in the `getDeepProperty()` to debug this:
```diff
diff --git a/src/lib/config.ts b/src/lib/config.ts
index e958fc8..d34ac0d 100644
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1707,6 +1707,9 @@ export const DEFAULTS = mergeDeepCull(
 export function getDeepProperty(obj, target: string[]) {
     if (obj !== undefined && obj !== null && target.length) {
         if (obj["🕷🕷INHERITS🕷🕷"] === undefined) {
+            if (target.length >= 1 && target[0] === "readerurlintitle") {
+                console.log(obj, obj[target[0]])
+            }
             return getDeepProperty(obj[target[0]], target.slice(1))
         } else {
             return getDeepProperty(
```
- The `obj` clearly has `readerurlintitle: true` key-value in it, and yet `obj[target[0]]` somehow shows `undefined`. Wtf?
- I put a breakpoint in the console.log, then run again. This time when debugger is hit, obj[target[0]]` actually shows true. And when I resume code, the code works as expected. Wtf?

Is there some kind of race condition here? I think I will need your help with this.

natrys:
That makes it work, I think.

bovine3dom:
Looks great, thanks! Sorry it took so long to merge :)

PR #4764, opened 2023-09-04
fix wrong direction for RTL language in reader mode

Monirzadeh:
tridactyl  show RTL language it LTR direction. this Pull Request fix that.


bovine3dom:
LGTM, thanks!

PR #4766, opened 2023-09-10
Add WIP session URL save/load

bovine3dom:
Todo:

- get rid of new tabs in each new window
- work out what to do with the old window
- docs

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3296,6 +3296,31 @@ export async function qall() {
     return Promise.all(windows.map(window => browser.windows.remove(window.id)))
 }
 
+//#background
+export async function sessionsave(...name: string[]){
+    let namestr = name.join(" ")
+    let tabs = await browser.tabs.query({})
+    let sessions = {}
+    for (let tab of tabs){
+        sessions[tab.windowId] = sessions[tab.windowId] || []
+        sessions[tab.windowId].push(tab.url)
+    }
+    let sessionstr = JSON.stringify(sessions)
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -3296,6 +3296,31 @@ export async function qall() {
...
+    let sessions = {}
+    for (let tab of tabs){
+        sessions[tab.windowId] = sessions[tab.windowId] || []
+        sessions[tab.windowId].push(tab.url)
+    }
+    let sessionstr = JSON.stringify(sessions)
+    return config.set("sessions", namestr, sessionstr)
+}
+
+//#background
```
pataquets, line 3312: Background/async? What if there are *lots* of tabs or there is heavy CPU pressure and Firefox is responding slow? Will the user get any feedback while session is being saved? When the session is saved?

src/excmds.ts
```diff
@@ -3296,6 +3296,31 @@ export async function qall() {
...
+export async function sessionsave(...name: string[]){
+    let namestr = name.join(" ")
+    let tabs = await browser.tabs.query({})
+    let sessions = {}
+    for (let tab of tabs){
+        sessions[tab.windowId] = sessions[tab.windowId] || []
+        sessions[tab.windowId].push(tab.url)
+    }
+    let sessionstr = JSON.stringify(sessions)
+    return config.set("sessions", namestr, sessionstr)
```
pataquets, line 3309: [docs] I assume existing session with same name will get overwritten without warning.
[docs] [command?]: How to list sessions?

PR #4776, opened 2023-10-07
Fixed the hint example for matching on the comment collapsers on Hacker News

hommels:
Hacker News added an additional class which prevents the example hint for matching on the Reddit/HN comment collapsers from matching correctly. By matching on any of the classes this seems to be fixed. I thought it might be nice to update the example in the documentation as well for anyone that stumbles upon it. 

I noticed that @bovine3dom made the same change in his example [.tridactylrc](https://github.com/tridactyl/tridactyl/blob/608744bc36fd0c8a3ad39ab94274952596f005eb/.tridactylrc#L44).

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -5123,7 +5123,7 @@ const KILL_STACK: Element[] = []
 
     Element selection flags:
         - -c [selector] hint links that match the css selector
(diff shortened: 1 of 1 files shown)


bovine3dom:
Thanks!

PR #4781, opened 2023-10-22
Add hint flag for specifying a selector that also keeps default hints

### LINKED ISSUE #4780: Unable to have a CSS hint in addition to default hints
I would like to add a custom CSS selector hint to default hints that are displayed when pressing `f`.
I've tried the following but the regular `f` hints are not displayed anymore:
```
bindurl www.imdb.com f hint -c .vertical-ellipsis
```
This is expected and not a bug. When a `-c` selector is provided, the default hints are not added:
https://github.com/tridactyl/tridactyl/blob/7f626ad4dde1d4d2f5ca414778c336f6de365276/src/lib/hint_util.ts#L406-L413

petoncle:
Fixes #4780.

@bovine3dom I am no good with names and chose `-C` for that new flag. Please fix :).

With this change, the Reddit comment collapser hints could be part of `f` (instead of having its own [custom shortcut](https://github.com/tridactyl/tridactyl/blob/850acdf2ca11018ed3c2fc7a8ae157d77f50bc17/src/excmds.ts#L5126)) like so:
```
bindurl www.reddit.com f hint -C .threadline
```
With that, when pressing `f` on Reddit you get both default hints and comment collapser hints.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -5125,6 +5125,7 @@ const KILL_STACK: Element[] = []
         - -c [selector] hint links that match the css selector
           - `bind ;c hint -c [class*="expand"],[class*="togg"]` works particularly well on reddit and HN
           - this works with most other hint modes, with the caveat that if other hint mode takes arguments your selector must contain no spaces, i.e. `hint -c[yourOtherFlag] [selector] [your other flag's arguments, which may contain spaces]`
+        - -C [selector] like `-c [selector]` but also hints all elements that would normally be hinted given the other options selected
         - -x [selector] exclude the matched elements from hinting
         - -f [text] hint links and inputs that display the given text
           - `bind <c-e> hint -f Edit`
--- a/src/lib/hint_util.ts
+++ b/src/lib/hint_util.ts
@@ -66,6 +66,7 @@ export class HintConfig implements HintOptions {
     public pipeAttribute = null
     public selectors = []
     public selectorsExclude = []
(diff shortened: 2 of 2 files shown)


src/excmds.ts
```diff
@@ -5125,6 +5125,7 @@ const KILL_STACK: Element[] = []
         - -c [selector] hint links that match the css selector
           - `bind ;c hint -c [class*="expand"],[class*="togg"]` works particularly well on reddit and HN
           - this works with most other hint modes, with the caveat that if other hint mode takes arguments your selector must contain no spaces, i.e. `hint -c[yourOtherFlag] [selector] [your other flag's arguments, which may contain spaces]`
+        - -C [selector] include default hints and links that match the css selector
```
bovine3dom, line 5128:
```suggestion
        - -C [selector] like `-c [selector]` but also hints all elements that would normally be hinted given the other options selected
```

bovine3dom:
Thanks for this PR! I think `-C` is probably fine. Do you really hate it?

petoncle:
> Thanks for this PR! I think `-C` is probably fine. Do you really hate it?

I'm fine with it if you are.

bovine3dom:
Works great, thanks for all your work on it!

I don't understand your comment about "bindurl www.reddit.com f hint -C .threadline" though - I don't see any .threadline elements on either old.reddit.com nor on www.reddit.com

PR #4809, opened 2023-10-31
Reduce likelihood of losing characters while command line is opening

petoncle:
Typing `tgithub` does not always result in the command line input (clInput) receiving `tabopen github`. Sometimes, the first few characters following the command end up missing (for example, clInput will contain `tabopen thub`). 

1. When typing `t` in the page, the page process (controller_content.ts) sends `fillcmdline t` to the command line process.
2. In the command line process, `fillcmdline()` (commandline_frame.ts) triggers the focusing of clInput.
3. Once clInput is focused it can start receiving native keyboard events via a regular keydown listener.

From 2. up until 3., the page has the focus and can receive keyboard events (if the user is fast enough and/or their machine is slow enough). Those events are lost since they are never passed to the command line process.

The proposed solution makes the page (controller_content.ts) start buffering typed characters as soon as `fillcmdline t` is sent to the command line process, up until clInput is focused (commandline_frame.ts). When clInput is focused, the command line process consumes the content of the buffer and places it into clInput.

Possible improvement ideas:
- buffer non-character key presses like backspace (from the page process), and manually process them in the command line process
- if clInput receives a native Enter key event that is going to execute the command, delay its processing until the page keys buffer has been consumed and the command is actually complete. Same for tab key.

### DIFF
--- a/src/content/controller_content.ts
+++ b/src/content/controller_content.ts
@@ -212,7 +212,7 @@ function* ParserController() {
 
                 if (response.exstr) {
                     exstr = response.exstr
-                    if (exstr.startsWith("fillcmdline")) { // TODO That's ugly. I need a way to know if this command is going to open the command line. Is there a better way?
+                    if (exstr === "fillcmdline" || exstr === "fillcmdline_notrail") {
                         logger.debug("Starting buffering of page keys")
                         bufferingPageKeysBeginTime = performance.now()
                         mustBufferPageKeysForClInput = true
--- a/src/content/controller_content.ts
+++ b/src/content/controller_content.ts
@@ -203,7 +203,7 @@ function* ParserController() {
 
                 if (response.exstr) {
                     exstr = response.exstr
-                    if (exstr.startsWith("fillcmdline ")) { // Ugh. That's ugly. I needed a way to know if this command is going to open the cmdline.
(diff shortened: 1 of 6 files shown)


bovine3dom:
This looks great. It will take me a while to review.

Mostly for my own reference: it looks like you haven't added any new key event listeners - you're only using existing ones. That's great because I don't have to worry about new security bugs :)

src/content/controller_content.ts
```diff
@@ -184,13 +207,17 @@ function* ParserController() {
...
 
                 if (response.isMatch && keyevent instanceof KeyboardEvent) {
-                    keyevent.preventDefault()
-                    keyevent.stopImmediatePropagation()
                     canceller.push(keyevent)
                 }
 
                 if (response.exstr) {
                     exstr = response.exstr
+                    if (exstr.startsWith("fillcmdline")) { // TODO That's ugly. I need a way to know if this command is going to open the command line. Is there a better way?
```
bovine3dom, line 215: We need to exclude _nofocus and _tmp because they're used for displaying messages and don't get focus.

I don't think there's a better way to know. We probably need to have some sort of function decorator...

petoncle: Indeed, so it's down to `fillcmdline` and `fillcmdline_notrail` only?

bovine3dom: Yeah, exactly. Otherwise this looks good. I'll probably release it in a beta in early January

Issue #4763, opened 2023-09-03
make ex strings starting with space skip History 


bovine3dom:
Probably just need to copy the logic for `:winopen -private`

PR #4820, opened 2023-11-04
Add historycmdcount* configs to limit cmdHistory size

GHolk:
Add config `historycmdcountalert` and `historycmdcountlimit` to limit `state.cmdHistory` size.

If `state.cmdHistory` size is larger then the historycmdcountalert when tridactyl start, tridactyl will reduce its size to the historycmdcountlimit.

The cmdHistory size check has a 15 seconds delay,
so the start up will not be to slow.

Feel free to change to a better config name or move the code snippet.

### DIFF
--- a/src/background.ts
+++ b/src/background.ts
@@ -66,6 +66,16 @@ controller.setExCmds({
     hint: HintingCmds,
 })
 
+// limit the command history size
+setTimeout(async () => {
+    const alert = await config.getAsync('historycmdcountalert')
+    const limit = await config.getAsync('historycmdcountlimit')
+    const cmdHistory = state.cmdHistory
+    if (cmdHistory.length > alert) {
+        state.cmdHistory = cmdHistory.slice(-limit)
+    }
+}, 15 * 1000)
+
 // {{{ tri.contentLocation
 // When loading the background, use the active tab to know what the current content url is
 browser.tabs.query({ currentWindow: true, active: true }).then(t => {
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1172,6 +1172,16 @@ export class default_config {
     historyresults = 50
 
     /**
+     * Reduce the size (line count) of ex command history if the size is larger than the historycmdcountalert when tridactyl started. The size will be reduce to the historycmdcountlimit.
+     */
+    historycmdcountalert = 768
+
+    /**
+     * Reduce the size (line count) of ex command history if the size is larger than the historycmdcountalert when tridactyl started. The size will be reduce to the historycmdcountlimit.
+     */
+    historycmdcountlimit = 512
+
+    /**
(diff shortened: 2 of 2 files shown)


GHolk:
Im not sure if there are issues about command history size.

By the way, I see #4763 so I resolve it in this branch, too.

src/lib/config.ts
```diff
@@ -1171,6 +1171,16 @@ export class default_config {
...
 
+    /**
+     * Reduce the size (line count) of ex command history if the size is larger than the historycmdcountalert when tridactyl started. The size will be reduce to the historycmdcountlimit.
+     */
+    historycmdcountalert = 768
+
+    /**
+     * Reduce the size (line count) of ex command history if the size is larger than the historycmdcountalert when tridactyl started. The size will be reduce to the historycmdcountlimit.
+     */
+    historycmdcountlimit = 512
```
bovine3dom, line 1182: I'd rather have one setting rather than two. Maybe remove 1/3 when it hits the limit.

512 seems pretty small, I'd set it to something much bigger like 100k. I think most of the time people are searching through it rather than just scrolling up

src/background.ts
```diff
@@ -66,6 +66,16 @@ controller.setExCmds({
     hint: HintingCmds,
 })
 
+// limit the command history size
+setTimeout(async () => {
+    const alert = await config.getAsync('historycmdcountalert')
+    const limit = await config.getAsync('historycmdcountlimit')
+    const cmdHistory = state.cmdHistory
+    if (cmdHistory.length > alert) {
```
bovine3dom, line 74: We should support -1 for unlimited

src/lib/commandline_cmds.ts
```diff
@@ -160,7 +160,8 @@ export function getCommandlineFns(cmdline_state: {
             // Save non-secret commandlines to the history.
             if (
                 !browser.extension.inIncognitoContext &&
-                !(func === "winopen" && args[0] === "-private")
+                !(func === "winopen" && args[0] === "-private") &&
```
bovine3dom, line 163: Maybe we should support zero to totally disable saving of history?

bovine3dom (commented):
This looks good. Have you experienced any bugs caused by big command history in practice?

GHolk:
> This looks good. Have you experienced any bugs caused by big command history in practice?

I can not remember exactly.
I add a TriStart hook to trim history commands long times ago, but I am not sure whether it became faster after that.

My tridactyl is not very fast since I have so many tabs.

I believe someone (maybe me?) asked about history command size issue but I can not find it.
Anywan, we can just put it here, and merge it if something bad happens someday.

src/lib/commandline_cmds.ts
GHolk: I does not like the idea that do `State.getAsync('historycmdcountlimit')` everytime before save ex command;
but we already do this while save history.

Issue #4840, opened 2023-11-15
`f` does not work on google searches


thomasaarholt:
Hello! I'm on tridactyl beta `1.23.0pre6867-ff696b5c` (commit appears to date to Nov 12, three days ago), and when [I search on google](https://www.google.com/search?client=firefox-b-d&q=tridactyl), I can use `j` and `k` to scroll, but `f` doesn't work.

This is on macOS 14, latest Firefox 119.01.

I thought #4765 would fix this, but it appears not.

bovine3dom:
Hi, those binds are in the exemplar RC file so if you used them you'll have to update them in your own RC file

The relevant lines are here https://github.com/tridactyl/tridactyl/commit/5635b91dee9ffe3f3191745ebda6e68ce9942596

thomasaarholt:
Maybe I've misunderstood, but should I not expect `f` to work "out-of-the-box" on google.com? If I search for anything on google and press `f`, nothing happens at all. On e.g. github I get highlighted boxes with characters in order to select links or buttons.

bovine3dom:
If you have removed your RC file, `:reseturl [the Google regex] f` should get the bind back to normal.

It does work out the box :) The RC file changes the bind so that only the search results are hinted. It's fragile and breaks every few months which is why it's hidden away in a file for advanced users

thomasaarholt:
Thank you! That worked! For any future viewers with the same problem: `:reseturl www.google.com f` worked. 

I struggled a little, expecting there to be a `reset config` kind of function in tridactyl. Is there such a function? I did observe that even though I deleted my `~/.tridactylrc` and restarted Firefox, `f` was still not working on google for me.

bovine3dom:
You're welcome!

`:sanitise tridactyllocal` wipes your entire Tridactyl configuration.

Tridactyl simply reads the `tridactylrc` file at startup and runs each line in it as if you typed them into Tridactyl manually. (RC stands for "run commands").

But Tridactyl has persistent storage of configuration, so if your RC file modifies your configuration, those changes will persist even if you remove or change the RC file.

It's counter-intuitive because most programs that have RC files do not have persistent configuration, so RC files feel like they are configuration.

I would like to make the persistence of configuration optional in Tridactyl so that RC files make more sense, but it would mean rewriting a fair bit of Tridactyl, so it's a low priority.

Closing this because your issue seems to be resolved :)

thomasaarholt:
[...]

PR #4848, opened 2023-11-20
Add configuration reset instructions to readme

thomasaarholt:
Following #4840, I thought I'd add a line to the readme to help future-me and people like him :)


bovine3dom:
Thanks for this. Not quite as simple as you're saying - there's unbind, unset, reset, URL variants of all of those and also sanitise.

We should add the word "reset" to the sanitise help and then it should appear with ":apropos reset", which is how I hope most people would search for it

PR #4879, opened 2023-12-26
Improve readability of fonts in midnight.css

debnath-d:
Font weight of 200 makes it hard to read. Default font weight makes it easier to read.

### DIFF
--- a/src/static/themes/midnight/midnight.css
+++ b/src/static/themes/midnight/midnight.css
@@ -62,7 +62,6 @@
 }
 
 :root #completions table {
-    font-weight: 200;
     table-layout: fixed;
     padding: 1rem;
     padding-top: 0;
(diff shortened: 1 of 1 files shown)


bovine3dom:
Sorry that this took an absurdly long time to reply to.

I agree with you but I don't use either of these themes (shydactyl has the same weight)

Can anyone who uses `midnight` or `shydactyl` please comment with their opinions about whether the font weight (the 'thickness' of the characters in the completions under the command line) should be increased?

debnath-d:
@bovine3dom Consider this PR now that someone has confirmed it?

bovine3dom:
I asked again on Discord but no-one replied :(

I'll add a variable so people can revert it easily and then merge this. Thanks!

Issue #4889, opened 2024-01-18
Incompatible usage of sed for tutorial generation build step on Darwin


chrishollandaise:
## Problem: 

Hi y'all! I started using Tridactly about a month ago, and love it! I wanted to start tinkering and contributing, but ran across this build issue during my onboarding process. I am building on MacOS.

It appears that the `scripts/make_tutorial.sh` script tries to make an in-place edit on line 19, however the the way the command is formed (`sed -i "s|\.md|.html|g" "$dest$fileroot.html"`) only supports GNU implementation of sed. The problem is that MacOS by default uses a BSD implementation which requires an extension argument.

## Solution:

This is a well-known problem, and I noticed that other `install.sh` script circumvents this by specifying the file extension when creating a backup file.

The easiest fix is to append `.bak` to -i flag, which I did and build ran successfully. This is better (i.e. fewer barriers to entry for onboarding darwin folks) than having MacOS contributors install a GNU compatible version of sed (i.e. gsed) and shadowing the default sed. I can go ahead and make a PR request for this unless I am missing something here.

## Steps to reproduce (on MacOS):

1.  Run through onboarding process outlined [here](https://github.com/tridactyl/tridactyl#building-and-installing)
2. Run `yarn run build`
3. Tridactyl doesn't build and successfully and outputs following sed errors:

```bash
yarn run v1.22.21
$ sh scripts/build.sh
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
sed: 1: "../../../generated/stat ...": invalid command code .
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

## System info:
-   Tridactyl version: 1.24.0
-   Operating system: Darwin 22.6.0 arm64

bovine3dom:
Hi, thanks for filing this issue.

[...]

PR #4892, opened 2024-01-20
Add additional sh lint check for incompatible sed -i cases

chrishollandaise:
## Changes
Fixes: #4889

Introduces additional warning lints if contributor uses only a GNU-compliant version of sed in-place command. An example of this warning as shown below (shows which scripts and line numbers).

SCR-20240120-mxxq

## Test cases
[Regexr test cases](https://regexr.com/7qntv)

## Additional Discussion
* Because linting is done on ubuntu-latest (which always used GNU grep), I made the decision to keep the command to find theses cases simple instead portable. [See this Stack Overflow thread for more information.](https://stackoverflow.com/a/6565519/9262657)
* Depending on future incompatibility problems with other coreutil commands, it is worth thinking about whether we want to have a solution that can do this sort of linting for us. Our current script linter `shellcheck` does not check for these sort of compatibility issues or even detects userspaces. [See this issue on shellcheck for more information](https://github.com/koalaman/shellcheck/issues/2902)
    * I will stipulate that this only becomes a real problem for scripts called on from `scripts/build.sh` so every time that yarn run build or any other alias runs this script. So efforts should probably be focused on just checking or minimizing the incompatibility cases.

### DIFF
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -5,5 +5,10 @@ if [ -x "$(command -v shellcheck)" ]; then
 else
 	echo "Warning: shellcheck is not installed, skipping shell scripts"
 fi
+incompatible_sed=$(! grep -RPn --color=always --exclude-dir=node_modules --exclude-dir=.git 'sed(?:\s-\w+)*\s-i(?!\S|\s"")')
+if [ "$incompatible_sed" ]; then
+	printf "\nWarning: avoid non-portable sed flag -i without backup extension, use -i.bak\n"
+	printf "%s\n\n" "$incompatible_sed"
+fi
 yarn run lint
 "$(yarn bin)/eslint" --rulesdir custom-eslint-rules --ext .ts .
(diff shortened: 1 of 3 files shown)


ci/lint.sh
```diff
@@ -5,5 +5,10 @@ if [ -x "$(command -v shellcheck)" ]; then
 else
 	echo "Warning: shellcheck is not installed, skipping shell scripts"
 fi
+incompatible_sed=$(! grep -RPn --color=always --exclude-dir=node_modules --exclude-dir=.git 'sed(?:\s-\w+)*\s-i(?!\S|\s"")')
```
chrishollandaise, line 8: If you check the test cases, this particular regex pattern won't support all edge cases, would it be worth covering those edges? It will require more, and I probably have to consider awk to cover everything

bovine3dom: I think it's probably fine as it stands; we won't be trying to trick it, it'll just be people using `sed -i` rather than `sed -i.bak`

PR #4896, opened 2024-01-26
hotfix: update web-ext 7.9.0 to 7.10.0

chrishollandaise:
## Problem

Running `yarn run run` with `web-ext 7.9.0`, results in error where Firefox is not able to launch on MacOS.

`Error: connect ECONNREFUSED 127.0.0.1:59544`

## Solution

[According to this issue on `mozilla/web-ext`](https://github.com/mozilla/web-ext/issues/2975#issuecomment-1883405462), the issue has be solved with version [`7.10.0`](https://github.com/mozilla/web-ext/releases/tag/7.10.0). This will be helpful update for MacOS contributors.


bovine3dom:
thanks for this. We also need package-lock.json to be updated or bad things will happen. If you don't get round to it, dependabot will eventually :)

bovine3dom:
I'm not convinced you've done the right thing - it should only have lines related to web-ext and it's dependencies changing. I can't remember how you get yarn to do it correctly...

chrishollandaise:
> I'm not convinced you've done the right thing - it should only have lines related to web-ext and its dependencies changing. I can't remember how you get yarn to do it correctly...

`yarn install --mode update-lockfile` could be this one? But, I noticed that you updated some other dependencies too? I'm pretty sure I had to merge those changes as well

chrishollandaise:
This one has [108 changes ](https://github.com/tridactyl/tridactyl/pull/4896/commits/1111196209b302ff30f3d349cfdf5903976f71e6). Appears to be web-ext deps before merge. I noticed that my formatter changed package.json, so I'll give it wait

bovine3dom:
Closing in favour of #4899 which hopefully will be less inexplicably broken :)

PR #4899, opened 2024-01-29
Bump web-ext to 7.10.0

bovine3dom:
Supersedes #4896 since that was failing e2e tests and diff looked messy


bovine3dom:
fwiw it was as simple as `yarn upgrade web-ext@^7.10.0`

PR #4910, opened 2024-02-01
Expose node-qrcode as global variable in text2qr command page

GHolk:
so user can do whatever they want with it.

web-driver test maybe because I does not install the native messanger.
ts-lint fail, I don't know how to fix it.

```
~/tridactyl/src/qrCodeGenerator.ts
  1:1  error  Definition for rule 'unsupported-apis' was not found  unsupported-apis

✖ 1 problem (1 error, 0 warnings)
```

### DIFF
--- a/src/qrCodeGenerator.ts
+++ b/src/qrCodeGenerator.ts
@@ -1,8 +1,9 @@
 import QRCode from "../vendor/qrcode/qrcode"
 import * as Logging from "@src/lib/logging"
-import {browserBg, activeTab} from "./lib/webext"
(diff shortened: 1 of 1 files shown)


bovine3dom:
Is there an easy we can attach it to tri instead? I think people are more likely to find it then.

Tests seem to be passing fine

GHolk:
You mean the `tri` in content script?
Maybe we can wrap dynamic import in an ex command;
I know little about tridactyl's bundler.

bovine3dom:
I did mean that, `tri` in content but what you're doing here is fine so I'm merging it :)

Issue #4106, opened 2022-02-15
Remove background color from hinted elements to keep original styling


Jeusto:
Tridactyl version: 1.22.0
Firefox version: Mozilla Firefox 97.0
Operating system: linux

Hi, I wanted to create my own theme for Tridactyl and I've stumbled upon a problem. In hint mode, I couldn't leave elements like buttons with their original background color. This is a bit annoying for some websites with a lot buttons. I've tried styling the class "TridactylHintElement" or the root variable "--tridactyl-hint-bg" but with no success. I have to either choose a custom background color or have no background color at all, I would like to keep the original background color for all hinted elements.

Here's how it looks and how I would like to make it look like.
![example1](https://user-images.githubusercontent.com/65134804/154115706-ed23d7cf-4a77-4f12-bead-bc8090bfaa9c.png)

![example2](https://user-images.githubusercontent.com/65134804/154115716-71eaa05a-64fe-4888-a1ae-ae93f4a56c22.png)

Also another unrelated thing: there's always a slight delay when opening new tabs where the default styling appears and then it changes to what it should be. Would it be possible to disable the css injection that the extension does, and leave it to the user to add the styling in the "userChrome.css" file?

bovine3dom:
I can't think of an obvious way of doing what you want, but maybe someone else will.


> there's always a slight delay when opening new tabs where the default styling appears and then it changes to what it should be. Would it be possible to disable the css injection that the extension does, and leave it to the user to add the styling in the "userChrome.css" file?


See https://github.com/tridactyl/tridactyl/issues/2510#issuecomment-763198138 - I think that solves your problem :)

davidscotson:
I don't think this is possible to do from a theme with the way the current core themeing code works.

Because we use CSS vars, it's impossible to set a broken or null value, it'll just default to setting the background to transparent.

If it was the color rather than the background-color we wanted, then using 'currentColor' would work, but we only are able to override, and I don't think there's any way to get the current background color (see https://github.com/w3c/csswg-drafts/issues/5292)

So probably the core theme code needs tweaked in some way to make this possible.

I was looking at using ::after to create overlays rather than directly affecting the page content when styling hints, but styling injected code on a random webpage is pretty wild, seems there's lots of gotchas to look out for. But I'll report back if it seems to be working out, as it gives some extra flexability.

Jeusto:
[...]

PR #4919, opened 2024-02-07
Fix issue #4106

robbilky:
I guess this is a bit hacky, but removing the background colour under hint.css will make it easier for people to create their own themes should they want to disable element highlighting and keep the original background colour of the element.

I added the element highlighting to the existing themes instead. So the authors of those themes preferences are still preserved. I added a comment so anyone editing an existing theme knows to delete that section if they don't want element highlighting disabled.


bovine3dom:
Thanks for this PR - sorry that I never got around to looking at it seriously (other than seeing it was a bit hacky and thinking that I needed to think about it :) )

Superseded and solved quite nicely by #5097

PR #4924, opened 2024-02-16
Make short hint name generation more precise

dzhu:
The previous method skipped too many or too few hint names in some cases; the new one should skip the optimal number in all cases.

If the number of hints was equal to the number of hint characters, then a hint name would be skipped unnecessarily. (In a way, that wasn't so bad: the skipped hint name would be the first hint character and the extra last hint name would then be the same character doubled, so typing the first character once would select it. But if the hint character was also a valid Tridactyl binding, then the full two-character hint name would also trigger that binding.)

There were also (infinitely) many cases where too few hint names would be skipped, leading to prefixes among the hint names. For example, if the hint characters were "abc" and there were 8 hints, the previous method would skip only 2 hint names, leading to "c" and "ca" both being generated as hint names.

The new method correctly skips 0 and 3 hint names in those cases.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -601,16 +601,24 @@ function* hintnames_simple(
 
 /** Shorter hints
 
-    Hints that are prefixes of other hints are a bit annoying because you have to select them with Enter or Space.
+    Hints that are prefixes of other hints are a bit annoying because you have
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM, thanks!

PR #4930, opened 2024-02-22
prevent reader from being called more than once

### LINKED ISSUE #4920: reader can be called multiple times
-   Brief description of the problem:

The new tridactyl `reader` can be called multiple times, each time adding an extra entry to the navigation history (and an extra header). This is inconsistent with the native reader, which is just a toggle (so you can only go one back or forward).

-   Steps to reproduce:

1. `:reader`
1. `:reader`
1. repeat...

-   Tridactyl version (`:version`): 1.24.0

-   Firefox version (Top right menu > Help > About Firefox): 122.0.1

-   URL of the website the bug happens on: anywhere `reader` can be called

hejops:
fixes #4920

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -6145,7 +6145,7 @@ export async function readerurl() {
  */
 //#content
 export async function reader(...args: string[]) {
-    if (window.location.href.includes("static/reader.html#")) return
+    if (window.location.href.includes("static/reader.html#")) return back()
     switch (args[0]) {
         case "--old":
             readerold()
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -6145,7 +6145,8 @@ export async function readerurl() {
  */
 //#content
 export async function reader(...args: string[]) {
-    switch(args[0]) {
+    if (window.location.href.includes("static/reader.html#")) return back()
```
bovine3dom, line 6148: @hejops do you have any objections to doing this instead? it seems to work fine in my limited testing

hejops: Oh yeah, that would simulate `readerold` more accurately. Didn't notice there was a `back` function, that's all.

PR #4989, opened 2024-06-01
wip messing around with sentence similarity

### LINKED ISSUE #936: Beyond fuzzy search
Fuzzy search is a bit pants, really.

It'd be pretty cool if we could use some newfangled machine learning model in JavaScript so that when I'm looking through my history I can find results when I slightly misremember a word, e.g. "All the king's mules". Essentially, we want a metric on sentence fragments.

I know these models exist but I'm struggling to find them right now.

bovine3dom:
Closes #936 if ever finished

thoughts:

1. slightly slow to embed docs (~30 seconds) so need to cache between sessions
2. probably too slow to do an embedding as you type but worth checking
3. results often follow the pattern `irrelevant, bizarre, totally irrelevant, THING YOU ARE LOOKING FOR, irrelvant, kooky` ... but given that `:apropos` for the same pattern often gives zero results, maybe that's useful. example: searching for "version control" gives you lots of random stuff but ~5th among them is mktridactylrc which is what you probably want. Which you might not have found otherwise
5. should try on tab urls / title and see what happens

### DIFF
--- a/scratch.js
+++ b/scratch.js
@@ -1,0 +1,50 @@
+const extractor = await tri.pipeline('feature-extraction', 'Supabase/gte-small')  
+// const extractor = await tri.pipeline('feature-extraction', 'Xenova/all-MiniLM-L6-v2')  
+const sentences = ['This is an example sentence', 'Each sentence is converted']
+const output = await extractor(sentences, { pooling: 'mean', normalize: true })
+
+let probe = 'sample'
+let outputProbe = await extractor([probe], { pooling: 'mean', normalize: true })
+
+// adapted from https://github.com/xenova/transformers.js/blob/da2688626d7812ad1ea47fd304c2072cc685051b/examples/semantic-image-search-client/src/app/worker.js#L46
+function cosineSimilarity(query_embeds, database_embeds) {
+    const numDB = database_embeds.dims[0]
+    const EMBED_DIM = database_embeds.dims[1]
+    const similarityScores = new Array(numDB)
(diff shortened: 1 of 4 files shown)


scratch.js
```diff
@@ -0,0 +1,50 @@
...
+
+let sims = cosineSimilarity(outputProbe, output)
+
+Array.from(sentences.entries()).sort((l, r) => sims[l[0]] < sims[r[0]])
+
+const funcs = tri.metadata.everything.getFile("src/excmds.ts").getFunctions().filter(f => !f[1].hidden && f[1].doc.length > 0)
+
+const docEmbeds = await extractor(funcs.map(f => f[0] + ": " + f[1].doc.slice(0, 512)), { pooling: 'mean', normalize: true }) // need to check max length. takes a few seconds
+
+probe = 'easter egg'
```
bovine3dom, line 47: results: `"hint", "wintitle", "snow_mouse_mode", "neo_mouse_mode", "escapehatch", "back", "markjumplocal", "yankimage", "markjump", "forward"`

for "version control": `"issue", "credits", "updatecheck", "nativeinstall", "updatenative", "containerupdate", "mktridactylrc", "firefoxsyncpull", "setnull", "markaddlocal"`

bovine3dom:
Worth checking if we can speed up search with a nearest neighbour index https://unum-cloud.github.io/usearch/javascript/

I think the slowness is in the embedding of the search term but I could be wrong

PR #4995, opened 2024-06-05
added support for the colemak keyboard layout

zorzem32:
greetings this patch adds support for the colemak keyboard layout


bovine3dom:
LGTM, thanks!

PR #4990, opened 2024-06-01
Fix #4964, maybe: switch to mozdownload

bovine3dom:

PR #4998, opened 2024-06-06
Interpret addresses ending with / as URLs

alanpearce:
related to #1002, specifically https://github.com/tridactyl/tridactyl/issues/1002#issuecomment-932990365.

I don't actually know if the test passes because the e2e tests don't seem to do anything with the browser after opening it (starting from `:rssexec works`)

### DIFF
--- a/src/lib/webext.ts
+++ b/src/lib/webext.ts
@@ -88,7 +88,7 @@ export async function activeTabOnWindow(windowId?: number) {
     return (
         await browser.tabs.query({
             windowId,
-            active: true
+            active: true,
         })
     )[0]
 }
@@ -186,7 +186,13 @@ export async function openInNewTab(
     url: string,
 
     // NB: defaults are actually enforced just below
-    kwargs: { active?; related?; cookieStoreId?; bypassFocusHack?; discarded? } = {
+    kwargs: {
+        active?
+        related?
+        cookieStoreId?
+        bypassFocusHack?
+        discarded?
+    } = {
         active: true,
         related: false,
         cookieStoreId: undefined,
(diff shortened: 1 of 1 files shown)


e2e_tests/e2e.test.ts
```diff
@@ -271,6 +271,16 @@ describe("webdriver", () => {
...
+    test("`:tabopen foo/<CR>` opens http://foo.", async () => {
+        const driver = await getDriver()
+        return newTabWithoutChangingOldTabs(driver, async () => {
+            await sendKeys(driver, ":tabopen foo/<CR>")
+        }).then(async ([newtab, _]) => {
+            expect(newtab.active).toEqual(true)
+            await driver.wait(untilTabUrlMatches(driver, newtab.id, "http://foo"), 10000)
+        }).finally(() => killDriver(driver))
+    })
+
```
bovine3dom, line 283: Thanks for this PR! I thought I had added this feature years ago :)


Annoyingly (as you discovered) the end-to-end tests are extremely fragile. At the moment they seem totally broken, but even when they aren't, they often take 10-20 attempts to succeed on CI. It gets worse with every test we add. So please don't add any new E2E tests :)

You should be able to add a unit test by making a file with the extension `.test.ts`, which should automatically swap `browser` out for a fake one via this package https://github.com/clarkbw/jest-webextension-mock - but we don't seem to be using it anywhere else so it's possible that it doesn't work.

From my perspective, I don't really mind if the code doesn't have a test. But I understand I am a bit of a heretic for having such views.

alanpearce: I removed the e2e test.

I tried to add some unit tests, but it seems like the test environment isn't mocking enough to allow `webext.ts` to be included (`getContext` is easy enough, but I stopped at `this.addEventListener.bind` because it doesn't seem to be part of the source).  
Maybe moving `queryAndURLwrangler` to a different file would allow it to work.  I could work on that in a follow-up PR, if you'd like.  I have the impression the function could be refactored and unit tests would be a nice safety net for anyone working on that; otherwise I'm normally also a bit of a heretic.

bovine3dom: I don't think moving it to a different file is worth your time, but if you want to do it, go ahead.

If you're in the mood to fight with tests, I'd much rather know why the e2e tests fail all the time at `:rssexec`. I had a go looking at it in #4990

PR #5037, opened 2024-09-02
Prevent Non-HTML Content-Type inject iframe

michael2to3:
Added a check to ensure iframes are only injected when `Content-Type` is `text/html`. This prevents Firefox from rendering `text/xml` files as HTML.
Before:
image
After:
image


src/content/commandline_content.ts
```diff
@@ -35,6 +35,12 @@ let enabled = false
 async function init() {
     const noiframe = await config.getAsync("noiframe")
     const notridactyl = await config.getAsync("superignore")
+
+    if (document.contentType !== "text/html") {
```
bovine3dom, line 39: Thanks for this PR! I've finally got around to reviewing it and I've noticed that it stops us injecting the iframe into other URLs like images e.g. https://upload.wikimedia.org/wikipedia/commons/2/25/Siam_lilacpoint.jpg where currently you can access the command line.

I think it is probably better to have a blacklist instead - are there any other content types you think we shouldn't inject into?

michael2to3: I tried executing an XSS attack using non-printable characters, but modern browsers block such methods. In theory, if no encoding is specified and part of the content is controlled by the user, iframe injection could lead to JS execution when browsers render it as HTML. However, after spending significant time attempting the attack, Firefox halts XML processing after an ESC character. The following POC didn’t work, but perhaps someone else can achieve it. I recommend blocking iframe insertion where XML content is rendered by checking `document.contentType.includes("xml")`. For further insight into encoding issues, refer to this [SonarSource research](https://www.sonarsource.com/blog/encoding-differentials-why-charset-matters/).

Here’s the POC code:

```python
from flask import Flask, Response
from markupsafe import escape

app = Flask(__name__)

def jis_encode(text):
    return text.encode("iso2022_jp")

@app.route("/get-xml")
def get_xml():
    escape_sequence = "\x1b$@ />"
    descape_sequence = "\x1b(B"
    
    malicious_script = '<x:script xmlns:x="http://www.w3.org/1999/xhtml">alert(document.domain)</x:script>'
    
    payload = escape_sequence + descape_sequence + malicious_script
    safe_payload = escape(payload)
    
    xml_template = f"""<?xml version="1.0"?>
<m:EntityDescriptor m:xmlns="http://www.w3.org/2000/09/xmldsig#" xmlns:m="http://www.w3.org/1999/xhtml#">
  <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
    <SignedInfo>
      <CanonicalizationMethod Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
      <SignatureMethod Algorithm="http://www.w3.org/2001/04/xmldsig-more#rsa-sha256"/>
      <Reference>
        <Transforms>
          <Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
          <Transform Algorithm="http://www.w3.org/2001/10/xml-exc-c14n#"/>
        </Transforms>
        <DigestMethod Algorithm="{safe_payload}"/>
        <DigestValue></DigestValue>
      </Reference>
    </SignedInfo>
    <KeyInfo>
      <X509Data>
      </X509Data>
    </KeyInfo>
  </Signature>
  <m:SPSSODescriptor protocolSupportEnumeration="urn:oasis:names:tc:SAML:2.0:protocol" WantAssertionsSigned="true">
    <m:KeyDescriptor use="signing">
      <KeyInfo xmlns="http://www.w3.org/2000/09/xmldsig#">
        <X509Data>
        </X509Data>
      </KeyInfo>
    </m:KeyDescriptor>
    <m:NameIDFormat>urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified</m:NameIDFormat>
  </m:SPSSODescriptor>
  <m:ContactPerson contactType="administrative">
    <m:Company>test</m:Company>
    <m:GivenName>test</m:GivenName>
  </m:ContactPerson>
</m:EntityDescriptor>"""
    
    encoded_xml = jis_encode(xml_template)
    
    response = Response(encoded_xml)
    response.headers["Content-Type"] = "text/xml"
    return response

if __name__ == "__main__":
    app.run(debug=True)
```

To block iframe injection into XML-rendered content:

```javascript
if (document.contentType.includes("xml")) {
    console.warn("XML content detected, blocking iframe insertion.");
    return;
}
```

bovine3dom: Thanks for looking into it. I _think_ if we block injection on XML content types like you're suggesting we should be good, right?

michael2to3: Yes, I think we're good now. I've updated my commit

src/content/commandline_content.ts
```diff
@@ -35,6 +35,12 @@ let enabled = false
 async function init() {
     const noiframe = await config.getAsync("noiframe")
     const notridactyl = await config.getAsync("superignore")
+
+    if (document.contentType.includes("xml")) {
+        logger.info("Content type is not text/html; aborting iframe injection.")
```
bovine3dom, line 40:
```suggestion
        logger.info("Content type is xml; aborting iframe injection.")
```

PR #5056, opened 2024-10-06
Adds a Flake + Dev Shell.

cooperpellaton:
This makes onboarding easier for folks who have Nix installed! 

This PR provides a [`devshell`](https://github.com/numtide/devshell) made easily accessible via [`nix-direnv`](https://github.com/nix-community/nix-direnv). This means the dependencies are immutably managed and can be sourced when cd'ing into the directory. For some context you can see read [this](https://fasterthanli.me/series/building-a-rust-service-with-nix/part-10) post. 

A natural next step would be to source CI from this as well ensuring the same environment from developer to CI/CD.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+if ! has nix_direnv_version || ! nix_direnv_version 3.0.6; then
+	source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.6/direnvrc" "sha256-RYcUJaRMf8oF5LznDrlCXbkOQrywm0HDv1VjYGaJGdM="
+fi
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ compiler/**/*.js
 .DS_Store
 .build_cache/
 yarn-error.log
+.direnv/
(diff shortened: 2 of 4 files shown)


bovine3dom:
Hi, thanks for this, it's an interesting idea. I often find myself fighting with node versions.

The trouble I foresee is that lockfiles are only useful if everyone uses them. Would it still be useful to you if we added flake.lock to gitignore?

cooperpellaton:
While it is possible to forego the `flake.lock` it would make it less reproducible as the version of the dependencies would depend on the contents of the [packages](https://search.nixos.org/packages) at the time the environment was sourced.

If there's interest I could look at moving to something higher level that still uses nix under the hood and is similarly easy to use like [`devenv`](https://devenv.sh) which I believe hides the lock file.

bovine3dom:
Yeah, I get that without flake.lock it stops being reproducible. My argument is that if ~80% of people don't use the lockfile, it isn't reproducible anyway, so we'd be lying to the people who were using it.

If we fix that with some "easy" tool and make it mandatory, we're adding friction for new contributors and that might reduce the number of them. We're already a bit weird in requiring yarn rather than npm.

I guess if there's a tool we can add that's bundled by yarn I would be ok with it?

PR #5058, opened 2024-10-08
Fix CI Pipeline & E2E Tests

cooperpellaton:
This PR fixes the CI pipeline. ~~It stacks on top of https://github.com/tridactyl/tridactyl/pull/5056.~~

To achieve this I:
* Changed how the extension is installed. It's now done using `installAddOn`.
* Changed how Firefox is installed in CI by using a native Github action.
* Cleaned up the `utils` and `e2e.test` so that they use `async`/`await` for readability and have more readable logic.
* Fixed the flaky test calling `qwant` which is not registered as a search provider on the master branch.

The tests are now stable which lets me cut the retries down and help the pipeline execute more quickly!

### DIFF
--- a/e2e_tests/e2e.test.ts
+++ b/e2e_tests/e2e.test.ts
@@ -1,12 +1,15 @@
-import * as process from "process"
-const env = process.env
-import * as fs from "fs"
+import { promises as fs } from "fs"
 import * as os from "os"
 import * as path from "path"
-import { Browser, Builder, By } from "selenium-webdriver"
-import { Options, Driver } from "selenium-webdriver/firefox"
+import { By } from "selenium-webdriver"
+import { Driver } from "selenium-webdriver/firefox"
 import * as Until from "selenium-webdriver/lib/until"
-import { getNewestFileIn, sendKeys } from "./utils"
+import {
+    getDriver,
+    getDriverAndProfileDirs,
+    iframeLoaded,
+    sendKeys,
+} from "./utils"
 
 jest.setTimeout(100000)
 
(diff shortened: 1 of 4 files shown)


e2e_tests/e2e.test.ts
```diff
@@ -230,108 +299,54 @@ describe("webdriver", () => {
...
-
-//     test("`:tabopen -b -c work search qwant<CR>` opens about:blank in a container.", async () => {
-//         const driver = await getDriver()
-//         return newTabWithoutChangingOldTabs(driver, async () => {
-//             await sendKeys(driver, ":tabopen -b -c work search qwant<CR>")
-//         }).then(async ([newtab, _]) => {
-//             expect(newtab.active).toEqual(false)
-//             expect(newtab.cookieStoreId).toMatch("firefox-container-")
-//             await driver.wait(untilTabUrlMatches(driver, newtab.id, new RegExp("^https://www.google.com/search.*qwant")))
-//         }).finally(() => killDriver(driver))
```
bovine3dom, line 333: If the tests are less flakey now it'd be good to re-enable these three since they test quite complicated bits of Tridactyl

bovine3dom (commented):
wow, thanks so much for looking at this. These tests have been flakey and unbearably slow for about five years.

I couldn't believe the tests were actually running this quickly so I broke them on purpose and it seems they are https://github.com/tridactyl/tridactyl/actions/runs/11231559694/job/31221251800 :exploding_head: 

Could you take out the flake related files so we can talk about them in #5056 and then squash the commits? Happy to do it for you if you like.

Thanks again for this. It's like Christmas has come early :heart:

cooperpellaton:
I've rebased so the commits from the other PR are no longer contained. I've also added the other tests back!

e2e_tests/e2e.test.ts
cooperpellaton: Added.

bovine3dom:
it looks like the tests are flakey again :(

the errors about quit() are a little odd

busy for the next few days so not sure I'll get a chance to look further at this till the weekend

cooperpellaton:
I'll keep looking into it and see what I can find.

bovine3dom:
Thinking about this maybe we should have a different number of retries for master vs pull requests since P(real failure | failed attempt && pull) >>  P(real failure | failed attempt && daily master)

PR #5060, opened 2024-10-11
Switch to bun

bovine3dom:
speedup is ~10x for no-ops and 1.5x for rebuilds, probably worth switching?

I think if we're switching to something that most people don't have we might as well go the whole hog and switch to `devenv` like was suggested in #5056 

It would probably stop the annoying problems I get every year or so trying to update `node` on the buildbot too

:thinking: 

https://stackoverflow.com/questions/74810175/migrate-yarn-lock-dependencies-to-bun-lockb-file/77302434#77302434

```
# convert yarn.lock to package-lock.json using synp
npx synp --source-file yarn.lock
 
# upgrade package-lock.json to version 3 using npm
npm i --lockfile-version 3 --frozen-lockfile
 
# convert package-lock.json to bun.lockb using bun
bun pm migrate
```


cooperpellaton:
I'll give this a look and help you out.

cooperpellaton:
Can you rebase on `master` so I can open a PR from this [branch](https://github.com/cooperpellaton/tridactyl/pull/3) to this PR? I got it all working 😄

cooperpellaton:
I'll work on moving directly to `bun` from `esbuild` tomorrow. 

https://bun.sh/docs/bundler/vs-esbuild

bovine3dom:
FWIW I'm reasonably happy with esbuild, it's already quick enough. The speedup to rebuild I'm talking about is just from running esbuild with bun rather than yarn.

My hope with switching to bun is that it will speed up switching branches where `yarn install` needs to be called with the git hook

But I need to balance that against making it annoying to cross the bun/yarn boundary in history when doing e.g. `git bisect` to hunt down a bug

cooperpellaton:
I understand, though I do think that leaning into the idiomatic way of doing things from a tool is generally favorable and in this case removes a dependency. I've got that working in my branch if you want to check it out!

PR #5073, opened 2024-10-24
fix switching focus in the same shadow root does not send focusin event outside

GHolk:
If a shadow root has 2 textareas inside,
switch focus between them will not dispatch focusin event on the root document object.
The focusin event will only happen on the host.shadowRoot .

This patch attach focusin event listener recursively to all the focused shadow root,
and set the focused element properly.

### DIFF
--- a/src/lib/dom.ts
+++ b/src/lib/dom.ts
@@ -541,7 +541,9 @@ export function getLastUsedInput(): HTMLElement {
  *  https://bugzilla.mozilla.org/show_bug.cgi?id=1406825
  * */
 function onPageFocus(elem: HTMLElement): boolean {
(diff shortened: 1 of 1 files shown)


GHolk:
One can test this with one of my bookmarklet:
https://gholk.github.io/bookmarklet/#prompt-annotate-description
(just click it the highlight hyperlink in the page if you trust me)

This bookmarklet create a shadow-root with two textbox inside.
without this patch, switching focus between them does not change the `LAST_USED_INPUT`,
and commands like  `text.forward_char` will not work properly.

bovine3dom:
Works great, thanks!

PR #5084, opened 2024-11-05
Do not parse numbers as IPs

### LINKED ISSUE #5081: Counter-intuitive behavior when open'ing an integer (integer gets converted to IP address)
-   Brief description of the problem / repro:

`open 538` opens `https://0.0.2.26/` while I would expect it to open my default search engine with `538` as the query.

When the integer is large enough, the behavior is correct, e.g. `open 2549536629329` does opens my default search engine with `2549536629329` as the query.

-   Tridactyl version (`:version`): `1.24.1`

-   Firefox version (Top right menu > Help > About Firefox): `129.0.1`

-   Contents of ~/.tridactylrc or ~/.config/tridactyl/tridactylrc (if they exist): https://github.com/Alexis-D/.dotfiles/blob/master/.tridactylrc

Hypothesis:

`open` delegates to `openInTab` (line 1630 which GH doesn't render in the snippet below):

https://github.com/tridactyl/tridactyl/blob/bd23136c142e118c399ff29cca2761ab01b0eb93/src/excmds.ts#L1618-L1632

`openInTab` calls `queryAndURLwrangler`:

https://github.com/tridactyl/tridactyl/blob/bd23136c142e118c399ff29cca2761ab01b0eb93/src/lib/webext.ts#L398-L399

Which in turn has this logic:

https://github.com/tridactyl/tridactyl/blob/bd23136c142e118c399ff29cca2761ab01b0eb93/src/lib/webext.ts#L359-L366

Which seems to be the culprit:

```
>> new URL("http://538")
URL { href: "http://0.0.2.26/", origin: "http://0.0.2.26/", protocol: "http:", username: "", password: "", host: "0.0.2.26", hostname: "0.0.2.26", port: "", pathname: "/", search: "" }

>> new URL("http://2549536629329")
Uncaught TypeError: URL constructor: http://2549536629329 is not a valid URL.
    <anonymous> debugger eval code:1
```

Proposed behavior/fix:

Naively (?) I would assume that folks are more likely to expect `open [0-9]+` to open their search engine rather than rely on a somewhat obscure integer -> IP conversion. Instead we could check `address.match(/^\d+$/)` and if that's the case skip the `Maybe it's a domain without protocol ` logic in `queryAndURLwrangler`.

Issues I can think of with changing the current behavior (there might be more!):

* It's a breaking change if my assumption that no-one relies on this is incorrect ([xkcd – workflow](https://xkcd.com/1172/))
* On the off-chance someone has a search engine that matches `^\d+$` then this should probably be honored first

Alexis-D:
Fix #5081

The fix itself is in 5d24ab72d85fa2a01ae0f5734e98b4b38f278c41. I tested it locally and can confirm the following work as expected (the first two open my search engine, the third one goes to `http://1.1.1.1`):

```
:open 538
:open 3.14
:open 1.1.1.1
```

There are a couple unrelated commits on this branch because things didn't work out-of-the-box for me, happy to split those into their own PR (or remove them altogether), just lmk.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -59,16 +59,15 @@ If you want to use advanced features such as edit-in-Vim, you'll also need to in
 
 **Snap and Flatpak:** Native Messaging support here is fairly recent and may require:
 
-* Upgrading to a beta version of Firefox (`>= 106.0b6`)
-* Enabling webextension permissions: `flatpak permission-set webextensions tridactyl snap.firefox yes`
-* Rebooting your system (and likely nothing short of it)
+-   Upgrading to a beta version of Firefox (`>= 106.0b6`)
+-   Enabling webextension permissions: `flatpak permission-set webextensions tridactyl snap.firefox yes`
+-   Rebooting your system (and likely nothing short of it)
 
 See [this call for testing thread](https://discourse.ubuntu.com/t/call-for-testing-native-messaging-support-in-the-firefox-snap/29759) and [this PR](https://github.com/tridactyl/tridactyl/pull/4406) for more details and troubleshooting tips.
 
 **Firejail** will require explicit path whitelisting, but should be feasible based on https://github.com/netblue30/firejail/issues/2109.
 
-For other containerized installs, see troubleshooting steps in https://github.com/tridactyl/tridactyl/issues/2406 and the links above.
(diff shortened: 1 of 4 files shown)


bovine3dom:
LGTM, thanks!

PR #5097, opened 2024-11-30
Style Hints According to Config

Chic-Tweetz:
As discussed in https://github.com/tridactyl/tridactyl/issues/4106 and others, hint element styles can't be completely disabled in themes.
This adds a config key, `hintstyles` so that styles are only added if they are enabled, otherwise the original page style remains intact.
Outlines, backgrounds and font colour can individually be enabled/disabled, or only added to active hint using `set hintstyles.bg`, `hintstyles.outline` and `hintstyles.fg`. Values can be `all`, `active`, or `none`

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -95,12 +95,10 @@ export class default_config {
                 F: "hint -bc [tabindex]:not(.two)>div,a",
             },
         },
-        "^https://teams.microsoft.com": {
-            // #5054
+        "^https://teams.microsoft.com": { // #5054
             superignore: "true",
         },
-        "^https://teams.live.com": {
-            // #5054
+        "^https://teams.live.com": { // #5054
             superignore: "true",
         },
     }
@@ -1788,7 +1786,7 @@ export const keyboardlayouts = {
         KeyY: ["", "!"],
     },
     workman: {
-        Quote: ["'", '"'],
+        Quote: ["'", "\""],
         Digit8: ["8", "*"],
         Digit1: ["1", "!"],
         Digit2: ["2", "@"],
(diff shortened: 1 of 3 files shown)


bovine3dom:
this is a neat solution, thanks! I need to find some time to check through it to make sure e.g. it isn't breaking people's own custom themes as far as possible. probably this weekend.

Chic-Tweetz:
On that point, all three default settings should be "all" in config.ts rather than "active" to keep themes as they are now.

src/lib/config.ts
```diff
@@ -1334,6 +1334,15 @@ export class default_config {
      * Whether to show article url in the document.title of Reader View.
      */
     readerurlintitle: "true" | "false" = "false"
+
+    /**
+     *  Which css styles to add for hint elements.
+     **/
+    hintstyles: { [key: string]: "all" | "active" | "none" } = {
+        fg: "active",
+        bg: "active",
```
bovine3dom, line 1343:
```suggestion
        fg: "all",
        bg: "all",
```

src/lib/config.ts
```diff
@@ -1334,6 +1334,15 @@ export class default_config {
      * Whether to show article url in the document.title of Reader View.
      */
     readerurlintitle: "true" | "false" = "false"
+
+    /**
+     *  Which css styles to add for hint elements.
```
bovine3dom, line 1339: We should say what fg/bg/outline mean and give an example of how to set it, e.g. `:set hintstyles.bg none`

bovine3dom (commented):
Thanks, just a couple of things then, happy to change them myself when I get a proper chance to look at this this weekend

PR #5108, opened 2024-12-15
Change preferHtml now defaults to false

cassepipe:
Sets the `preferHtml` option to false when using the editor adaptor

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -357,7 +357,7 @@ export async function editor() {
 
     let ans
     try {
-        const editor = getEditor(elem, { preferHTML: true })
+        const editor = getEditor(elem, { preferHTML: false })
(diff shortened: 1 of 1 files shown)


bovine3dom:
please add a key to src/lib/config.ts and use config.get / config.getAsync to read it so users can change preferHtml with set/seturl :)

cassepipe:
I thought it was agreed that defaulting to `false` was a better idea ?

I'd love to do that but I am currently unable to work locally on the project as it does not build so I don't feel confident changing anything more than what I did until I fix it.

bovine3dom:
Yes - but default implies that it is a setting that can be changed, so there are two steps:

1) make a new setting
2) set its default value to "false" (which has to be a string for boring reasons)

No worries if you can't do it at the moment.

PR #5142, opened 2025-01-14
Add `:open` note to readme.md

Xorger:
Add a note explaining how to set the newtab page for `:open`, which does not reference the default search engine


bovine3dom:
Thanks!

Issue #4721, opened 2023-06-05
filter :bmarks by folder


bovine3dom:
Currently we only match bookmarks by title and URL. We do not display or match folders.

We want need to tweak the deduplication in providers.ts to leave identical bookmarks in different folders 

Related: #1771

satyamkr1729:
I was thinking it to be more like: 
```command
:bmarks --folder <folder name> query string
```
The completions will then list URLs of only that folder. What do u think?

bovine3dom:
I'm not sure, to be honest. I don't use bookmarks, let alone folders, so I'm not the best person to ask.

My instinct would be to make it as simple as possible so that e.g. `:bmarks path/to/` would list all the bookmarks in that folder

vchernetskyi993:
As a person who uses bookmarks and likes to structure them by folder, want to add my five cents to this issue. It would be great to have the following 2 capabilities:
1. Search by **folder +** name + URL. Currently, the search only includes name + URL.
2. Print folder in addition to name to distinguish different bookmarks with same/similar names.

Let me give an example, why it would be helpful for me. I have AWS account bookmarks for different projects I work on. They are differentiated by project folders. But the names of the bookmarks are almost the same. Here is one really powerful mockup to add a clarity to the example:
![image](https://github.com/user-attachments/assets/e2f461c6-79f0-45d6-aa45-38af92fb3639)

vchernetskyi993:
I understand that my (2) is different to the issue itself, but I had some inspiration and tried coding it.

When you have time, please take a look at the PR: https://github.com/tridactyl/tridactyl/pull/5145, if this makes sense at all to you.

bovine3dom:
Added in #5216

PR #5145, opened 2025-01-15
Added path prefixes for bmarks

vchernetskyi993:
## Changes
1. Added folder prefixes to bookmarks list:
    * `:bmarks`: 
    image
    * `:tabopen`:
    image
2. Search looks for bookmark path in addition to title and URL. This is done by caching all bookmarks in memory, since Firefox doesn't search in path in its `browser.bookmarks.search()` function.
image
3. Added configuration property `bmarkfoldersearch` to disable this functionality by default.
    * `bmarkfoldersearch=false` works as before:
    image

Relates to #4721.

## TODO:
- [x] Add search by path
- [x] Limit recursion by undefined parent instead of magic constant for root
- [x] Add configuration property to disable folders by default
- [x] Include folders in "Bookmarks and History" view
- [x] Add screenshots for different scenarios

### DIFF
--- a/src/completions/Bmark.ts
+++ b/src/completions/Bmark.ts
@@ -10,18 +10,21 @@ class BmarkCompletionOption
     constructor(
         public value: string,
         bmark: browser.bookmarks.BookmarkTreeNode,
+        allBookmarks: { string?: browser.bookmarks.BookmarkTreeNode },
     ) {
         super()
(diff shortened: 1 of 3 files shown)


bovine3dom:
Thanks for this, it looks good at a glance but I need to find time to review it more thoroughly which probably won't happen till mid February :)

PR #5181, opened 2025-03-01
Added ability to move TST trees

vchernetskyi993:
It's annoying that it's impossible to move TST trees using keyboard commands (i.e. `:tabmove` works with indices, which are irrelevant in tree representation). Also, scripts don't help with this issue, since moving tabs to arbitrary position without completion is hard if not impossible.

Added 3 commans:
* `:tstmove` - same as `:tabmove` but moves the whole tree.
* `:tstmoveafter` - moves tree after the tab specified.
* `:tstattach` - attached current tree as a last child of the tab specified.

Tab representation is the same as before (linear) the main difference in logic is that TST commands use tab ID instead of indices.

Relates to #377.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -98,6 +98,7 @@ import { OpenMode } from "@src/lib/hint_util"
 import * as Proxy from "@src/lib/proxy"
 import * as arg from "@src/lib/arg_util"
 import * as R from "ramda"
+import * as treestyletab from "@src/interop/tst"
 
 /**
  * This is used to drive some excmd handling in `composite`.
@@ -2859,8 +2860,8 @@ export async function tabopen_helper({ addressarr = [], waitForDom = false }): P
         // and browser.search.search() seems to fix that problem.
         // See https://github.com/tridactyl/tridactyl/pull/4791.
         return openInNewTab(null, args, waitForDom)
-                   .then(tab => browser.tabs.get(tab.id))
-                   .then(tab => browser.search.search({tabId: tab.id, ...maybeURL}))
+            .then(tab => browser.tabs.get(tab.id))
+            .then(tab => browser.search.search({ tabId: tab.id, ...maybeURL }))
     }
 
     // Fall back to about:newtab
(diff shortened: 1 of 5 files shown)


src/excmds.ts
```diff
@@ -6208,3 +6209,27 @@ export async function elementunhide() {
     elem.className = elem.className.replace("TridactylKilledElem", "")
 }
 // vim: tabstop=4 shiftwidth=4 expandtab
+
+/**
+ * Move the current TST tree to be just in front of the tab specified.
+ */
+//#background
+export async function tstmove(tabId: string) {
+    treestyletab.moveTreeBefore(Number(tabId))
```
bovine3dom, line 6218: if we made `tstmove` operate on Tridactyl tab indices we wouldn't need such extensive changes to the completions - is there a downside to that?

```suggestion
export async function tstmove(tabindex: string) {
    const id = await idFromIndex(tabindex)
    treestyletab.moveTreeBefore(id)
```

vchernetskyi993: The TST itself works with IDs: https://github.com/piroor/treestyletab/wiki/API-for-other-addons#move-tree-to-different-position

What we can try doing is to pass an index here and calculate tab ID from it to pass to TST. Is this what you're proposing? If so, I can try reworking like this.

bovine3dom: yes, that's what I'm proposing :)

it should simplify the completion changes enormously - all you would need to do is add tstmove etc to the commands which get tab completions

vchernetskyi993: Ok, sounds reasonable to me. I'll try this rework during my spare time this week and ping you in this thread once ready. Thanks!

bovine3dom: Thanks! it shouldn't be much work - your functions in excmds.ts would gain a single line like my patch above, the completions code would get reverted and you'd add the command names to the array that triggers the Tab completions 

if you get stuck let me know

vchernetskyi993: @bovine3dom applied your suggestion. The code looks much cleaner now, thanks!

PR #5195, opened 2025-03-11
tabmove - added branch for previous tab index

### LINKED ISSUE #5182: `:tabmove #` produces an error
While playing around with https://github.com/tridactyl/tridactyl/pull/5181. I've noticed that I can't use `:tabmove` with `#` argument. It produces the following error: `controller in excmd: Error: Type error for parameter moveProperties (Error processing index: Expected integer instead of null) for tabs.move.`.

Looked through [`tabmove()` code](https://github.com/tridactyl/tridactyl/blob/f1156d5c7218fcc5a7e4840603e838e31eff85f4/src/excmds.ts#L3124), and it seems that branch for `#` is simply missing there.

Looks like a minor defect to me.

vchernetskyi993:
Simple support for `:tabmove #`.

Resolves #5182

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3124,6 +3124,14 @@ export async function undo(item = "recent"): Promise<number> {
 //#background
 export async function tabmove(index = "$") {
     const aTab = await activeTab()
+    if (index === "#") {
+        const previousTab = await prevActiveTab()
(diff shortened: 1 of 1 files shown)


bovine3dom:
LGTM, thanks!

PR #5198, opened 2025-03-14
Explain what the config.test.ts test does

cmcaine:


bovine3dom:
if you're playing with the key sequence parser, you might want to take a look at https://github.com/tridactyl/tridactyl/pull/4727 where I was trying to add support for explicit keydown / keyup events

it would let us do lots of cool stuff, the main one people care about being smooth scroll

PR #5200, opened 2025-03-17
add new theme (gold) based on midnight

MrGueest:


bovine3dom:
Thanks for this. Something had broken in our build process so I can't test this until we fix that - see issue #5202 :(

PR #5213, opened 2025-04-25
excmd help edit

FugueSoft:
- Added a few lines to editorcmd description for kitty nvim users
- Typographical update.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -270,12 +270,12 @@ export function fillinput(selector: string, ...content: string[]) {
 
     // CodeMirror support (I think only versions prior to CodeMirror 6)
     if (inputToFill?.parentNode?.parentElement?.className?.match(/CodeMirror/gi)) {
-        ;(inputToFill.parentNode.parentElement as any).wrappedJSObject.CodeMirror.setValue(content.join(" "))
+        ; (inputToFill.parentNode.parentElement as any).wrappedJSObject.CodeMirror.setValue(content.join(" "))
         return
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -330,6 +330,11 @@ import { getEditor } from "editor-adapter"
  * set editorcmd terminator -u -e "vim %f '+normal!%lGzv%c|'"
  * ```
  *
+ * Your editor of choice needs a terminal to run it. For example, this command opens neovim with kitty and exits after closing the editor:
+ * ```vim
+ * set editorcmd kitty nvim
```
bovine3dom, line 335: stupid question: does this actually work? i thought you said it needed --hold?

FugueSoft: It does! The `--hold` flag actually holds kitty open even after the child process exits. So the default is to close the doogle and get back to the boogle. Ya feel me?

src/excmds.ts
```diff
@@ -330,6 +330,11 @@ import { getEditor } from "editor-adapter"
  * set editorcmd terminator -u -e "vim %f '+normal!%lGzv%c|'"
  * ```
  *
+ * Your editor of choice needs a terminal to run it. For example, this command opens neovim with kitty and exits after closing the editor:
```
bovine3dom, line 333:
```suggestion
 * Your editor of choice may need to run in a terminal. For example, this command opens neovim with kitty and exits after closing the editor:
```

Issue #4413, opened 2022-10-09
Tridactyl install on Android Firefox nightly failing 


yashasolutions:
## Brief description of the problem

Installation of Tridactyl on Firefox Nightly for Android 

-   Steps to reproduce:

   1.  Using the AOM method
    - Create a collection on AOM
    - Add Tridactly to the collection
    - Select installation of Tridactyl
    - Get message that installation failed. 

   2. Using the web-ext method
    - download the latest release (1.22.1)
    - install using `$ web-ext run -t firefox-android --adb-device R9XXXXXV --firefox-apk org.mozilla.fenix`
    - get following error:
```
WebExtError: installTemporaryAddon: Error: Error: Could not install add-on at '/data/local/tmp/web-ext-artifacts-1665287819161/tridactyl-1.22.1.xpi': Error: Extension is invalid
    at RemoteFirefox.installTemporaryAddon (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/firefox/remote.js:100:13)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async FirefoxAndroidExtensionRunner.rdpInstallExtensions (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/extension-runners/firefox-android.js:506:23)
    at async FirefoxAndroidExtensionRunner.run (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/extension-runners/firefox-android.js:87:5)
    at async Promise.all (index 0)
    at async MultiExtensionRunner.run (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/extension-runners/index.js:72:5)
    at async run (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/cmd/run.js:180:3)
    at async Program.execute (file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/lib/program.js:289:7)
    at async file:///home/user/.nvm/versions/node/v16.17.0/lib/node_modules/web-ext/bin/web-ext.js:13:1

```

So, unsure if it's a bug, a version of firefox that's not compatible, or maybe not meant to work on android (even if it shows up in the AddOns store for android)...

Has anyone a working version on android? Is there anything that can be done based on the error messages I got?

Thanks!

bovine3dom:
[...]

PR #5215, opened 2025-05-05
Add compatibility layer for Firefox Android

bovine3dom:
In theory #4413 should just become a game of whack-a-mole now working through the errors and adding compat / notImplemented to them


bovine3dom:
```sh
~/p/tridactyl (android_gates|✔) $ node_modules/.bin/eslint --rulesdir custom-eslint-rules --ext .ts . | awk '{print $3}' | sort | uniq

bookmarks.create
bookmarks.getTree
bookmarks.remove
bookmarks.search
commands.getAll
commands.onCommand
commands.update
find.find
history.search
omnibox.onInputEntered
omnibox.setDefaultSuggestion
problems
runtime.sendNativeMessage
search.get
search.search
sessions.getRecentlyClosed
sessions.getTabValue
sessions.getWindowValue
sessions.removeTabValue
sessions.removeWindowValue
sessions.restore
sessions.setTabValue
sessions.setWindowValue
sidebarAction.close
sidebarAction.open
sidebarAction.setPanel
sidebarAction.toggle
tabs.discard
tabs.duplicate
tabs.getZoom
tabs.hide
tabs.move
tabs.onMoved
tabs.setZoom
tabs.show
tabs.toggleReaderMode
windows.get
windows.getAll
windows.getCurrent
windows.getLastFocused
windows.remove
windows.update
```

src/lib/compat.ts
```diff
@@ -0,0 +1,37 @@
...
+    return _isAndroid
+}
+isAndroid()
+
+export const isChrome = () => typeof browser === undefined // if they ever support web extension standards this will break
+
+export const windows = {
+    create: async (props: browser.windows._CreateCreateData) => {
+        if (!(await isAndroid())) {
+            // eslint-disable-next-line unsupported-apis
```
bovine3dom, line 25: one problem with this pattern is that we're promising to support all browsers that reach this line

not sure if there's an obvious fix?

Iey4iej3: It might be useful to simply put such questions on https://support.mozilla.org/

bovine3dom:
```
  /home/runner/work/tridactyl/tridactyl/src/lib/compat.ts:11
          return browser.runtime.getPlatformInfo().then(info => {
                                 ^
  
  TypeError: browser.runtime.getPlatformInfo is not a function
```

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getPlatformInfo

i guess we need to add a mock for it

.eslintrc.js
```diff
@@ -44,6 +44,10 @@ module.exports = {
             "warn",
```
bovine3dom, line 43: surely this should already be "error" 🤔

bovine3dom:
i am once again debating what the point of this is if `history.search` doesn't exist on android (!)

i have been wanting to maintain our own copy of history so we can have tree navigation like we do for tab history, but it would be a big scary feature and potentially quite badly surprising to users if they use the firefox menu to clear history but find that tridactyl's is not cleared (then again, we _sort of_ already have this issue with history in the command line)

bovine3dom:
> Indeed I disagree [that it is pointless without history]. Yes full support would be better... but even if it's only navigation on a page (even without hinting), between tabs, search within a page, search with defined engines, then it would still be better than what's possible otherwise, namely only typing content in the URL bar or going up or down with arrows.

> That said it is a niche within a niche. I imagine very few people need that (hence why Mozilla itself does not bother with even the most basic shortcuts) but those who do, including me, will appreciate!

from @utopiah in #4413 

i am ok with supporting a niche within a niche. i am less ok with things that have zero users because zero users = zero issues filed = we don't notice when it breaks (as happened with the previous Android version)

so the blocker on this goes from `history.search` back to "the current eslint workflow makes me uncomfy". doing as much runtime validation as possible (entirely missing APIs) is probably a good idea, but dealing with subtly different APIs probably still needs us to use the browser compat DB somehow

bovine3dom:
image

kind of a odd reason to be motivated to make progress on this but i smashed my phone screen this morning so i've been using `scrcpy` all day and getting annoyed at how bad the web on android is to navigate using the keyboard

the code is horrific and unpushed/uncommitted and needs days/weeks of finessing but it does load and most non-`:` binds work

`:tab` (!) and `:open` completions don't work, `:tab [n]` doesn't work, `gt`/`gT` don't work. and the extension crashed at some point and i couldn't work out why

bovine3dom:
having sat on this for a while i am a bit frustrated with the quality of the BCD data for web extensions on android.

i think BCD can only be a kind of first pass where if an API claims to be missing in BCD it is probably missing on android. but an API being present doesn't mean it actually works.

so probably the best thing to do is to change tactics slightly and use BCD as advisory only, with some kind of Android emulator running in CI as an (invariably flaky) source of truth

PR #5221, opened 2025-05-18
Improve docs for js & jsb

bew:
Note: The added `Usage` lines help better show that `[arg]` is only available when `-p` is used, and same for `-d{char}`.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -5896,22 +5896,34 @@ async function js_helper(str: string[]) {
  *
  * Usage:
  *
- *        `js [-p] javascript code ... [arg]`
+ *     `js javascript code ...`
  *
- *        `js [-s|-r|-p] javascript_filename [arg]`
+ *     `js -p javascript code ... arg`
+ *
+ *     `js [-s|-r] javascript_filename`
+ *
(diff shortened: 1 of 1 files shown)


src/excmds.ts
```diff
@@ -5896,23 +5896,35 @@ async function js_helper(str: string[]) {
...
  *
- *        `js [-s|-r|-p] javascript_filename [arg]`
+ *     `js -p javascript code ... arg`
+ *
+ *     `js [-s|-r] javascript_filename`
+ *
+ *     `js -p [-s|-r] javascript_filename arg`
+ *
+ *     `js -d# [-s|-r] javascript_filename # arg1 arg2 ...`
+ *     (where `#` is any char)
```
bovine3dom, line 5908:
```suggestion
 *     `js -d³ [-s|-r] javascript_filename³ arg1 arg2 ...`
 *     (where `³` is any char  that you can guarantee won't appear in your JS code)
```

`#` works at the moment but i'm reluctant to encourage it because it's also our comment character

bew:
Good call on #, I personally use § in my config 😀

Issue #5240, opened 2025-07-25
(smooth) scrolling broken since Firefox 141.0 update


HattDroid:
**bovine3dom edit: as a workaround run** `:unset smoothscroll`

---

Using j/k to navigate web pages seems to be broken since Firefox 141.0 update.
Pressing the keys once or twice would work, but a few more times in a row than that would render them inactive. 

Did they add maybe a policy on repeated keys?

Tried on two different machines and also reverted back to 140.4 where it works. It seems 141.0 broke something.

tobsz:
Same problem with C-d/C-u.

bovine3dom:
I'll look into this ASAP. In the meantime downgrade to F140 if you can

bovine3dom:
could you confirm your operating systems and how Firefox is packaged?

quisquama:
I'm on Arch Linux, using the packages `firefox`,  `firefox-tridactyl` and `firefox-tridactyl-native`.

bovine3dom:
I can only reproduce this with `:set smoothscroll true`, can you confirm that `:unset smoothscroll` fixes it?

quisquama:
`:unset smoothscroll` fixes this for me.

HattDroid:
Thanks a lot for checking in that quick and for the workaround @bovine3dom . 👍

The only thing that seems somehow related in the changelog seems to be that below:

>The [scrollMargin](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/scrollMargin) property of the [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) interface is now supported. It adds a margin to all nested [scroll containers](https://developer.mozilla.org/en-US/docs/Glossary/Scroll_container) within the root element of the observer, which allows targets inside those elements to be observed before (or after) they are scrolled into view—instead of only when they first become visible. ([Firefox bug 1860030](https://bugzil.la/1860030))."

I have very little knowledge on Firefox internals, do you think it could be that?

HattDroid:
Thanks a ton for the quick fix @bovine3dom .
Wishing you good luck to find a fix for the smoothscroll feature!

bovine3dom:
Fixed in #5244 (thanks oldshensheep :) )

PR #5244, opened 2025-07-30
Fix smooth scrolling in Firefox 141

oldshensheep:
This pull request fixes the issue: https://github.com/tridactyl/tridactyl/issues/5240 .

Relevant code:
https://github.com/tridactyl/tridactyl/blob/344fc9fad7d6787f8ec4425984740baa56d72abf/src/content/scrolling.ts#L88-L103

On Firefox 140 `this.elem[this.scrollDirection]` will always return an interger number. However, in Firefox 141, it sometimes returns a float.

The following assumptions are based on my testing in Firefox 141 and 140:

Assume:
`pixelToScrollTo = 564.7`
`this.elem[this.scrollDirection] = 564.7`

With `pixelToScrollTo = Math.ceil(pixelToScrollTo)`, we get:
`pixelToScrollTo = 565`

Thus, `pixelToScrollTo == this.elem[this.scrollDirection]` returns `false`, and the function will return 565.

https://github.com/tridactyl/tridactyl/blob/344fc9fad7d6787f8ec4425984740baa56d72abf/src/content/scrolling.ts#L106-L111

In the function `scrollStep`, we assign `pixelToScrollTo` to `this.elem[this.scrollDirection]`, but in Firefox 141, this doesn't work as expected. After the assignment, `this.elem[this.scrollDirection]` remains 564.7, so `prevScrollPos === this.elem[this.scrollDirection]` still holds true, and the function returns `false`.

I don't know why `this.elem[this.scrollDirection]` is a float in Firefox 141 but not in Firefox 140. I also don’t have a high-DPI display to test the behavior in Firefox 140 (e.g., what happens if `this.elem[this.scrollDirection]` is 564.7 and you assign it 565? If it behaves like Firefox 141, then the same problem would occur).

By using `Math.ceil(this.elem[this.scrollDirection])`, we ensure that the delta is always >= 1. I _think_ this works — and in my testing, it does.

### DIFF
--- a/src/content/scrolling.ts
+++ b/src/content/scrolling.ts
@@ -83,11 +83,11 @@ class ScrollingData {
             // We need to ceil() because only highdpi screens have a decimal this.elem[this.pos]
             pixelToScrollTo = Math.ceil(pixelToScrollTo)
             // We *have* to make progress, otherwise we'll think the element can't be scrolled
-            if (pixelToScrollTo == this.elem[this.scrollDirection])
+            if (pixelToScrollTo == Math.ceil(this.elem[this.scrollDirection]))
                 pixelToScrollTo += 1
         } else {
             pixelToScrollTo = Math.floor(pixelToScrollTo)
-            if (pixelToScrollTo == this.elem[this.scrollDirection])
+            if (pixelToScrollTo == Math.floor(this.elem[this.scrollDirection]))
                 pixelToScrollTo -= 1
         }
         return pixelToScrollTo


bovine3dom:
Works great, thanks so much for nailing this down. I'll try to do a release on Sunday.

> I also don’t have a high-DPI display to test the behavior in Firefox 140 (e.g., what happens if this.elem[this.scrollDirection] is 564.7 and you assign it 565? If it behaves like Firefox 141, then the same problem would occur).

on my high DPI display in FF140 `this.elem[this.scrollDirection]` is always an integer and the bug doesn't occur :shrug:

Issue #5259, opened 2025-09-08
Call document.write after page loaded cause Tridactyl stop working


GHolk:
Tridactyl does not work if the page call document.write() after the page load.

## Example
Any page load with [html preview](https://htmlpreview.github.io/) .

* https://htmlpreview.github.io/?https://github.com/FiloSottile/age/blob/main/doc/age.1.html

## What Happen
Call `document.write()` after the page loaded (or when document is closed)
will detach the old root element (`<html>` or document.documentElement),
and create a new one.

This will not trigger any DOM Mutation event on the old root element.
They are still accessible if any js code hold a reference to it.

A new root will also clear all(?) event listener on the window,
although properties on window are intact, like `window.tri` .

## How to detect document.write
I don't know. It remove the document.onreadystatechange listener.
Maybe HTMLElement.prototype.isConnected property?
Or just hijack the `document.write` method?
We already do this in HTMLElement.prototype.focus .

### Other Extension Workround (LinkHints)
They just hijack document.write to detect document.write event.

* https://github.com/lydell/LinkHints/issues/86
* https://github.com/lydell/LinkHints/commit/5241c826253813df9af38ef2bdc354766a1ae287

## How to fix this
Call `listen(window)` in `src/content.ts` , append excmd iframe and mode indicator again after document.write

## Similiar Bugs in Tridactyl
*rehydration* in react:

* https://github.com/tridactyl/tridactyl/issues/5050
* https://github.com/tridactyl/tridactyl/pull/5053

But react still send a DOM mutation event,
because it remove only the `cmdline_iframe` but not the root.

We can still reuse the re-append iframe's code here.

GHolk:
Wow, very fast! Thank you!

PR #5261, opened 2025-09-09
Attempt to restore Tridactyl if document.open, write or writeln are called

Chic-Tweetz:
Regarding https://github.com/tridactyl/tridactyl/issues/5259.

Some sites use the (deprecated) `document.open` / `document.write` / `document.writeln` functions which create entirely new `document` objects. Event listeners and elements are consequently lost.

This hijacks those functions to ~dispatch an event~ call an exported function before the `document` is replaced, allowing us to readd listeners, the commandline and status indicator.

I'm not sure that I've found every listener which needs re-listening.

This also assumes that `document.close` will be promptly called, as that's when everything will be readded.

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -111,6 +111,7 @@ const visual = await import("@src/lib/visual")
 const metadata = await import("@src/.metadata.generated")
 const { tabTgroup } = await import("@src/lib/tab_groups")
 const completion_providers = await import("@src/completions/providers")
+const registry = await import("@src/content/element_registry")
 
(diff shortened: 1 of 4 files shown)


bovine3dom:
> Is there anything else in Tridactyl that has to run at transpile time?

yeah, the `compiler/` folder which gives us type checking at runtime for `:set` and type conversion for ex-commands.

it is more or less witchcraft, if it's too tricky we can just do it runtime like you're doing at the moment

PR #5265, opened 2025-09-15
More Consistent Status Indicator Classes

matystroia:
This PR fixes two issues I had when trying to style the status indicator:

1. Mode `className` (e.g. `TridactylModevisual`) "flickers" on every keypress even when mode doesn't change
2. Sometimes (for example when exiting insert mode) mode `className` gets out of sync with status indicator text, leading to multiple mode classes being applied at once

As far as I can tell this shouldn't break anything since I've kept the same class order.


src/content.ts
```diff
@@ -401,38 +396,49 @@ config.getAsync("modeindicator").then(mode => {
...
+            ? "TridactylPrivate"
+            : ""
+        const modeCls = `TridactylMode${result}`
+        const invisibleCls =
             config.get("modeindicator") !== "true" ||
             config.get("modeindicatormodes", mode) === "false"
-        ) {
-            statusIndicator.classList.add("TridactylInvisible")
-        } else {
-            statusIndicator.classList.remove("TridactylInvisble")
```
bovine3dom, line 413: huh, I think the real bug here is the typo for `Invisible`

is the rest of the PR necessary?

(sorry for accidentally making you wait a year)

matystroia: No worries! :)

Yes, I still think this is a bug. Currently `statusIndicator.className` is being assigned to multiple times, and this leads to the DOM rapidly changing (thus the flickering I mentioned) even when the classes shouldn't change. Building the `className` before assigning fixes this by removing the interim states and changing the DOM only when the final class list actually changes.

Issue #5281, opened 2025-10-16
reader mode loses favicon


repparw:
Toggling reader mode on a website gets me the globe favicon 

<img width="34" height="32" alt="Image" src="https://github.com/user-attachments/assets/34ca52cd-85bf-4aa5-8b19-4fdbc6738509" />

I'm guessing because of the moz-extension link? Is there a way to change/workaround this?

Chic-Tweetz:
Nice idea! The favicon can be grabbed before loading the reader page and added with the rest of the content. [Adding a \<link rel="icon"> with the right href seems to be all that's needed](https://github.com/Chic-Tweetz/tridactyl/commit/b85f6192905ee9378ab554fabd0df2aa7be1aa38)

Chic-Tweetz:
I was wondering whether the `<link>` should include a MIME type and/or icon sizes, but I don't think they're needed and it seems to work as is.

I've opened a PR for it now https://github.com/tridactyl/tridactyl/pull/5283

PR #5283, opened 2025-10-17
favicon in reader

Chic-Tweetz:
For #5281

When running `:reader`, adds the favicon URL associated with the tab to the `article` object. Then uses the URL as the href for a `<link rel="icon">` element when creating the elements on the `:reader` page in order to display the original page's icon in the tab bar.


bovine3dom:
Works great, thanks!

Issue #5239, opened 2025-07-19
Warn before executing tabonly or prevent calling with args, consider undo functionality


micimize:
I just ran `:tabo blah` thinking tabo would alias to tabopen, only to realize after my sidetab bar jittered a bunch that it had closed something like 200 tabs in the window.

Destructive actions like this should be hard to do accidentally by default, not easily confused with other common commands

bovine3dom:
Hi, sorry, I can see that that wouldn't be much fun. I think we copied the `:tabo` alias from Vimperator/Pentadactyl without thinking about it too much.

Adding a warning if arguments are supplied seems like a sensible safety feature

ztstroud:
I got some feedback on Matrix, I am working on this issue.

PR #5293, opened 2025-11-23
Improve user experience with tabopen and tabonly

ztstroud:
tabonly can be destructive. The tabo alias looks like it could be for tabopen, and running tabonly with arguments still closed all other tabs.

Fixes: #5239

---

`yarn run pretty` seems to have a few changes it wants to make across `excmds.tx`. I made sure that my change has run through prettier and checked it against existing source, but didn't want to include everything else here because it seems unexpected. It also gives warnings about APIs not supported in `chrome`, which also seems odd.

Is this something that others run into? Or is this some kind of local configuration that I have interfering? I don't see any other prettier configurations, and that isn't standard anyways, so I am not quite sure what that would be.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -2965,7 +2965,7 @@ async function tabFromIndex(index?: number | "%" | "#" | string): Promise<browse
 //#background
 export async function tabonly(...args: string[]) {
     if (args.length > 0) {
-        fillcmdline_tmp(3000, "# tabonly doesn't accept arguments, run without arguments to close all tabs in this window")
+        fillcmdline_tmp(3000, "# tabonly doesn't accept arguments, run without arguments to close all background tabs in this window")
         return
     }
 


bovine3dom:
Thanks! don't worry about the pre commit hooks. In theory prettier should only run on the lines you have changed but that broke ages ago and now it messes with the whole file. Have made #5295

The warnings about chrome are for future cross browser compatibility. Don't worry about them :)

ztstroud:
After discussion on Matrix, we thought it would be good to remove the `;tabo` alias entirely. Users can add it back if they want, but @bovine3dom wasn't aware of high usage.

With the removal of the alias, there isn't as much of an internal link between `:tabonly` and `:tabopen`, so I think it makes sense to remove that specific mention in the error. The error itself is still useful, because calling `:tabonly <name>` still indicates that a user is about to get an unexpected result.

PR #5294, opened 2025-11-23
Add save-as filename marker support and minor clean-ups

gsbabil:
Among other minor clean-ups, this pull request introduces a marker/placeholder string for :saveas targets. If the save-as path contains the marker, it is replaced with the filename derived from the URL. Paths without the marker continue to work as before, so there are no breaking changes.

This also makes repeated :saveas calls—including using . to repeat commands across multiple tabs—much smoother, since each save naturally picks up a unique name using your chosen prefix/suffix without needing to retype anything.

## Example

```vim
:saveas ~/Documents/ex-...
```

With the URL:

```
https://example.com/log.txt
```

This will create the file:

```
~/Documents/ex-log.txt
```

## Notes

- If the path does not contain the marker, it is used as-is.
- Multiple occurrences of the marker are all replaced with the derived filename.
- Users can run `:set downloadfilenamemarker XXXX` to change the default `...` to something else.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -508,7 +508,9 @@ export async function unloadtheme(themename: string) {
  *
  * If THEMENAME is set to any other value except `--url`, Tridactyl will attempt to use its native binary (see [[native]]) in order to load a CSS file named THEMENAME from disk. The CSS file has to be in a directory named "themes" and this directory has to be in the same directory as your tridactylrc. If this fails, Tridactyl will attempt to load the theme from its internal storage.
  *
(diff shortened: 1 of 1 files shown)


bovine3dom:
Hi, thanks for this. I will take a look in detail later

One thing that's jumped out at me is you're adding `bun` support. I don't mind people using different package managers for development but we can't have multiple lockfiles - there needs to be one source of truth for which version of each package a commit was using.

#5060 was opened a while ago but went stale. I can't remember what the problem was (... it might have been trying to do too much in one PR...). So please remove the bun stuff from here and if you're really motivated we can talk about it somewhere else

bovine3dom:
Closing in favour of #5304 and #5305 :)

PR #5297, opened 2025-11-23
Fix reader on big pages

bovine3dom:
Not sure how I feel about using sessionStorage. It means if you close the tab and reopen it, you lose the article content

Maybe we can add some kind time-based of GC? I don't want to overcomplicate things.

There are some potential security considerations too - we've gone from storing the entire page in the web history (... arguably bad) to storing the entire page in local storage (...also arguably bad)

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1368,6 +1368,12 @@ export class default_config {
         bg: "all",
         outline: "all",
     }
+
+    /**
+     * Internal temporary storage for :reader, mapping UUIDs to base64 encoded html strings of articles
+     */
+    reader_articles: { [id: string]: string } = {
+    }
 }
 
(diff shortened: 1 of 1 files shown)


bovine3dom:
```
src/lib/math.ts(74,13): error TS2365: Operator '+' cannot be applied to types 'number[]' and 'number'.
src/reader.ts(10,41): error TS2345: Argument of type '"reader_articles"' is not assignable to parameter of type '"update" | "configversion" | "subconfigs" | "modesubconfigs" | "priority" | "exmaps" | "ignoremaps" | "imaps" | "inputmaps" | "superignore" | "nmaps" | "vmaps" | "hintmaps" | ... 97 more ... | "hintstyles"'.
```

🥱

PR #5342, opened 2026-02-16
Cache Fuse.js instances instead of rebuilding on every search

repparw:
## Summary

- Add getter/setter for `options` that automatically invalidates Fuse index when options change
- Add `buildFuseIndex()` method for lazy initialization instead of rebuilding on every search
- Add `setOptionsWithoutInvalidation()` for internal reordering without triggering rebuild

This addresses the performance issue noted in the code comments: "PERF: Could be expensive not to cache Fuse() // yeah, it was."

Before: Fuse.js index was rebuilt on every call to `scoredOptions()`, which happens on every keystroke during completion search.

After: Fuse index is built once when `options` change, then reused for subsequent searches until options are updated again.

## Testing

- [ ] Test completions work correctly (tabs, bookmarks, history, etc.)
- [ ] Verify search results are still accurate
- [ ] Check for any regression in completion performance

### DIFF
--- a/src/completions.ts
+++ b/src/completions.ts
@@ -154,7 +154,6 @@ export interface ScoredOption {
 
 export abstract class CompletionSourceFuse extends CompletionSource {
     public node
-    public options: CompletionOptionFuse[]
 
     fuseOptions = {
         keys: ["fuseKeys"],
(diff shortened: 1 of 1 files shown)


bovine3dom:
closing in favour of #5374 which fixes a few bugs

Issue #5309, opened 2025-12-05
reader hangs with 100% cpu usage


repparw:
Sometimes (?) reader tabs makes firefox hang. Checking about:processes shows Extensions using 100% cpu usage.

How can I go about getting logs/debugging this to nail what is happening? Help appreciated

<details>
<summary>tridactylrc</summary>

```" General Settings
set configversion 2.0
set newtab about:blank
set markjumpnoisy false
set modeindicatormodes.ignore false
set theme midnight
set editorcmd foot nvim
set smoothscroll true

" Binds
bind , hint -W mpvsafe
bind ;c hint -c [class*="expand"],[class*="togg"],[class="comment_folder"]

unbind <F1>
unbind <C-e>

bind gd tabdetach

bind yy clipboard yankshort

bind J tabnext
bind K tabprev

bind e reader

" Subconfig binds
bindurl .*.youtube.com yy composite urlmodify_js -Q list | urlmodify_js -ru .*\.youtube\.com/watch\?v= https://youtu.be/ | clipboard yank

bindurl ^moz-extension:\/\/.*\/static\/reader\.html yy clipboard yankcanon

unbindurl x.com e
unbindurl x.com j
unbindurl x.com k

" Subconfig Settings
seturl youtube.com modeindicator false
seturl jellyfin.repparw.me modeindicator false

autocmd DocStart tradingview.com mode ignore
```

</details>

I had multiple reader tabs in a tab group. when loading one/a couple of them it hangs. Had to disable extension or close/unload tabs

repparw:
yeah, sorry I didn't clarify, tridactyl's reader. I have bind e reader. It doesn't seem to happen when turning a new tab to reader, only when loading an unloaded tab that is on reader mode (e.g moz-extension://{id}/static etc)

repparw:
1.24.4

PR #5348, opened 2026-02-24
fix: prevent infinite loop on reader mode

repparw:
## Summary
- Add a reentrancy guard to `getURL()` to prevent infinite loops when the config system triggers itself during reader mode initialization
- This fixes 100% CPU usage when using Tridactyl's reader mode (at least happening with [my config](https://github.com/repparw/nix/blob/main/modules/hm/gui/browser/tridactyl.nix) ~~in beta 1.24.4pre7305~~ wrong!)

## Problem
When using Tridactyl's reader mode (`:reader`), Firefox would sometimes go to 100% CPU usage. The stack trace shows an infinite loop between:
- `getURL` → `_getURL` → `filter` → `get` → `getAsync` → message handler → `getURL` (repeats)

## Root Cause
The `getURL()` function can trigger config reads which indirectly call `getURL()` again, creating an infinite recursion. This is likely triggered by site-specific config patterns or the config initialization sequence.

## Fix
Added a simple reentrancy guard (`inGetURL`) that prevents recursive calls to `getURL()`. If a recursive call is detected, it returns `undefined` early, which allows the config system to fall back to default values.

## Testing
Tested with my config (which I think is related to the recursion I got in the first place)

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -39,6 +39,8 @@ const CONFIGNAME = "userconfig"
 const WAITERS = []
 /** @hidden */
 export let INITIALISED = false
+/** @hidden */
+let inGetURL = false
 
 /** @hidden */
 // make a naked object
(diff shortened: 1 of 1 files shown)


repparw:
I really needed sleep haha was going insane.

I actually don't know how to recreate it in beta, without the old tridactyl reader mode. (I'm pretty sure it's what caused #5309 in stable)

So you mostly can ignore it unless we get recursion somewhere else I guess

Issue #4945, opened 2024-03-10
tabdetach doesn't work on #


bovine3dom:
`:tabdetach #` "can't convert to number"

bovine3dom:
Duplicate of #4244

Issue #4244, opened 2022-05-24
tabdetach completions can't convert # to number


bovine3dom:
Try to use `:tabdetach` to detach the previously used tab. The completion fills in # which tabdetach can't apparently cope with.

PR #5355, opened 2026-03-05
Harden detachtab with window creation and support symbolic tab indexes

wskozel:
To address Issue https://github.com/tridactyl/tridactyl/issues/4244
and https://github.com/tridactyl/tridactyl/issues/4945

The excmd "tabdetach" was failing in two ways during workflows:

1. Firefox sometimes threw NS_ERROR_MALFORMED_URI during browser.windows.create(...).
2. The command parser treated tabdetach index as number, so symbolic tab selectors like #/% (and some completion flows) could trigger  "# controller while accepting: Error: Can't convert to number: #"

This change:

1. Creates the temporary detach window with url: "about:blank" and adds a fallback create path.
2. Updates tabdetach to accept index?: string, matching idFromIndex/tabFromIndex behavior.
3. Awaits temp-tab cleanup/focus calls for deterministic command completion.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3040,14 +3040,22 @@ export async function tabduplicate(index?: number) {
         The 1-based index of the tab to target. index < 1 wraps. If omitted, this tab.
 */
 //#background
-export async function tabdetach(index?: number) {
+export async function tabdetach(index?: string) {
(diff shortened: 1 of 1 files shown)


bovine3dom:
Closing in favour of #5454

Issue #3408, opened 2021-02-24
Visual mode binds aren't 100% the same as in vim


bovine3dom:
As mentioned in #3405 

`w` should highlight the first letter of the next word. It doesn't.

In general the original character of the selection should always stay highlighted, usually the first character of the highlighted word unless someone has pressed `o`.

Visual mode is defined here: https://github.com/tridactyl/tridactyl/blob/9e188159be02472f34da7f1df8376f08e5bf797a/src/lib/config.ts#L353

bovine3dom:
From dementorr on Matrix:

```
Hello! When selecting in visual mode and pressing `b`, the first letter of the word won't be selected. I think the problem is here: https://github.com/tridactyl/tridactyl/blob/9e188159be02472f34da7f1df8376f08e5bf797a/src/lib/config.ts#L365 . Can someone extend it by adding at the end of the line `js

document.getSelection().modify("extend","backward","character")` ? It would be the same as with `w`. Also if you add `extend forward character` again for w, it will close issue #3408.

My bad, actually the line is https://github.com/tridactyl/tridactyl/blob/9e188159be02472f34da7f1df8376f08e5bf797a/src/lib/config.ts#L369 and you should delete the last part `document.getSelection().modify("extend","forward","character")`

```

bovine3dom:
Fixed in #5362

PR #5362, opened 2026-03-22
Added a guard to prevent selection collapse using h, l motion in visu…

Arjan-P:
Added a guard for h, l motions: 
```
if(sel.isCollapsed) {
    // create a new selection where old selection focus = new selection anchor instead of collapsing selection
}
```

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -399,8 +399,36 @@ export class default_config {
         y: "composite js document.getSelection().toString() | clipboard yank",
         s: "composite js document.getSelection().toString() | fillcmdline open search",
         S: "composite js document.getSelection().toString() | fillcmdline tabopen search",
-        l: 'js document.getSelection().modify("extend","forward","character")',
-        h: 'js document.getSelection().modify("extend","backward","character")',
+        l: `js
+            const sel = document.getSelection();
+            const oldFocusNode = sel.focusNode;
+            const oldFocusOffset = sel.focusOffset;
+
+            sel.modify("extend", "forward", "character");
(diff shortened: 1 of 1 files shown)


Arjan-P:
```              
sel.setBaseAndExtent(
    oldFocusNode,
    oldFocusOffset,
    sel.focusNode,
    sel.focusOffset + 1
);
```
Setting the offset manually as +1, -1 can break at DOM node boundary.

```
src/lib/visual.ts
export function extendByCharacter(selection: Selection, direction: string): void {
    const oldFocusNode = selection.focusNode;
    const oldFocusOffset = selection.focusOffset;
    const oldAnchorNode = selection.anchorNode;
    const oldAnchorOffset = selection.anchorOffset;
    if (!oldFocusNode) return;

    selection.modify("extend", direction, "character");
    if(!selection.isCollapsed) return;
    selection.setBaseAndExtent(
        oldAnchorNode,
        oldAnchorOffset,
        oldFocusNode,
        oldFocusOffset
    );
    reverseSelection(selection);
    selection.modify("extend", direction, "character");
}
```
could this be a more robust solution? 
I think this function could also be used to solve https://github.com/tridactyl/tridactyl/issues/3408, missing selection of first character using w, b

bovine3dom:
yes. thanks

annoyingly i wrote that first then changed my mind and decided + 1 was tidier

a PR would be most welcome :)

Arjan-P:
Sorry to ask, but I'm not sure how to. Locally I managed to build and test by manually adding modify as a property in the type declaration of Selection type, but by default modify does not exist as a property and fails the test.

PR #5366, opened 2026-03-31
Fix selection across DOM node

Arjan-P:
Added function:
```
src/lib/visual.ts
export function extendByCharacter(
    selection: Selection,
    direction: string,
): void {
    const oldFocusNode = selection.focusNode
    const oldFocusOffset = selection.focusOffset
    const oldAnchorNode = selection.anchorNode
    const oldAnchorOffset = selection.anchorOffset
    if (!oldFocusNode) return

    selection.modify("extend", direction, "character")
    if (!selection.isCollapsed) return
    selection.setBaseAndExtent(
        oldAnchorNode,
        oldAnchorOffset,
        oldFocusNode,
        oldFocusOffset,
    )
    reverseSelection(selection)
    selection.modify("extend", direction, "character")
}
```

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -401,33 +401,11 @@ export class default_config {
         S: "composite js document.getSelection().toString() | fillcmdline tabopen search",
         l: `js
             const sel = document.getSelection();
-            const oldFocusNode = sel.focusNode;
-            const oldFocusOffset = sel.focusOffset;
-
-            sel.modify("extend", "forward", "character");
-            if (sel.isCollapsed) {
-              sel.setBaseAndExtent(
-                oldFocusNode,
-                oldFocusOffset,
-                sel.focusNode,
-                sel.focusOffset + 1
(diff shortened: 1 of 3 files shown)


bovine3dom:
we generally add Firefox specific types here https://github.com/tridactyl/tridactyl/blob/master/src/tridactyl.d.ts

it's not exactly the easiest thing to do so if you're struggling ping me and I'll take a look

Arjan-P:
Would something like this work?
```
// Firefox-specific dom properties
interface Window {
    scrollByLines(n: number): void
    scrollByPages(n: number): void
    eval(str: string): any
}
interface Selection {
    modify(
    alter: "move" | "extend",
    direction: "forward" | "backward" | "left" | "right",
    granularity: "character" | "word" | "line"
    ): void;
}
```
By merging with the existing declaration?

bovine3dom:
This works great, thanks!

PR #5376, opened 2026-04-19
Add bottom indicator for key input visibility (fix #2966)

### LINKED ISSUE #2966: Add visual indicator to communicate when Tridactyl can receive inputs or not (page has focus)
Currently, Tridactyl is unusable if the page doesn't have focus, e.g. the Firefox chrome ui is focused, the address bar, etc.
As such, I think it should be as clear as possible when my inputs will work and when they won't.
This could be done in several ways, some probably harder than others..

- Hide the existing mode indicator. (hard to see)
- Dim the entire page.
- A red edge around the whole page.
- A red band at the bottom and/or top of the screen

shubhambudde1:
Added a visual indicator to show when Tridactyl can receive keyboard input.

Currently, when the page is unfocused (e.g., address bar or Firefox UI is focused), Tridactyl appears unresponsive with no clear feedback. This makes it confusing for users.

To improve this, I added a small bottom-left indicator bar that:
- Shows typed keys in real time
- Hides when the window loses focus
- Reappears when focus returns

This provides a simple and non-intrusive way to understand whether key inputs are being captured.

This approach avoids more intrusive options like dimming the page or adding borders, while still giving clear feedback.

Fixes #2966


bovine3dom:
Hi, unfortunately this PR does not fix the issue it claims to - it does not detect when focus is on Firefox chrome. I do not want to waste any more time reviewing this code, I strongly suspect you are either an AI agent or you have used an LLM to generate most of this content, so I don't feel too bad about closing this PR.

I have just added AGENTS.md, please ensure that any future contributions follow that guide.

Thanks!

PR #5381, opened 2026-04-22
Fix #1971: add listeners to same-origin iframes

### LINKED ISSUE #1971: Tridactyl doesn't work on Google Docs
Tridactyl version: 1.16.2pre2999-95db0a8
Firefox version: Mozilla Firefox 70.0
Operating system: linux

When I edit any public Google Docs document Tridactyl stays in `normal` mode even though I'm editing and actively inserting characters in the document. Since it's in `normal` mode all the time I can't use escape to go to `normal` mode and use any of the Tridactyl features.

*   Steps to reproduce:

1. Open Firefox,
2. go to e.g. [this document](https://docs.google.com/document/d/1OxHtRDS65d6Te2mbzCjNXzjFjeTdm8gAgPFCuwGENcg/edit),
3. type something, notice you're typing, notice you're in `normal` mode,
4. try to press `f` to go activate links following, notice it doesn't work.

https://docs.google.com

https://gist.github.com/cprn/2df156c5fba10ff41a17cc82c9f07a3c

*   Contents of ~/.tridactylrc or ~/.config/tridactyl/tridactylrc (if they exist):

```sh
$ ls .tridactyl .config/tridactyl
ls: cannot access '.tridactyl': No such file or directory
ls: cannot access '.config/tridactyl': No such file or directory
```

Chic-Tweetz:
#1971 discusses the escape key not working on google docs - this is mainly because listeners aren't actually being added to iframes' `contentWindow`s but there are related problems which must be fixed to have keypresses behave correctly

There are several problems with iframe key and focus listeners currently:
- adding the iframe listeners correctly
- discovering all iframes
- `thing instanceof Thing` being unexpectedly false in various places due to `thing` originating from an iframe
- `keydown` events having a different target to corresponding `keyup` events causing them not to be cancelled

this PR:

adds listeners to iframes `contentWindow`s instead of the iframe elements

uses mutation observers to watch for iframes being added to the page or inside other iframes

replaces some uses of `thing instanceof Thing` with equivalent functions that work with objects from iframes  - this lets the key canceller cancel key events from iframes

removes the `target` check from the key canceller because the equivalent `keyup` for a given `keydown` may have a different `target` - particularly bad in cases like google docs where pressing escape to unfocus the document unfocuses it, then a `keyup` listener refocuses it! This may not be the best solution, that check may have been important in some cases?

---

also added an `activeElement` function which delves into iframes so the mode indicator can go to insert mode if a text editable element is focused in an iframe

### DIFF
--- a/src/content.ts
+++ b/src/content.ts
@@ -196,6 +196,7 @@ function listenInIframe(frame: FrameElement) {
         if (!doc?.defaultView || observedIframeRoots.has(doc)) return
         listen(doc.defaultView)
         observeIframeRoot(doc)
+        dom.hijackPageAttachShadow(observeIframeRoot, doc.defaultView)
         dom.setupFocusHandler(doc)
     } catch (e) {
         logger.warning("Could not hijack iframe due to CSP:", e)
@@ -304,6 +305,7 @@ window["tri"] = Object.assign(Object.create(null), {
 logger.info("Loaded commandline content?", commandline_content)
 
 try {
+    dom.hijackPageAttachShadow(observeIframeRoot)
     dom.setupFocusHandler()
     dom.hijackPageListenerFunctions()
 } catch (e) {
--- a/src/content.ts
+++ b/src/content.ts
@@ -173,10 +173,55 @@ function listen(elem: Window | HTMLElement | HTMLFrameElement) {
         true,
     )
 }
+
 listen(window)
-document.addEventListener("readystatechange", _ =>
-    getAllDocumentFrames().forEach(f => listen(f)),
-)
+listenInIframes()
+document.addEventListener("readystatechange", listenInIframes)
+createIframeObserver(document)
+
+// Add listeners to existing iframes
+function listenInIframes() {
+    getAllDocumentFrames().forEach(f => {
+        try {
+            if (f.contentDocument?.readyState === "complete") {
+                listen(f.contentWindow)
+                dom.setupFocusHandler(f.contentDocument, listen)
+                createIframeObserver(f.contentDocument)
+            } else {
+                f.addEventListener("load", () => {
+                    listen(f.contentWindow)
+                    dom.setupFocusHandler(f.contentDocument, listen)
+                    createIframeObserver(f.contentDocument)
+                })
+            }
+        } catch (e) {
+            logger.warning("Could not hijack iframe due to CSP:", e)
+        }
+    })
+}
+
+// Add listeners to iframes added to the DOM
+function createIframeObserver(doc = document) {
+    const iframeObserver = new MutationObserver((mutations, observer) => {
+        for (const mutation of mutations) {
+            if (mutation.type === "childList") {
+                for (const node of mutation.addedNodes) {
+                    if (node instanceof doc.defaultView.HTMLIFrameElement &&
+                        !node.src.startsWith("moz-extension:")
+                    ) {
+                        try {
+                            listen(node.contentWindow)
+                            dom.setupFocusHandler(node.contentDocument, listen)
+                            createIframeObserver(node.contentDocument)
+                        } catch(e) {}
+                    }
+                }
+            }
+        }
+    })
+    iframeObserver.observe(doc, { subtree: true, childList: true })
+    return iframeObserver
+}
 
 // Prevent pages from automatically focusing elements on load
 config.getAsync("preventautofocusjackhammer").then(allowautofocus => {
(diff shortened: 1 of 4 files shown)


src/content/controller_content.ts
```diff
@@ -46,6 +46,11 @@ function PrintableKey(k) {
     return result
 }
 
+function isKeyboardEvent(ke: any): ke is KeyboardEvent {
```
Chic-Tweetz: `ke instanceof KeyboardEvent` is equivalent to `ke instanceof window.KeyboardEvent` but `window.KeyboardEvent` won't be the same as `iframe.contentWindow.KeyboardEvent`

this is an iframe-friendly workaround

src/lib/dom.ts
```diff
@@ -618,10 +621,34 @@ export function setupFocusHandler(): void {
         }
         setFocus(elem)
     }
-    listen(document)
+
+    if (!focusListenerDocs.has(doc)) {
+        listen(doc)
+
+        // Use focusout to check if we've shifted focus to a new iframe we're yet to add listeners to
```
Chic-Tweetz: when a window is unfocused, it may be because focus was moved to an iframe inside that window. We can loop through iframes and see if there's any we haven't discovered yet, adding listeners to new ones

src/content.ts
```diff
@@ -137,32 +137,67 @@ const guardedAcceptKey = (keyevent: KeyboardEvent) => {
 }
 function listen(elem) {
     elem.removeEventListener("keydown", guardedAcceptKey, true)
-    elem.removeEventListener(
```
Chic-Tweetz, line 145: tangentially related, cancelled `keydown` events will automatically prevent `keypress` events

src/content.ts
```diff
@@ -367,7 +402,7 @@ config.getAsync("modeindicator").then(mode => {
         statusIndicator.className =
             "cleanslate TridactylStatusIndicator " + privateMode
         if (
-            dom.isTextEditable(document.activeElement) &&
+            dom.isTextEditable(dom.activeElement()) &&
```
Chic-Tweetz: mode indicator will display "insert" if text editable elements are focused in iframes

src/lib/dom.ts
```diff
@@ -584,10 +587,10 @@ function hijackPageFocusFunction(): void {
         }
      })(HTMLElement.prototype.focus, ${exportedName})`
 
-    window.eval(eval_str + `;delete ${exportedName}`)
+    win.eval(eval_str + `;delete ${exportedName}`)
 }
 
-export function setupFocusHandler(): void {
+export function setupFocusHandler(doc = document, onNewIframeFound = null): void {
```
Chic-Tweetz: added parameters so iframe documents can be passed and `listen` from content.ts can be used as a callback when new iframes are discovered

src/content/controller_content.ts
```diff
@@ -90,10 +95,9 @@ class KeyCanceller {
                 ke.composed === ke2.composed &&
                 ke.ctrlKey === ke2.ctrlKey &&
                 ke.metaKey === ke2.metaKey &&
-                ke.shiftKey === ke2.shiftKey &&
-                ke.target === ke2.target,
```
Chic-Tweetz, line 91: `keyup`s wouldn't be cancelled if the `keydown` changes the current focus

bovine3dom: i _think_ this is a security feature, otherwise a genuine event can be hijacked and followed by a spoof event on a different target?

i guess it doesn't matter if the keys are all forced to be the same 🤔

bovine3dom (commented):
this might be great but i am kind of terrified of touching the key event listeners because they're our main interface with the web, and the web is a hostile environment

and while "working on google docs" would be nice, "tridactyl gets another cve" would be less so

so i think what i'll do is try to improve our tooling around ensuring that key events are real and not spoofed by websites, and then we can rebase this on that and i will be slightly less cowardly...

Chic-Tweetz:
> and while "working on google docs" would be nice

ehhhh, it's alright 😄 it's all too clever and custom to just work nicely with tridactyl without extra effort

i don't think this should let tridactyl be fooled by spoofed key down events as long as we can trust the constructors in iframe window objects

but definitely understandable to be on the safe side

src/content/controller_content.ts
Chic-Tweetz: i'm not sure whether it's for security or just matching key ups with key downs

i should probably look for discussion from when this was first written

PR #5383, opened 2026-04-25
Add TrustedKeyboardEvent and compel usage

bovine3dom:
I got caught out by my own pre-commit hook and it mangled this to fix the formatting. I've tried my best to split it into two commits but didn't manage it perfectly.

~~This currently breaks `keyfeed` which is 1) great, those are untrusted key events but 2) it would be nice to make it work again~~

Blocks #5381

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -61,6 +61,9 @@ module.exports = {
                     "String": {
                         "message": "Avoid using the `String` type. Did you mean `string`?"
                     },
+                    "KeyboardEvent": {
+                        "message": "Use `TrustedKeyboardEvent` to prevent remote code injection from hostile pages."
+                    },
                     "Symbol": {
                         "message": "Avoid using the `Symbol` type. Did you mean `symbol`?"
                     }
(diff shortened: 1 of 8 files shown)


.eslintrc.js
```diff
@@ -270,5 +273,23 @@ module.exports = {
...
+                    "error",
+                    {
+                        "types": {
+                            "TrustedKeyboardEvent": {
+                                "message": "Events must be validated with `isTrustedKeyboardEvent` at runtime"
+                            },
+                            "KeyboardEvent": {
+                                "message": "Use `Event` instead"
+                            },
+                        }
```
bovine3dom: anywhere using `addEventListener` directly mustn't use TrustedKeyboardEvents because they're implicitly `any` and bypass all the checks

maybe there's a better way of doing it than white/blacklisting them by file

(also this override removes all the other ban-types types, it probably shouldn't)

src/lib/keyseq.ts
```diff
@@ -26,9 +26,31 @@ import { Parser } from "@src/lib/nearley_utils"
...
+// this should be ~the only place in the code that accepts KeyboardEvent
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type TrustedKeyboardEvent = KeyboardEvent & { readonly isTrusted: true }
+
+const _guarded =
+    (f: (ke: TrustedKeyboardEvent) => void) => (maybekeyevent: Event) => {
+        if (!isTrustedKeyboardEvent(maybekeyevent)) return
+        return f(maybekeyevent)
+    }
+export const guarded = memoise(_guarded)
```
bovine3dom: worth double checking whether this memoisation really does let us cancel the event listeners we have added

Issue #4970, opened 2024-04-19
Not able to unbind s in visual mode


pzzxqj:
After typing `:unbind --mode=visual s`,  `s` still exists in visual map but not appear `:viewconfig`.

[20240419_153601.webm](https://github.com/tridactyl/tridactyl/assets/107164025/41c12804-d285-4455-9336-c03b72c9f6b8)

version: 1.24.1

GodSaveEarth:
Yes, I confirm that. Unbinding any key bind (visual mode) restores it to default command, instead of clear.

GodSaveEarth:
**Seems working now.** I moved to new Firefox installation and updated Tridactyl to 1.24.2 and the problem has gone.

bovine3dom:
@GodSaveEarth that will set vmaps to the four character string `"{}"` which, honestly, i don't really know what it would do. There's not any particularly ergonomic way to totally wipe out a mode like you're suggesting.

Why do you want to remove all of the default binds? If you don't like visual mode you can disable it with `:set visualenterauto false`

bovine3dom:
Great! Closing as I think the problem was down to a mangled user config :)

GHolk:
@bovine3dom @GodSaveEarth @pzzxqj 

I figure out that if `tri.config.USERCONFIG.vmaps['🕷🕷INHERITS🕷🕷']` is set,
unbind will not make `tri.config.get('vmaps').s` become null.

For anyone want to fix this temporarily, try this command:
`js delete tri.config.USERCONFIG.vmaps['🕷🕷INHERITS🕷🕷']; tri.config.set('configversion', '2.0')`

The last set command can set anything.
It means to force tridactyl update the config cache.

GHolk:
> I figure out that if `tri.config.USERCONFIG.vmaps['🕷🕷INHERITS🕷🕷']` is set, unbind will not make `tri.config.get('vmaps').s` become null.
> 
> For anyone want to fix this temporarily, try this command: `js delete tri.config.USERCONFIG.vmaps['🕷🕷INHERITS🕷🕷']; tri.config.set('configversion', '2.0')`

This happen because `src/lib/config.ts` calls `mergeDeepCull` to handle inherits, and mergeDeepCull will remove the null property (the unbinded one); then the default one get chose.

bovine3dom:
Yeah. We were talking about this here right https://discord.com/channels/854326924402622474/854326925031374860/1418230643787169882

[...]

PR #5389, opened 2026-05-20
fix: unbind default and default config overwritten by mergeDeep

GHolk:
The mergeDeepCull set a unbind key (null) to undefined. In get and getURL, there are multiple merge operations, and undefined is ignored in the following merge.

We still need to remove null because null cause problem in keymap. In most situation, we access the config with `tri.config.get` method, so remove removeNull at return statement in get method is enough.

### DIFF
--- a/src/lib/config.test.ts
+++ b/src/lib/config.test.ts
@@ -1,7 +1,11 @@
 import { testAll, testAllObject } from "@src/lib/test_utils"
 import { canonicaliseMapstr } from "@src/lib/keyseq"
 import { default_config } from "@src/lib/config"
+import * as tri_config from "@src/lib/config"
 import { zip } from "ramda"
+
+const tri = { config: tri_config }
+const { getURL, get } = tri.config
 
 const config = new default_config()
 // todo: test subconfigs and platform_defaults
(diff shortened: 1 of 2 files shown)


GHolk:
* The deepMerge fix may fix #4336 
* The mergeDeepCull fix may fix #4970 
* Replace the legacy draft #4370 

## TODO
Maybe we should replace the strange Object.create function call
in modern JavaScript.

## Test In Beta?
Because this change Tridactyl's low level behavior,
maybe we should test in Beta channel for a longer time?

Or do I have to test it in my daily useing before merge into Beta channel?

Please convert this into Draft if required.

GHolk:
## TODO
Check all `/config.get(URL|.+)\(/` function call and make sure this PR will not break them.

bovine3dom:
by strange object.create call do you mean

```js
export function o(object) {
    return Object.assign(Object.create(null), object)
}
```

my recollection is that it's necessary to get a sensible object without any extra weird keys in it. i think the 'modern JS' way around it is to use a `Map` instead. `o()` avoids stuff like

```
> "toString" in {}
true
```

GHolk:
I notice an old issue #4763 and PR #4820 
It seem that the max history size is not very neccessory,
but ignore exstr starting with space is still usefull.

Should I also pick the commit 4b407b464e74ba7c25620ac2a444de769aed2132 ?

bovine3dom:
I'd rather "space skips history" went in a different PR because this PR is scary and might break things, whereas "space skips history" should be a simple change that never needs reverting

GHolk:
This command show there are only few `tri.config.getURL()` function calls outside `src/lib/config.ts`

```sh
egrep -r 'config.get(URL|.+)\(' |
    sed -r 's/config\.(get|getAsync|getDynamic|getAsyncDynamic)\(//' |
    less '+/config\.'
```

Therefore there are nothing we need to modify, in reply to

> TODO: Check all `/config.get(URL|.+)\(/` function ...

bovine3dom:
is this ready to merge do you think? it looks fine to me and in my limited testing seems to work ok

GHolk:
I use this build in home for 4 days, and the mktridactylrc output is so far so good,
but I notice there are still some problem in custom mode.

I will try to fix it and make a new build latter.

GHolk:
I'm confused that mergeDeep detect `o1 === null` and `o2 === undefined` .

Anyway I will rewrite the get logic that merge 3 only if there are more than 1 object
and all the others are `[object|null|undefined]` .

What kind of situation make it undefined or null?

bovine3dom:
`null` is how we let users 'delete' bits of inherited config, e.g. `unbind` and `unbindurl`, `setnull`

`undefined` means that there is nothing in the inherited config to override

GHolk:
Given `[defult, user, site]`:

* custom mode: `[undefined, object, object|undefined]`
* followpagepatterns: `[object, object|undefined, object|undefined]`

The last commit fix the custom mode case
`[undefined, object, object]` .

I can only imgine these 2 special cases.

I don't know how should we handle merge in a general way,
so I just fix the custom-mode case and make sure that
the follow-page-patterns still work.

Beside, pass string arguments to the mergeDeep function
will return a broken object.

the built-in mode is `[object, object|undefined, object|undefined]`
which always works fine.

GHolk:
@bovine3dom I think this PR is ready now. I don't encounter any bug in these days.

bovine3dom:
i'll try to remember to let this gestate in beta for a while before we put it in a stable version. thanks!

Issue #5266, opened 2025-09-17
scrollpage (Ctrl+F) results in glitchy state where webpage is endlessly micro jittering up/down, and unable to use other scroll methods.


exaiyo:
Tridactyl 1.24.4
Firefox 142.0.1 (64-bit)
Void Linux (tested on both AMD drivers and Nvidia proprietary)

I use CTRL+F, CTRL+B, etc, for most of my scrolling, but this common glitching has destroyed my browsing workflow.
I've tried disabling smoothscroll in tridactyl, and then altogether in firefox with general.smoothScroll = false, and the problem remains.
The glitch occurs in Flatpak Firefox and in repository firefox, even with fresh profiles with no other addons or user customizations of any kind.
This only started happening recently, I'm guessing in the last month or so, but I haven't had time to investigate further until now.

I've thought maybe I need to try something else like Vimium, but it's hard to imagine browsing with anything other than tridactyl...

Let me know if there is any other testing I can do that would be beneficial.

EDIT: In the meantime, if  anyone else is suffering this glitch, scrollline is seemingly unaffected. You can use/tweak the following binds to (poorly) approximate the scrollpage functionality. Better than glitching for sure:
```
bind <C-f> scrollline 28
bind <C-b> scrollline -28
bind <C-d> scrollline 16
bind <C-u> scrollline -16
```

bovine3dom:
I can't reproduce this on BBC News :(

Are there any other websites that either of you can reproduce it on?

bovine3dom:
Brilliant, thanks. Can reproduce after `xrandr --dpi 120` on a 4K monitor.

(it's horrible)

xiangsheng:
I use `scrollline 10` and also meet this bug. And I notice that when webpage starts jittering, I can't use the mouse to scroll the website any more. I have to run `scrollline 10` several times to get out of the jittering scenes and then the mouse works again.

YggdrasiI:
I'm also affected (Tridactyl 1.24.4, Firefox 145.0 (64-bit), Manjaro, i3-wm, AMDGPU driver, AMD Ryzen 5 5600G) and glad to find this issue by accident. 

I was using dpi 96 so maybe you could reproduce it with this value, bovine3dom.

Had the font settings of FF some effect? I'm using Noto Sans 24 and disallow other fonts for rendering.

bovine3dom:
Closing as in theory fixed in #5392

PR #5392, opened 2026-05-26
Fix display dpi-dependent scroll jitter

Chic-Tweetz:
This addresses #5266

Actual values that can be set for `Element.scrollTop` / `scrollLeft` can change depending on display scaling.

For instance, if I set my display scale to 125% and then run this repeatedly on this page

```
document.scrollingElement.scrollTop = 210
document.scrollingElement.scrollTop
```

the actual value of `document.scrollingElement.scrollTop` alternates between `209.600...` and `210.399...`

Because the value does change, when scrolling with Tridactyl, `scrollStep` always returns `true` and steps are infinitely scheduled, bouncing between the two possible values around the unreachable one.

This pull request adds a check in `scrollStep` to see if we try to scroll to `endPos` and if so, sets `endPos` to the actual value, so that on the next `scrollStep` that definitely reachable value will be the target and it can return `false`.

I also altered the checks in `getStep` which make sure progress is made as I found that smooth scrolling could stop early due to `pixelToScrollTo` accidentally ending up in the wrong direction through unfortunate rounding.

### DIFF
--- a/src/content/scrolling.ts
+++ b/src/content/scrolling.ts
@@ -80,15 +80,17 @@ class ScrollingData {
             this.startPos +
             ((this.endPos - this.startPos) * elapsed) / this.duration
         if (this.startPos < this.endPos) {
+            const curPosRounded = Math.ceil(this.elem[this.scrollDirection])
             // We need to ceil() because only highdpi screens have a decimal this.elem[this.pos]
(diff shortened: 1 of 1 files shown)


bovine3dom:
Brilliant, LGTM, thanks!

PR #5395, opened 2026-05-28
Generate excmd metadata using typedoc

bovine3dom:
> @bovine3dom I don't know if you still have this issue, but I wrote a patch that enables getting rid of .metadata.generated.ts by using typedoc's JSON output instead. It has a few advantages (namely enable upgrading the typescript version Tridactyl uses), but quite a few downsides too (build is slower, extension gets heavier due to the metadata getting embedded in both the content and background script). Feel free to ignore it, I'm not particularly attached to it, this was mostly an experiment to satisfy my curiosity :)


bovine3dom:
```
Code             Message       Description             File         Line   Column
FILE_TOO_LARGE   File is too   This file is not        content.js                
                 large to      binary and is too                                 
                 parse.        large to parse. Files                             
                               larger than 4MB will                              
                               not be parsed.                                    
                               Consider moving large                             
                               lists of data out of                              
                               JavaScript files and                              
                               into JSON files, or                               
                               splitting very large                              
                               files into smaller                                
                               ones.                                             
```

oops

it feels like there's a huge amount of guff in `.metadata.generated.json` and that we ought to be able to walk the tree and only keep the bits we need. but right now i can't immediately work out what we need and what we don't

---

edit: fixed

kind of annoyed that it was as simple as `JSON.stringify`. there's probably more to be got out of binning unused fields but it's already good enough to stop failing Mozilla's linting

bovine3dom:
i found some actual bugs with get/set and wrong types and fixed them but that suggests to me that there are more scary bugs hidden away somewhere.

i would also like to fix our fast rebuilds - previously they didn't mind metadata being a bit stale but now they get very upset

PR #5404, opened 2026-06-17
User define completions

GHolk:
Similar syntax as jsurls.

How ever I found a problem that javascript code in `commandline_frame.ts`
can not access webpage's dom. This make usercompletions very limited.

I found that src/completions/Goto.ts use message call to retrieve dom content.
Maybe I should do so for usercompletions.

This revision will not pass the linter and typescript's type checker.

### DIFF
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -738,6 +738,8 @@ export class default_config {
         prev: "^(prev(ious)?|older)\\b|«|<<",
     }
 
+    usercompletions = {}
+
     /**
      * The default search engine used by `open search`. If empty string, your browser's default search engine will be used. If set to something, Tridactyl will first look at your [[searchurls]] and then at the search engines for which you have defined a keyword on `about:preferences#search`.
      */
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -738,7 +738,7 @@ export class default_config {
         prev: "^(prev(ious)?|older)\\b|«|<<",
     }
 
-    usercompletions = {}
+    completionscustom = {comp: "()=>{}"}
 
     /**
      * The default search engine used by `open search`. If empty string, your browser's default search engine will be used. If set to something, Tridactyl will first look at your [[searchurls]] and then at the search engines for which you have defined a keyword on `about:preferences#search`.
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -1919,14 +1919,14 @@ export function getDeepProperty(obj, target: string[]) {
             return getDeepProperty(obj[target[0]], target.slice(1))
         } else {
             return getDeepProperty(
-                mergeDeepCull(get(obj["🕷🕷INHERITS🕷🕷"]), obj)[target[0]],
+                mergeDeep(get(obj["🕷🕷INHERITS🕷🕷"]), obj)[target[0]],
                 target.slice(1),
             )
         }
     } else {
         if (obj === undefined || obj === null) return obj
         if (obj["🕷🕷INHERITS🕷🕷"] !== undefined) {
-            return mergeDeepCull(get(obj["🕷🕷INHERITS🕷🕷"]), obj)
+            return mergeDeep(get(obj["🕷🕷INHERITS🕷🕷"]), obj)
         } else {
             return obj
         }
@@ -1965,9 +1965,7 @@ export function mergeDeep(o1, o2) {
             key => typeof o1[key] === "object" && typeof o2[key] === "object",
         )
         .forEach(key =>
-            r[key] == null
-                ? null
-                : Object.assign(r[key], mergeDeep(o1[key], o2[key])),
+            r[key] == null ? null : (r[key] = mergeDeep(o1[key], o2[key])),
         )
     return r
 }
@@ -2010,7 +2008,7 @@ export function getURL(url: string, target: string[]) {
     const deflt = _getURL(DEFAULTS, url, target)
     if (user === undefined || user === null) return deflt
     if (typeof user !== "object" || typeof deflt !== "object") return user
-    return mergeDeepCull(deflt, user)
+    return mergeDeep(deflt, user)
 }
 
 /** Get the value of the key target.
@@ -2036,14 +2034,14 @@ export function get(target_typed?: keyof default_config, ...target: string[]) {
 
     // Merge results if there's a default value and it's not an Array or primitive.
     if (typeof defult === "object") {
-        return mergeDeepCull(mergeDeepCull(defult, user), site)
+        return removeNull(mergeDeep(mergeDeep(defult, user), site))
     } else {
         if (site !== undefined) {
-            return site
+            return removeNull(site)
         } else if (user !== undefined) {
-            return user
+            return removeNull(user)
         } else {
-            return defult
+            return removeNull(defult)
         }
     }
 }
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -738,7 +738,7 @@ export class default_config {
         prev: "^(prev(ious)?|older)\\b|«|<<",
     }
 
-    completionscustom = {comp: "()=>{}"}
+    completionscustom = {}
 
     /**
      * The default search engine used by `open search`. If empty string, your browser's default search engine will be used. If set to something, Tridactyl will first look at your [[searchurls]] and then at the search engines for which you have defined a keyword on `about:preferences#search`.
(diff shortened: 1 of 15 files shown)


GHolk:
Uhm, some old commits show up here :thinking: 

Syntax example:

```
set usercompletions.foo (argv,ctx)=>{
    if (!ctx.comp) ctx.comp= [
        'arg1',
        ['option2', 'display text 2']
    ];
    return ctx.comp
}
```

bovine3dom:
> I found that src/completions/Goto.ts use message call to retrieve dom content.

i think in theory the `tabsProxy` should make that a bit easier, but I haven't used it for a while so I can't remember how it works

src/lib/config.ts
```diff
@@ -738,6 +738,8 @@ export class default_config {
         prev: "^(prev(ious)?|older)\\b|«|<<",
     }
```
bovine3dom, line 740: please don't forget to give this a good docstring :) . we should bikeshed the name a bit, we try to use vim-style pseudonamespacing where the category comes first

so `completionscustom` e.g.

Chic-Tweetz:
> How ever I found a problem that javascript code in `commandline_frame.ts` can not access webpage's dom. This make usercompletions very limited.

I had a crack at this (apparently 5 months ago 😮) and added a [completions_content.ts](https://github.com/Chic-Tweetz/tridactyl/blob/custom-completions/src/content/completions_content.ts) to handle the messaging between the iframe and the page. 

The main idea was to message with the prefix and reply with the options (and anything else from the config), that's `response` in `get_custom_completion`. Don't know if it'll be helpful at all, but take inspiration from that if you like

GHolk:
I move the completion generation into content-script,
because I believe most case user want to generate completion from page content.

Moreover, I'm rethinking the API.
Current API is just a wrapper around the existing completion API,
which is not very suit to my imaginary use case.

My first use case is to choose drop-down menu with completion api.

 1. `hint -c selection -F e=>{ var value = [], display = []; /* some code */; tri.excmds.completionsetup('--fuse=true -q #', e.id, '--complete', value.concat(display).join(' '))}`
 2. the excmd open cmdline, and show menu item as completion
 3. user can select item in ex mode.

I want to design a elegant API that be able to handle this task.
Im wondering whether we should design a `hint -F` like api
or make completion selection execute a normal ex command.

## approach 1
code example:

```
# execute this
completionsetup -F value=>$('#'+tri.compid).value=value -v value1 -p preview%20one -v value2 ...

# fillcmdline with this generate following completion
completionfill -F value=>$('#'+tri.compid).value=value
# preview one
# preview tow
# ...
```

In this case, user have to pass context object in other way,
because completionfill only recognize string arguments.

## approach 2

Or make the `completionfill` a **special** non-idempotent command,
that can not execute directly, but can only be triggered with `completionsetup` command,
and its behavior is dependent on the setup command.

---

I'm busy with dragon boat festival this weekend,
so previous commit is not test at all.
I may not have time to work on this this weekend.

src/lib/config.ts
GHolk: I am fine to follow vim-style naming convension :)
Juse remember to hint me because I dont use vim directly. I use emacs evil ;)

GHolk:
> I had a crack at this (apparently 5 months ago 😮) and added a [completions_content.ts](https://github.com/Chic-Tweetz/tridactyl/blob/custom-completions/src/content/completions_content.ts) to handle the messaging between the iframe and the page.

It is a very comprehensive implementation!
Will you work on it?

GHolk:
proposal, `completionsetup` and `comp` commands.
Shortly, I want to add a [fzf] like interface.

[fzf]: https://github.com/junegunn/fzf/

## javascript api
completionsetup will return a object
containing a promise field which resolved with the user input.

### easy example

    // fillcmdline with `comp -n `
    const exec = tri.excmds.completionsetup({
        action: null,
        data: 'youtube.com google.com'.split(' '),
        preview: ['YouTube', 'Google Inc']
    })
    // user choose *YouTube*, the exstr become `comp -n youtube.com `
    const d = await exec.promise
    // user press Enter, the promise resolve with 'youtube.com'
    tabopen(d)

### drop-down menu example

    command drop-down-menu-prompt hint -c select -F async e=>{
        const data=[], preview=[]
        for (const opt of e.options) {
            data.push(o.value)
            preview.push(o.textContent)
        }
        const exec = tri.excmds.completionsetup({
            action: null,
            data, preview
        })
        const v = await exec.promise
        e.value = v
    }

## excmd style api
this command will open cmdline with 2 completion option.

* `-d`: *real* value to insert when user select a completion
* `-p`: preview, optional
* `-x`: the exstr to execute
* `-F`: the js callback function to execute
* `-n`: do not execute everything, just resolve the promise with value

```
completionsetup -x tabopen -d youtube.com -p YouTube -d google.com -p Google%20Inc
```

the cmdline will fill with following exstr and 2 completions

    comp -x tabopen
    # YouTube
    # Google Inc

after user select Google Inc and call ex.insert_completion

    comp -x tabopen google.com

the comp command will simply execute the -x option
so the previous command is same to

    tabopen google.com

the `-x -F -n` options here are same to the `completionsetup` options

Chic-Tweetz:
> It is a very comprehensive implementation! Will you work on it?

I have been meaning to. It's funny though because since I've had it working I haven't actually used it 😄 so I have neglected it a bit.

I was planning on reimplementing existing completions in a way that would let them be easily reused for custom completions, but that feels like it'll take a bit of effort so I never started :)

GHolk:
Fine, the feature is finished now. We just need to type them correctly and refactor.

To build, use the `npm run rebuild` command which will accept all un-typed code.

## discuss
As `getGotoSelectors` , I put them in excmds.ts initially,
should I move the non-excmd code to `contents/completion_custom.ts` or somewhere?

Chic's branch makes use of `commandline_content` message type.
Should I use it too?

Now there are two features, completionsetup and `completionscustom.*` .
They interlace in content script (excmd) and background script (CompletionsCustom.ts) .
I think it is possible to split the background part.
Should I do that?

I am also thinking about adding a callback style completion interface inside Tridactyl.
Should I add it in this PR?

@bovine3dom @Chic-Tweetz how do you think?

bovine3dom:
this looks good as a stop-gap before we do #105 (lol)

i'm trying to get my head around `:comp`

is the idea that it is used for 'temporary' completions via `:completionsetup` and permanent ones are set via `:set completionscustom.foo` which then would provide completions for `:foo` ?

Chic-Tweetz:
I do like the `:completionsetup` idea, I imagine it'd be best used in conjunction with `:js` if you wanted to populate it with more than a few choices?

I haven't figured out how to get `completionscustom` to actually display using the `mygoto` example (I was expecting options to display if I type `:mygoto ` - is that right?)

> I am also thinking about adding a callback style completion interface inside Tridactyl. Should I add it in this PR?

If this is similar to the callback stuff I made, I'd like to add the juicy detail that a callback that triggers as you type into the command line is how I got `incsearch` working :) ... another thing I should get back to!

GHolk:
I have some examples in js-doc, `completionscustom` in lib/config.ts
and `completionsetup` in excmds.ts .

@bovine3dom Does it still worth today?

`comp` is a placeholder command. For example,
after `completionsetup` , we can not fillcmdline with `tabopen`
or it will trigger tabopen's completion, so I use `comp -x tabopen` .

`comp` do two things:

 1. execute the excmd of `-x` option or js code of `-F` option
 2. Trigger the **callback** (`-n`) defined in the `completionsetup({callback})` and clear it

There is no magic in the first situation,
so you can summon it from the cmdHistory and execute it again.

In second situation, it will not work unless we execute
`completionsetup({callback})` again before it.

---

About `completionscustom.foo` , there have to be a `foo` command.

One can define foo as an alias of `tabopen` or other excmd,
(alias will trigger origin command's completion before the fix in 1ca9c41)
or as an alias of `comp -x tabopen` , or any js command.

Another way, one can define the **callback** by set `option.callback`
(in the `set completionscustom.foo (argv,ctx,option) => { option.callback = ... }` )
and make `foo` as an alias to `comp -n` .

The **callback** style magically keep the closure context; we sometimes need this.

The first `-x|-F` has no magic and has no closure context too.

---

@Chic-Tweetz Right, `completionscustom.mygoto` work as you wrote.

I may not add many details about the data-structure of completion.
The callback style completion API is to let developer
define completions like `completionscustom.*` is tridactyl's source code,
because the current CompletionSource API is a little verbose to me.

bovine3dom:
hey @GHolk every few days i look at this and then realise it's really complicated and ... find something else that's easier to work on :sweat_smile: 

could you summarise 1) things you would like me to help you decide and 2) things i should copy and paste into the command line to test?

ideally in simple words that my caveman brain can understand

sorry and thank you :heart:

PR #5429, opened 2026-07-09
Continous scroll excmds for keydown/keyup binds

Chic-Tweetz:
This adds two excmds: `scrollstart` and `scrollstop`, intended to be bound to keydowns and keyups respectively.

Pass `scrollstart` an x and y velocity (in px / second), with an optional multiplier like `:scrollline` so a number key can be pressed first to increase the speed.

Example binds for `hjkl` scrolling:

```
bind <D-j> scrollstart 0 1000
bind <D-k> scrollstart 0 -1000
bind <D-h> scrollstart -1000 0
bind <D-l> scrollstart 1000 0

bind <U-j> scrollstop
bind <U-k> scrollstop
bind <U-h> scrollstop
bind <U-l> scrollstop
```

The default `hjkl` binds would shadow these so would need to be unbound.
```
unbind h
unbind j
unbind k
unbind l
```


src/excmds.ts
```diff
@@ -6516,4 +6517,31 @@ export function hintstylesdirect() {
...
+/**
+ * Set the x and y velocities to scroll continously until `:scrollstop` is called.
+ *
+ * Bind `:scrollstop` to the equivalent keyup:
+ * ```
+ * :bind <D-j> scrollstart 0 1000
+ * :bind <D-k> scrollstart 0 -1000
+ * :bind <U-j> scrollstop
+ * :bind <U-k> scrollstop
+ * ```
```
bovine3dom, line 6536: or use `:smoothscrollwizard` (i didn't catch this)

bovine3dom (commented):
this looks great, thanks! i need to check that it works on some of the most pathological pages before i merge it. the hard bit is finding those pages :)

src/excmds.ts
Chic-Tweetz: `:smoothscrollwizard` is a good addition, much handier!

Makes me wonder if a base speed should be set in config like `:set scrollspeed 1000` to make it easier to adjust all binds at once (all the different `1000`s in the binds would become `1`s, `5000` becomes `5` etc.)

Or just as an arg for `:smoothscrollwizard` eg `:smoothscrollwizard 1000`

bovine3dom: we could (should?) reuse scrollduration https://tridactyl.xyz/build/static/docs/classes/_src_lib_config_.default_config.html#scrollduration as a divisor

Chic-Tweetz: good shout, should be a quick change

PR #5430, opened 2026-07-10
Fix #639: load iframe lazily

### LINKED ISSUE #639: Option to inject iframe lazily

bovine3dom:
this is a bit of a scary change but i haven't yet found any bugs it has caused


bovine3dom:
so far the only website i can confirm it has fixed is https://codeforces.com/

all the other ones i can see in that thread seem to be working in vanilla tridactyl for me

Issue #5372, opened 2026-04-15
misplaced hints / hints for hidden links


aavogt:
https://hackage-content.haskell.org/package/template-haskell-2.24.0.0/docs/Language-Haskell-TH-Syntax.html#t:Match press f
<img width="516" height="359" alt="Image" src="https://github.com/user-attachments/assets/642183c7-7edf-4575-961e-f40f8e49e3ef" />

For example the AD hint there points to https://hackage-content.haskell.org/package/template-haskell-2.24.0.0/docs/Language-Haskell-TH-Syntax.html#t:Exp probably coming from the `Exp` inside `Lift Match` which is not visible.

Higher up there aren't as many extra hints:

<img width="213" height="240" alt="Image" src="https://github.com/user-attachments/assets/2eda5b97-77d9-4d60-8920-4770f166fde2" />

At the very top there aren't any extra hints.

Chic-Tweetz:
there was a [similar issue](https://github.com/tridactyl/tridactyl/issues/5350) about reddit comments

tridactyl does a few checks for element visibility, but some things aren't always caught like collapsed elements

since that issue I've been using a [different function](https://github.com/Chic-Tweetz/tridactyl/blob/63a8c7ca2ca8243c85509b66b5cf07745d6602bd/src/lib/dom.ts#L215) to check for visibility which uses an `IntersectionObserver`

I had forgotten about that... which means it must be working well enough to not cause trouble!

I'll make a pull request for that at some point because it does work well for pages like this one:

<img width="521" height="373" alt="Image" src="https://github.com/user-attachments/assets/93f57392-0622-4103-8296-b3dc76050b74" />

bovine3dom:
Fixed in #5434

Issue #5350, opened 2026-02-27
[Reddit] Collapsed comments create lots of hint targets


vulns-are-features-too:
## Description

On Reddit, when there are collapsed comments, using hints shows many hints to non-visible elements
(This wasn't an issue some before, maybe a few days ago)

## System info: 

- OS: Arch Linux
- Mozilla Firefox 147.0.4
- tridactyl v1.24.4

## Reproduction steps:

1. Open a post on https://www.reddit.com with multiple top-level comments
2. Collapse some comments
3. Use `f` or `F` hint

## Result

Before hint:

<img width="1276" height="731" alt="Image" src="https://github.com/user-attachments/assets/43350111-93b9-4cba-aa22-a2e4dd8112b1" />

After hint:

<img width="1292" height="874" alt="Image" src="https://github.com/user-attachments/assets/c4d2ebfa-b06e-4532-a46c-9c6b272faf08" />

Chic-Tweetz:
It might be worth switching to using an `IntersectionObserver` to check for visible elements - it works with reddit's collapsed comments and appears to be (quite a lot) more performant too

PR #5434, opened 2026-07-12
Use `IntersectionObserver` to check element visibility

Chic-Tweetz:
Addresses issues #5350, #5372 where lots of unhelpful hints show up on some sites.

This adds a `getVisibleElemsBySelector` function to `dom.ts` which uses `IntersectionObserver`s to decide whether elements are visible.

The `IntersectionObserver` technique doesn't work for elements with a `visibility: hidden;` rule so there's also an `isPainted` function to filter such elements out.

The result is that some elements which would be considered visible are correctly ignored, making hinting much more usable on certain sites:

current:
current

intersection observer:
intersection-observer

Intersection observers have no entries when first created so the function is async to resolve once it's populated. This means that to use it for hints, some other `async`s / `awaits` are sprinkled around.

Mostly that shouldn't be an issue, but it would break any user scripts which use `tri.hinting_content.hintables` which is now async.

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -1223,14 +1223,16 @@ function pushSpace() {
 
     @hidden
 */
-export function hintables(
+export async function hintables(
     selectors = DOM.HINTTAGS_selectors,
     withjs = false,
     includeInvisible = false,
 ) {
     const visibleFilter = DOM.isVisibleFilter(includeInvisible)
     const elems = changeHintablesToLargestChild(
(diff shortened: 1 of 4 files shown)


bovine3dom:
LGTM

I made your observer apply to JS hints too, could you check I didn't do anything stupid there?

(Also I moved hintable construction up because selector typos weren't being caught by the commandline any more)

Chic-Tweetz:
JS hint changes look good, all seems to still work.

Nice catch on the selector typos I didn't spot that

PR #5437, opened 2026-07-17
Make implicit any error in sensitive parts of the code

bovine3dom:
addEventListener is easy to use with implicit any which circumvents all of the checks

this tries to stop that from happening, but it also makes loads of totally safe stuff
that uses any trigger the linter too, so that needs fixing, but i can't be bothered
right this second so let's shove it on the backlog

supersedes #5390


bovine3dom:
i don't understand how these checks are passing. on my machine i was getting ~50 errors, which was the whole point :thinking:

PR #5450, opened 2026-07-19
ts6 migration

bovine3dom:
i feel justified in putting this off for five years or whatever it was. man. 12 hours of my life. but hey i got to get rid of the `structuredClone` type definition


bovine3dom:
oh, right, because this builds on #5395 it makes it look even more enormous than it actually is. Only the last five commits are really about TS6 migration

bovine3dom:
i think strategically the sooner this gets merged the better because rebasing tiny PRs on this will be easier than rebasing this on tiny PRs (maybe?)

so i'll do a beta release without it, merge it, then start working through the backlog and getting PRs rebased on it

Issue #3880, opened 2021-08-03
Random hintnames


bovine3dom:
Sorry for not explaining it correctly. As you can see in the above image, there are a lot of hints jumbled across and this makes it quite difficult for me to target a specific text block that I want to.

Currently if I start typing `y` then the UI starts removing hints not starting with `y` and hence I am able to narrow down closer to the area I want to target.

**The problem is that with typing `y`, all the related elements are close to each other and hence removal of unrelated links is of no help as I am still clueless on which character to hit after `y`. `yl`, `yz` `yx`, `yc`, etc are all close enough that I still am unsure which one to use to target the first paragraph.**

A possible solution to this is to not order the hints and have them as distinct as possible so that when I type the first character, the closer links are unrelated enough for me to be able to narrow down more clearly. 

# Example:

- Current Order :`yl`, `yz` `yx`, `yc` all jumbled up close to each other. pressing `y` still leaves me confused which element to target as they're all close to each other.

- Proposed Order (essentially no order): `ax`, `bx`, `cx`, `dx`, etc. pressing any of the first character will filter out the other nearly elements making it obvious which element I am trying to target.

This will go a long way to add more robustness to the hinting system since more and more sites are starting to use a lot of javascript based interactions all jumbled across and it tends to leave a cluster of extremely hard to target hints as demonstrated in the image above.

I understand that you have a lot of issues to deal with in this project, hence this is by no means a priority, but it always felt that something like this would seriously help make this amazing extension even better.

I hope this explanation made sense, a bit verbose but I hope it makes sense.

_Originally posted by @sangeet in https://github.com/tridactyl/tridactyl/issues/1027#issuecomment-891068527_

bovine3dom:
@sangeet you "just" need to make a hintnames_random function here: https://github.com/tridactyl/tridactyl/blob/935fa10efe03d473e9be6d173cb2cf352ad1534e/src/content/hinting.ts#L587 : )

sangeet:
Thanks. :)

Ayan-web:
```js
function* hintnames(
    n: number,
    hintchars = defaultHintChars(),
): IterableIterator<string> {
    switch (config.get("hintnames")) {
        case "numeric":
            yield* hintnames_numeric(n)
        case "uniform":
            yield* hintnames_uniform(n, hintchars)
        default:
            yield* hintnames_short(n, hintchars)
    }
}
```
I just have to replace the `hintnames_uniform` with `hintname_random`
[...]

PR #5482, opened 2026-07-27
feat: add random hintnames with built-in English word list

mmustafasenoglu:
Adds a new `random` option for the `hintnames` setting that displays random English words as hint labels instead of the default short/numeric/uniform patterns.

## Changes

- **`src/content/hinting.ts`**: Added `hintnames_random(n)` generator function with a built-in word list of ~460 common short English words (2-3 letters). Uses Fisher-Yates shuffle for uniform random sampling. Updated `defaultHintChars()` to return the full lowercase alphabet in random mode so vimperator filtering works correctly.
- **`src/lib/config.ts`**: Added `random` to the `hintnames` config union type.

## Usage

```
set hintnames random
```

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -616,6 +616,9 @@ function defaultHintChars() {
     if (config.get("hintnames") === "numeric") {
         return "1234567890"
     }
+    if (config.get("hintnames") === "random") {
+        return "abcdefghijklmnopqrstuvwxyz"
+    }
     return config.get("hintchars")
 }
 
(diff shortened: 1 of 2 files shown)


bovine3dom:
This is hilarious and I love it but I disagree that it really does what #3880 was asking for

Currently when there are >26 hints, they look like this

```
aa as ad af ag
```

#3880 wants to instead have

```
aa sa da fa ga
```

but as I said this is hilarious so I've merged it and renamed it to `:set hintnames words`. Thanks for your contribution :)

Issue #5483, opened 2026-07-27
`:find` slowdown with many Ranges


Chic-Tweetz:
I suspect this'll happen with either normal `:find` or the new `:find --regex`

I believe creating too many `Range` objects eats up resources until they're garbage collected

Try going to the Tridactyl new page (ideal because there's a lot of text)

Run some irresponsible searches with a lot of matches so that lots of `Range` objects are created:

```
await tri.excmds.find("--regex", ".")
console.log("1")
await tri.excmds.find("--regex", "..")
console.log("2")
await tri.excmds.find("--regex", "...")
console.log("3")
await tri.excmds.find("--regex", "..")
console.log("4")
await tri.excmds.find("--regex", ".")
console.log("finished")
```

then try to hint (or anything really) and see that everything is now unresponsive

I found this same thing when I made [my own](https://github.com/Chic-Tweetz/tridactyl/blob/amo/src/content/finding.ts) regex/incsearch `:find` ... I wouldn't bother trying to make sense of that file though :)

But I did find that using `StaticRange` where possible avoids the slowdown, with the caveat that you do need a proper `Range` to get client rects. I use one real range and move its start/end as needed when rects are needed.

bovine3dom:
I think I fixed this in b5ba02e6953517b3a1db47320c5f93d142a5c202 which, yeah, sorry, way too big. but somewhere in all that mad node caching and moving the range cloning out of just all the ranges into only the one that we're trying to text select, it got quicker.

I was toying with adding a worker so that regex wouldn't make the main thread hang but 🤷 it's kinda quick enough now, even on stupidly huge pages like https://en.wikipedia.org/wiki/Battle_of_the_Bulge


i can't really talk given that i just merged an insane tree walker caching thing _and_ a heading indexer but the switcheroo between StaticRange and a real Range to get clientRects is a bit scary to me so I think I'd rather avoid it if the current approach works ok

Chic-Tweetz:
I can still make firefox unresponsive if i do something like `:find --regex .` (with https://github.com/tridactyl/tridactyl/commit/b5ba02e6953517b3a1db47320c5f93d142a5c202)

I also don't particularly like the shared real Range but am not sure how to avoid too much `createRange` otherwise. Could use the real range exclusively for rects and pass starts/ends to StaticRange constructors instead of passing the real Range.

It's probably not much of an issue (or at least much less of an issue) if not doing silly things like searching for every character anyway.

[...]

PR #5484, opened 2026-07-27
Finding: favour StaticRanges

Chic-Tweetz:
For #5483

Avoid creating too many `Range`s, instead store `StaticRange`s for `:find` matches.

Set the start/end of one real `Range` so that we can still get client rects.

I haven't been too thorough here, for instance I'm not sure where `currentMatchRange` might be used and that might need to return a real `Range`.

I'll leave this as a draft as more of a suggested fix

### DIFF
--- a/src/content/finding.ts
+++ b/src/content/finding.ts
@@ -26,37 +26,35 @@ function getFindHost() {
 
 const NATIVE_HIGHLIGHTS = typeof Highlight === "function" && "highlights" in CSS
 
-const createRangeProxy = (() => {
-    const sharedRange = document.createRange()
-
-    return function(rangeSpec: {
-        startContainer: Node,
-        startOffset: number,
-        endContainer: Node,
-        endOffset: number,
-    }): Range & { ["range"]: StaticRange } {
-        return new Proxy(new StaticRange(rangeSpec), {
-            get(target, prop) {
-                if (prop === "range") return target
-                if (prop !== "toString" && prop in target) return target[prop]
-                if (prop in sharedRange) {
-                    return (...args: any[]) => {
-                        sharedRange.setStart(target.startContainer, target.startOffset)
-                        sharedRange.setEnd(target.endContainer,target.endOffset)
-                        return (sharedRange as any)[prop](...args)
-                    }
-                }
-                return undefined
-            }
-        }) as Range & { ["range"]: StaticRange }
+class SharedRangeView extends StaticRange {
+    private static shared = document.createRange()
+    public toString() {
+        return this.moveSharedRange().toString()
     }
-})()
+    public cloneRange() {
+        return this.moveSharedRange().cloneRange()
+    }
+    public getBoundingClientRect() {
+        return this.moveSharedRange().getBoundingClientRect()
+    }
+    public getClientRects() {
+        return this.moveSharedRange().getClientRects()
+    }
+    get commonAncestorContainer() {
+        return this.moveSharedRange().commonAncestorContainer
+    }
+    private moveSharedRange() {
+        SharedRangeView.shared.setStart(this.startContainer, this.startOffset)
+        SharedRangeView.shared.setEnd(this.endContainer, this.endOffset)
+        return SharedRangeView.shared
+    }
+}
 
 class FindHighlight extends HTMLSpanElement {
     public top = Infinity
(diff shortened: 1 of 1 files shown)


src/content/finding.ts
```diff
@@ -26,14 +26,35 @@ function getFindHost() {
...
     public top = Infinity
-    public nativeRange: Range
+    public nativeRange: StaticRange
     private background = `var(--tridactyl-search-highlight-color)`
 
-    constructor(public range: Range) {
+    constructor(public range: StaticRange) {
         super()
-        this.nativeRange = NATIVE_HIGHLIGHTS ? range.cloneRange() : range
+        // I can't tell why there are two ranges depending on NATIVE_HIGHLIGHTS am I stupid
```
bovine3dom: um, iirc, firefox does ... stuff ... to the ranges if you add a highlight to one? i can't remember exactly what the problem was but this fixed it by letting it mess with a clone

Chic-Tweetz: ah okay - that's probably not a problem with a static range at least

Chic-Tweetz:
I've wrapped the shared real Range in a closure and use proxies which set its start/end when needed which is a bit more palatable to me

i am less keen on the "range" property i added to access the real StaticRange from a proxy (and the `& { ["range"]: StaticRange` I threw in to make typescript happy)

src/content/finding.ts
```diff
@@ -26,6 +26,32 @@ function getFindHost() {
...
+            get(target, prop) {
+                if (prop === "range") return target
+                if (prop !== "toString" && prop in target) return target[prop]
+                if (prop in sharedRange) {
+                    return (...args: any[]) => {
+                        sharedRange.setStart(target.startContainer, target.startOffset)
+                        sharedRange.setEnd(target.endContainer,target.endOffset)
+                        return (sharedRange as any)[prop](...args)
+                    }
+                }
```
bovine3dom: this converts every other property into a function which is ... not ideal? i think we'd need some cleverer way of spotting values e.g. doing

```js
if (prop === "commonAncestorContainer") {
    sharedRange.setStart(target.startContainer, target.startOffset)
    sharedRange.setEnd(target.endContainer, target.endOffset)
    return sharedRange.commonAncestorContainer
}
```

but automatically. (this makes offscreen stuff dodgy i think? i haven't tested too hard)

bovine3dom:
something in this breaks cross-element matches

```
<span>foo</span><span>bar</span>
:find -r oob
# :(
```

src/content/finding.ts
Chic-Tweetz: it is unnecessary really anyway, the rect functions just can't be called straight from a proxy

but we only use the two rect functions and commonAncestorContainer anyway so it could definitely be simplified... could avoid using proxies entirely and make a small class with only the properties that are used

edit: no I was wrong there we also check start/end containers/offsets, I can see why I went with proxies. Still, many ways to skin a cat

Chic-Tweetz:
didn't realise i'd messed with it enough to break anything - i'll at least get this in a working state next time i update it

oops I see, calling createRangeProxy with the same nodeIndex and nodeOffset for start and end that was silly

bovine3dom:
no rush, i am kinda on a feature freeze for 1.25.0 now anyway, want to get the beta tested for a few weeks without shifting stuff too much and then do a real release

Chic-Tweetz:
fair, 1.25.0 has plenty going on already

i've fixed that cross-element issue and replaced the proxies with a class extending StaticRange and implementing only the Range methods that are used

happy to leave it for now as it's working properly (AFAICT)

bovine3dom:
do you have an example where `:find -r` is faster before the patch?

`:find -r .` on the new tab page is ... about the same? just counting seconds in my head. both seem to take ~7 seconds

also, this PR only seems to work on Tridactyl pages and fails on any real website :eyes:

Chic-Tweetz:
well that's what i get for saying it's working 🙄 

it's not the speed of the finding itself, it's that Ranges don't appear to get completely cleaned up between searches so the more you use `:find` the slower the browser gets until the page crashes (i'm leaning towards it being a firefox bug)

Chic-Tweetz:
I've given `SharedRangeView` all properties that are used for either StaticRanges or Ranges instead of extending StaticRange itself which was what had broken it for real pages

PR #5485, opened 2026-07-27
feat: add hintnames random option for shuffled hints

### LINKED ISSUE #3880: Random hintnames

mmustafasenoglu:
## Summary

Adds a new `hintnames random` option that generates standard lexicographic hints and shuffles them with Fisher-Yates, so nearby DOM elements get different first characters.

This makes typed filtering more effective on pages with many hints clustered together.

## Problem

When there are >26 hints, they are generated lexicographically in DOM order:

```
aa as ad af ag
```

If these hints are on nearby elements, typing `a` still leaves many similar hints (`as`, `ad`, `af`), making it hard to narrow down the target.

With `hintnames random`:

```
aa sa da fa ga
```

Now typing `s` immediately filters to `sa`, which is on a specific element.

## Usage

```
:set hintnames random
```

## Implementation

- `hintnames_random(n, hintchars)` collects all `n` hints from the standard lexicographic stream and shuffles them with Fisher-Yates
- Uses the existing `shuffleArray` helper
- Added `"random"` to the `hintnames` config type

Fixes #3880

### DIFF
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -762,13 +762,13 @@ function* hintnames_words(n: number): IterableIterator<string> {
     )
 }
 
-/** Randomly shuffled lexicographic hintnames.
- * Shuffles standard hints so that nearby DOM elements get different first characters,
+/** Reversed lexicographic hintnames.
+ * Reverses each hint so that nearby DOM elements get different first characters,
  * making typed filtering more effective. See #3880.
  * @hidden */
-function* hintnames_random(n: number, hintchars = defaultHintChars()): IterableIterator<string> {
+function* hintnames_reverse(n: number, hintchars = defaultHintChars()): IterableIterator<string> {
     const hints = [...islice(hintnames_simple(hintchars), n)]
-    yield* shuffleArray(hints)
+    yield* hints.map(hint => hint.split("").reverse().join(""))
 }
 
 /** @hidden */
@@ -783,8 +783,8 @@ function* hintnames(
             yield* hintnames_uniform(n, hintchars)
         case "words":
             yield* hintnames_words(n)
-        case "random":
-            yield* hintnames_random(n, hintchars)
+        case "reverse":
+            yield* hintnames_reverse(n, hintchars)
         default:
             yield* hintnames_short(n, hintchars)
     }
--- a/src/content/hinting.ts
+++ b/src/content/hinting.ts
@@ -762,6 +762,15 @@ function* hintnames_words(n: number): IterableIterator<string> {
     )
 }
 
+/** Randomly shuffled lexicographic hintnames.
+ * Shuffles standard hints so that nearby DOM elements get different first characters,
+ * making typed filtering more effective. See #3880.
+ * @hidden */
+function* hintnames_random(n: number, hintchars = defaultHintChars()): IterableIterator<string> {
+    const hints = [...islice(hintnames_simple(hintchars), n)]
+    yield* shuffleArray(hints)
+}
+
 /** @hidden */
 function* hintnames(
     n: number,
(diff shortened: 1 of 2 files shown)


bovine3dom:
thanks for this.

i think in the original issue we spoke about random, but i think a better solution would be just to reverse each hint, which should naturally increase the distance between them. would you like to give that a go? probably calling it hintnames reverse

mmustafasenoglu:
Updated as suggested - renamed to `hintnames reverse` and replaced shuffle with deterministic string reversal.

Each hint is now reversed (e.g., `abc` → `cba`), which naturally increases the distance between nearby DOM elements since leksikografik olarak yakın hint'ler farklı ilk harfler alıyor.

Usage: `:set hintnames reverse`

Chic-Tweetz:
I think reversed hints would be better off using `hintnames_uniform` than `hintnames_short` to avoid situations like both `AH` and `A` existing

I notice that the word hints match before typing the whole word, eg. if there are two hints`FOE` and `WIT`, typing `W` will select the `WIT` hint

that's not necessarily a problem but i was basically expecting `:hint -qk *` to activate typing of the dead mode :)

mmustafasenoglu:
Thanks for the feedback! Updated the approach based on both comments:

**@bovine3dom:** Changed from `hintnames_random` to `hintnames reverse` as suggested. The reverse approach is cleaner - it deterministically reverses each lexicographic hint so nearby elements naturally get different first characters.

**@Chic-Tweetz:** Switched from `hintnames_simple` to `hintnames_uniform` as the source for reversal. This avoids the length collision issue (e.g. both "AH" and "A" existing) since all hints have the same length before reversal.

```typescript
// Before: hintnames_simple produces prefixing hints (a, aa, ab...)
// After reversal: still problematic length differences
const hints = [...islice(hintnames_simple(hintchars), n)]

// Now: hintnames_uniform produces same-length hints (aa, ab, ac...)
// After reversal: all same length, no collisions
yield* map(hintnames_uniform(n, hintchars), hint => hint.split("").reverse().join(""))
```

Usage: `:set hintnames reverse`

mmustafasenoglu:
Hi @bovine3dom @Chic-Tweetz, friendly ping — the PR has been updated with both suggestions:

- Renamed to `hintnames reverse` ✓
- Uses `hintnames_uniform` as source ✓
- Deterministic string reversal ✓

Let me know if there is anything else to adjust. Thanks!

mmustafasenoglu:
Closing due to 22 days of inactivity. Feel free to reopen when ready to continue.

bovine3dom:
i can't seem to reopen this myself so please do so if you would like it merged

22 days unfortunately really is not very long in terms of PRs on this repo :)

PR #5511, opened 2026-08-27
Add @arg tag support for documenting excmd flags

uhs-robert:
## Context

This came out of a discussion on #5397 (the which-key PR). The flags for the `hint` command needed some kind of documented lookup for which-key to use, and rather than adding that onto an already massive PR, this got split out as its own thing.

Which-key will build on top of this once it's merged. This PR only touches docs/build tooling, nothing else.

## Problem

Right now, the flags for a command like `:hint` (i.e., `-t`, `-b`, `-y`, etc.) are only documented as hand-written notes on the `:help` page. There's no way for other code (like which-key) to look up "what does `-t` do?" programmatically.

So, if a command's flag descriptions ever need to be shown somewhere else, they'd have to be typed out a second time and kept in sync manually.

## Solution

This PR adds a simple way to tag each flag right in the code comment like so:

```text
@arg -t open in a new foreground tab
@arg -b open in background
```

A build script reads these tags and turns them into a small lookup table (flag => description) that which-key, `:help`, and anything else can query directly.

The `:help` page also needed a small change to handle this well. Out of the box, the docs generator (TypeDoc) treats each `@arg` as its own separate section: a full heading repeated for every single flag (some commands, like hint, have a list of 20+). So, I customized how `@arg` tags get rendered by grouping them into a single bullet list under an "Arguments" heading. I also added a way for whoever writes a command's docs to choose exactly where in the paragraph that list shows up, using a placeholder `{{tridactyl-arg-list}}`, rather than always having it dumped at the very bottom.

## Limited Scope

I limited the scope of this PR to just the `hint` command. Other commands would need the same treatment but before doing any work in that direction, I wanted to see how this would land.

## Testing

- Typechecks and lints clean
- All unit tests pass
- Full build succeeds, checked the generated `:help` page by hand and confirmed `hint`'s flags show up correctly with the right formatting.

## Screenshots

This is how the `help` page for the `hint` command looks with this change:
image

### DIFF
--- a/scripts/convert_typedoc_metadata.js
+++ b/scripts/convert_typedoc_metadata.js
@@ -32,11 +32,11 @@ function convertMetadata(project) {
             .map(part => part.text || "")
             .join("")
             .replace(/\n+$/, "")
-    // Extract @arg tags: `@arg -flag description` -> flags["-flag"] = "description"
+    // Extract @flag tags: `@flag -flag description` -> flags["-flag"] = "description"
     const argFlags = comment => {
         const flags = {}
         for (const tag of comment?.blockTags || []) {
-            if (tag.tag !== "@arg") continue
+            if (tag.tag !== "@flag") continue
             const text = (tag.content || []).map(part => part.text || "").join("")
             const m = /^(-\S+)\s+(.+)$/.exec(text.trim())
             if (m) flags[m[1]] = m[2]
--- a/scripts/convert_typedoc_metadata.js
+++ b/scripts/convert_typedoc_metadata.js
@@ -32,14 +32,21 @@ function convertMetadata(project) {
             .map(part => part.text || "")
             .join("")
             .replace(/\n+$/, "")
-    // Extract @flag tags: `@flag -flag description` -> flags["-flag"] = "description"
     const argFlags = comment => {
         const flags = {}
         for (const tag of comment?.blockTags || []) {
             if (tag.tag !== "@flag") continue
-            const text = (tag.content || []).map(part => part.text || "").join("")
-            const m = /^(-\S+)\s+(.+)$/.exec(text.trim())
-            if (m) flags[m[1]] = m[2]
+            const text = (tag.content || [])
+                .map(part => part.text || "")
+                .join("")
+            const m = /^(-\S+)[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(text.trim())
+            if (!m) continue
+            const [, flag, short, rest] = m
+            const elaboration = rest.trim()
+            flags[flag] = {
+                short,
+                description: elaboration ? `${short}\n\n${elaboration}` : short,
+            }
         }
         return flags
     }
--- a/scripts/convert_typedoc_metadata.js
+++ b/scripts/convert_typedoc_metadata.js
@@ -39,7 +39,8 @@ function convertMetadata(project) {
             const text = (tag.content || [])
                 .map(part => part.text || "")
                 .join("")
-            const m = /^(-\S+)[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(text.trim())
+            const flagText = text.split(/\r?\n[ \t]*\r?\n/, 1)[0]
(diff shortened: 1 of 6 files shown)


src/excmds.ts
```diff
@@ -5620,6 +5602,39 @@ const KILL_STACK: Element[] = []
...
+ * @arg -F run JS callback
+ * @arg -c hint CSS selector only
+ * @arg -C hint CSS selector and defaults
+ * @arg -x exclude CSS selector
+ * @arg -f filter hints by text
+ * @arg -fr filter hints by regex
+ * @arg -J disable JS hints
+ * @arg -V include invisible elements
+ * @arg -q rapid (stay in hint mode)
+ * @arg -! execute all hints immediately
```
bovine3dom: is there a compelling reason to not just render these where they're written?

scripts/convert_typedoc_metadata.js
```diff
@@ -32,6 +32,17 @@ function convertMetadata(project) {
             .map(part => part.text || "")
             .join("")
             .replace(/\n+$/, "")
+    // Extract @arg tags: `@arg -flag description` -> flags["-flag"] = "description"
+    const argFlags = comment => {
+        const flags = {}
+        for (const tag of comment?.blockTags || []) {
+            if (tag.tag !== "@arg") continue
```
bovine3dom: i'm also debating whether `@arg` is the right name for this. to me an argument is a variable that can contain anything. could we maybe call it `@flag` instead? i guess that breaks down for `--arg=foo` style snippets.

i don't have a strong feeling which the right approach would be. but having arguments within arguments feels deeply strange

uhs-robert: Good point, I think `@flag` is probably the better choice here.

I'll get that added in today!

---
Added in https://github.com/tridactyl/tridactyl/pull/5511/commits/e744f5651de8ccfd04ac2bb75ece52074af53d84

src/excmds.ts
uhs-robert: Yes, there's a reason, but I'm starting to think the approach I took here is a bit over-engineered and we can do better.

The problem is TypeDoc's template always renders `@param`/`@returns` after the comment summary, and any custom tags get lumped into that summary. So when `@flag` sits below `@param` in the source, it still renders above `@param` in the generated docs.

Also any text below a tag belongs to that tag until another tag shows up. So all text after the last `@flag` gets swallowed into it, even though it shouldn't.

I think the simplest fix for both is to just write the built-in tags like `@param`/`@returns` last in the source, since that's where they will always render anyway (make the source order match the render order). And add a tag like `@endflags` to mark the end of a flag list so the rest of the comment doesn't get swallowed into the last flag.

uhs-robert: I just finished making those updates and added some headings via markdown to create mini-sections, let me know what you think. Also updated the screenshot above for the Hint section of the Help docs for a quick preview.

My hands are off now 🙌🏻

bovine3dom:
i hacked on it to make the renderer detect blank newlines negating the need for `@endflags`

if you don't have any strong feelings @uhs-robert i would say this is good to merge

uhs-robert:
Thanks for adding the tsdoc.json, that's a much smarter approach.

As for the hack to get rid of `@endflags`, I completely agree. I much prefer the hack too. Looks good to me, feel free to merge!

As an aside, for whichkey, I may need to do a sweep of the rest of the excmds to see if we need to add flags anywhere else. Should I make that a seperate PR or tack it onto the whichkey PR?

bovine3dom:
separate PR i think

bonus points if you can at least think about how we might reuse it for excmd completions while you're there :)

uhs-robert:
> separate PR i think
> 
> bonus points if you can at least think about how we might reuse it for excmd completions while you're there :)

Great idea, I just got excmd completion working locally. I am still tweaking it a bit now.

image

I'm thinking... once this PR gets merged then I'll make another PR specifically for the excmd completion. Then once that is done, I'll focus on a PR to flesh out the rest of the excmds to use the new flag sytem. That way we're building on top of a solid foundation and not having to go back and redo documentation if any changes are needed.

bovine3dom:
awesome! that was quick :)

PR #5512, opened 2026-08-28
Add excmd flag suggestions to excmd completion while typing

uhs-robert:
## Summary

Follow-up to #5511, @bovine3dom asked if we could reuse the new `@flag` metadata for excmd completions, so that's what this does.

Typing `hint` followed by a space now pops up a completion box listing its flags with the short label from `@flag`:

- Single-char flags (`-t`, `-b`, `-c`, ...) combine into whatever token you've already typed, in any order, so typing `-Jc` still offers `-t` etc.
- Multi-char ones (like `-pipe`, `-fr`, `-wp`) are shown as their whole string.

## Mutual-exclusion groups

`hint` has "only one can be specified" action flags (`-t`, `-b`, `-y`, ...), so once you've picked `-t` we shouldn't keep suggesting `-b`/`-y`/etc alongside it.

`@flag` now takes an optional `{group}` marker (`@flag -t {action} ...`). Any flag sharing a group already present in the token gets excluded from suggestions.

Applied `{action}` to hint's 23 mutually-exclusive action flags. The rest (element selection, hinting mode) can be combined freely.

## Testing

- [ ] Typing `:hint -` shows the flag completion box with labels for each command and their flag short description
- [ ] Typing `hint -t` does not suggest any other `{action}` group flags like `-b`/`-y`, but instead suggests `-c`/`-J`/etc
- [ ] Typing `hint -ft` (mixing an ungrouped flag with an action flag) still excludes other action flags, same as `-t` alone
- [ ] Typing `hint -pi` offers completing to `-pipe`, `hint -w` offers `hint -wp`, and both suggest `-c`/`-J`/etc
- [ ] Typing an incorrect flag like `hint -Q` or `hint -tQ` does not return any suggestions
- [ ] Picking a multi-char flag (`-pipe`, `-W`, ...) inserts it with a trailing space
- [ ] Helps docs still render hint's flag list the same as before (no visible changes there)

### `hint -`
image

### `hint -p`
image

### `hint -Jc`
image

### DIFF
--- a/scripts/convert_typedoc_metadata.js
+++ b/scripts/convert_typedoc_metadata.js
@@ -40,13 +40,16 @@ function convertMetadata(project) {
                 .map(part => part.text || "")
                 .join("")
             const flagText = text.split(/\r?\n[ \t]*\r?\n/, 1)[0]
-            const m = /^(-\S+)[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(flagText.trim())
+            const m = /^(-\S+)(?:[ \t]+\{(\w+)\})?[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(
+                flagText.trim(),
+            )
             if (!m) continue
-            const [, flag, short, rest] = m
+            const [, flag, group, short, rest] = m
             const elaboration = rest.trim()
             flags[flag] = {
                 short,
                 description: elaboration ? `${short}\n\n${elaboration}` : short,
+                ...(group ? { group } : {}),
             }
         }
         return flags
--- a/scripts/typedoc-theme.mjs
+++ b/scripts/typedoc-theme.mjs
@@ -272,11 +272,13 @@ function parseFlagTag(tag) {
         ]
         break
     }
-    const m = /^(-\S+)[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(text.trim())
+    const m = /^(-\S+)(?:[ \t]+\{(\w+)\})?[ \t]+([^\n]+)\n*([\s\S]*)$/.exec(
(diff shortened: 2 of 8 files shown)


bovine3dom:
this is a good start but there are loads of messy edge cases

- `{action}` isn't flexible enough to describe e.g. `-cC` conflict or `-pipe` conflict with everything

- single character flags that require arguments like `hint -W [excmd]` don't know this and encourage you to add more flags afterwards

- `-pipe` is handled badly `:hint -qpi` never shows `-qpipe` and `:hint -pi` shows loads of invalid options

- aliases/commands don't work (but i could live with this)


i _think_ most of the problem comes from us trying to maintain a second hint arg parser here. we should instead reuse (extract?) the real hint arg parser and only use `@flag` for documentation. so `{action}` could then go away

uhs-robert:
> this is a good start but there are loads of messy edge cases
> 
> * `{action}` isn't flexible enough to describe e.g. `-cC` conflict or `-pipe` conflict with everything
> * single character flags that require arguments like `hint -W [excmd]` don't know this and encourage you to add more flags afterwards
> * `-pipe` is handled badly `:hint -qpi` never shows `-qpipe` and `:hint -pi` shows loads of invalid options
> * aliases/commands don't work (but i could live with this)
> 
> i _think_ most of the problem comes from us trying to maintain a second hint arg parser here. we should instead reuse (extract?) the real hint arg parser and only use `@flag` for documentation. so `{action}` could then go away

Good point, probably better to extract and reuse the main hint arg parser rather than duplicating effort.

I might have some time this weekend and/or next week to work on this. I'll follow up when I've got this put together!

PR #5516, opened 2026-08-30
Fix #5045: add optional tab argument to :tabmove

### LINKED ISSUE #5045: add :tabmove [direction] [index]
Defining an ex mode binding to move selected tab up or down within buffer list view just keeps moving currently opened tab and ignores the selected one.

`bind --mode=ex <A-≤> ex.execute_ex_on_completion_args tabmove -1`
`bind --mode=ex <A-≥> ex.execute_ex_on_completion_args tabmove +1` 

Not working

MarcGonzalezMoratona:
Fixes #5045.
  `tabmove` ignored extra arguments and always moved the active tab, so `ex.execute_ex_on_completion_args tabmove -1` over the buffer list moved the wrong tab. It now takes an optional second argument (the tab to move), resolved with `tabFromIndex`, defaulting to the current tab.

  ### Human test
  1. Open 4 tabs; stay on tab 1.
  2. `bind --mode=ex <A-j> ex.execute_ex_on_completion_args tabmove +1`
  3. `b`, select tab 3, press `<A-j>` → tab 3 moves (previously tab 1 moved). ✓
  4. With pinned tabs: selecting an unpinned tab moves the correct one. ✓
  5. Regression: `<<` / `>>` / `:tabmove 2` still act on the active tab. ✓

Note: Human test was done in a MacBook, so the bind I used was `bind --mode=ex <C-j> ex.execute_ex_on_completion_args tabmove +1`

  ### Scope note
  Uses the standard buffer-list numbering (what the issue's flow uses).
  The `:tabmove`-prefixed completion's physical/pinned-group numbering is a separate concern and left out of scope.

### DIFF
--- a/src/excmds.ts
+++ b/src/excmds.ts
@@ -3338,7 +3338,7 @@ export async function undo(item = "recent"): Promise<number> {
 }
 
 /**
- * Move the current tab to be just in front of the index specified.
+ * Move a tab, defaulting to the current one, to be just in front of the index specified.
  *
  * Known bug: This supports relative movement with `tabmove +pos` and `tabmove -pos`, but autocomplete doesn't know that yet and will override positive and negative indexes.
  *
@@ -3349,10 +3349,12 @@ export async function undo(item = "recent"): Promise<number> {
  * @param index New index for the current tab.
  *
  * 1,start,^ are aliases for the first index. 0,end,$ are aliases for the last index.
+ *
+ * @param tabToMove Which tab to move. Defaults to the current tab.
  */
 //#background
-export async function tabmove(index = "$") {
-    const aTab = await activeTab()
+export async function tabmove(index = "$", tabToMove?: string) {
+    const aTab = await tabFromIndex(tabToMove)
     if (index === "#") {
         const previousTab = await prevActiveTab()
         if (previousTab.index - aTab.index === 1) {


MarcGonzalezMoratona:
Hi! This is my first open source contribution, so thanks in advance for your patience.

I used AI agents to help me work through this. That said, I didn't just accept the output. I read and reviewed every line of the change, made sure I understood why `tabmove` was ignoring the selected tab (it only ever read the first argument and always operated on the active tab), and iterated on the approach until it was minimal and correct.

I also tested it manually end to end:
  - Reproduced the original bug: with a binding like `ex.execute_ex_on_completion_args tabmove +1`, selecting a non-active tab in the buffer list moved the active tab instead of the selected one.
  - Confirmed the fix: the selected tab now moves as expected.
  - Checked pinned tabs and verified `<<` / `>>` and `:tabmove <n>` still act on the current tab (no regression).

Happy to adjust anything — naming, scope, or the approach — if you'd prefer it done differently.

bovine3dom:
this works great, thanks! welcome to the world of open source contributors :)

i don't know what to do about the `Co-Authored-By` line - it has never come up before. to me, only humans can be coauthors, but i know that some people like to know if projects have had any LLM-assisted contributions and use that as an indicator of it.

so i have asked a few friends what they think and i'll wait before deciding whether to remove the Co-Authored-By line before i merge this. but i will definitely merge this eventually because it is a great first contribution.

MarcGonzalezMoratona:
> this works great, thanks! welcome to the world of open source contributors :)
> 
> i don't know what to do about the `Co-Authored-By` line - it has never come up before. to me, only humans can be coauthors, but i know that some people like to know if projects have had any LLM-assisted contributions and use that as an indicator of it.
> 
> so i have asked a few friends what they think and i'll wait before deciding whether to remove the Co-Authored-By line before i merge this. but i will definitely merge this eventually because it is a great first contribution.

Thank you so much! Feel free to remove the Co-Authored-By line, I just used Claude to help me understand the project and the issue and writing a polished commit message but I was the one implementing the fix.

This Co-Authored-By line it's a default watermark that I think they've added recently, but there's no issue in removing it.

PR #5530, opened 2026-09-13
Fix #5525: add gg/G follows focus setting

### LINKED ISSUE #5525: `gg` and `G` do not reliably scroll a separately scrollable area after I focus an element inside it with `;;`
### Description

`gg` and `G` do not reliably scroll a separately scrollable area after I focus an element inside it with `;;`.

I first noticed this on LinkedIn search pages, where the search results appear in an independently scrollable panel. The issue is also reproducible with scrollable iframes.

I have not yet tested every type of element that can have its own scrolling behavior, so I cannot say whether this affects all nested scroll containers or only particular cases such as iframes and LinkedIn's results panel.

### Steps to reproduce: iframe

1. Open a page containing a scrollable iframe.
2. Use `;;` to focus an element inside the iframe.
3. Use normal scrolling keys such as `j` and `k`; scrolling inside the iframe works.
4. Press `gg` or `G`.

### Expected behavior

After `;;` focuses an element inside a separately scrollable area, `gg` should scroll that area to the top and `G` should scroll it to the bottom.

For an iframe, this means scrolling the iframe document rather than the top-level page.

### Actual behavior

- `gg` and `G` do not reliably scroll the focused iframe or nested scrollable panel.
- `gg` sometimes works after repeated attempts, but is not dependable.
- Normal scrolling in the same focused area works after `;;`.

### Additional example

On LinkedIn search pages, the search-results panel can scroll independently of the overall page. After focusing an element in that panel with `;;`, normal scrolling works, but `gg` and `G` do not reliably move the panel to its beginning or end.

### Comparison

I tested the iframe case with Vimium-FF and Vimium C. In both extensions, `gg` and `G` correctly scroll the focused iframe to the top and bottom.

bovine3dom:
Fix #5525

`ggscrollsroot` is a terrible name and needs to be changed before merging


archerhush:
The current draft does not build because `scrollto` compares the Promise returned by `config.getAsync("ggscrollsroot")` directly with `"true"`:

```
TS2367: This comparison appears to be unintentional because the types 'Promise<any>' and '"true"' have no overlap.

Did you forget to use 'await'?
```

Since `scrollto` is synchronous, changing the lookup to `config.get(...)` fixes the build:

```
- config.getAsync("ggscrollsroot") === "true"
+ config.get("ggscrollsroot") === "true"
```

After rebuilding and loading the extension temporarily, I tested:

```
:set ggscrollsroot false
```

After `;;` focused a link inside my scrollable iframe, `j`/`k` scrolled the iframe, and `gg`/`G` correctly moved that same iframe context to its top/bottom.

With:

```
:set ggscrollsroot true
```

`gg` and `G` retained the existing root-document behavior.

So the intended implementation works for my reproduction once the synchronous config lookup is used.

Thank you.

bovine3dom:
yeah sorry i saw i had used the wrong function but didn't think anyone would be looking at this PR and decided i would fix it later

thank you for taking the time to test it

src/excmds.ts
```diff
@@ -1448,7 +1448,7 @@ export function scrollto(a: number | string, b: number | "x" | "y" = "y") {
             done = scrolling.recursiveScroll(
                 0,
                 percentage === 0 ? -Infinity : Infinity,
-                document.documentElement,
+                config.getAsync("ggscrollsroot") === "true" ? document.documentElement : undefined,
```
```suggestion
                config.get("ggscrollsroot") === "true" ? document.documentElement : undefined,
```

bovine3dom:
I need to look at this again because it's not working like I would expect it to :(

```
set scrolltofollowfocus true
# G/gg in a nested iframe. the whole page seems to scroll rather than just the most nested frame
```

Chic-Tweetz:
https://github.com/Chic-Tweetz/tridactyl/commit/44fb645d4eb2df415f9b44bc9e23dfd6376d3ef8
I've made `:scrollto` behave more like `:scrollpx` (and chucked a `--target=root|focused` check in there)

I also added a getter for the current scroll focus, but that's just because I thought it might be useful to expose after making those commands https://github.com/tridactyl/tridactyl/issues/5525#issuecomment-5794331414

and that seems to behave pretty well

couple of things stick out though

- it might be good to add a `scrollIntoView` in `:scrollto` if the scrolled element isn't the root, in case it's off-screen
- in @archerhush's iframe test https://archerhush.github.io/test/iframes.html the iframes don't get scroll focus when doing a normal `f` hint because the iframes don't exist until a link is selected... not sure there's much that can be done about that
after an additional `;;` hint the nested frames scroll properly

PR #5533, opened 2026-09-17
Fix #5532: access actual object rather than copy

### LINKED ISSUE #5532: `:reset` and `:unset` can fail
Testing on 1.25.1 Carpenter / latest

### Steps to reproduce

Try unbinding a key in a mode with an `🕷🕷INHERITS🕷🕷` key:
```
bind --mode=a a fillcmdline a
bind --mode=b 🕷🕷INHERITS🕷🕷 amaps
bind --mode=b b fillcmdline unbindme
reset --mode=b b
```

### Expected behaviour
The `b` bind should be removed from `bmaps`. It shouldn't show up in `:viewconfig bmaps`

### Actual behaviour
Nothing changes, the `b` bind continues to exist in `bmaps`

---

### Cause

When a config object has an `🕷🕷INHERITS🕷🕷` key, `getDeepProperty` returns a new object, not a subobject of `USERCONFIG`

`unset` uses `getDeepProperty` to get a parent object which it then deletes a property of

if the parent object isn't in `USERCONFIG` then deleting the property doesn't change anything

https://github.com/tridactyl/tridactyl/blob/aa225538eda171c218c9e7a8755345450642023e/src/lib/config.ts#L2301

---

It may be enough to simply follow the target path through `USERCONFIG`, not sure if there are situations where you'd need a merged object that `getDeepProperty` would produce

```
export async function unset(...target) {
    if (IN_BACKGROUND && EXCLUSIVE_PENDING) await EXCLUSIVE_QUEUE
    if (IN_BACKGROUND && !INITIALISED) await getAsync()

    let parent = USERCONFIG
    for (const key of target.slice(0, - 1)) {
        parent = parent[key]
        if (typeof parent !== "object" || Array.isArray(parent)) {
            parent = undefined
            break
        }
    }

    if (parent !== undefined) delete parent[target[target.length - 1]]
    if (!IN_BACKGROUND) return mutateInBackground("unset", target)
    return save()
}
```

bovine3dom:
Fix #5532

Untested, just a hunch

Test with

```
bind --mode=a a fillcmdline a
bind --mode=b 🕷 🕷 INHERITS🕷 🕷  amaps
bind --mode=b b fillcmdline unbindme
reset --mode=b b
viewconfig bmaps
```


bovine3dom:
Works :sunglasses:

